PR #62, opened 2024-08-09
reward: SupplyManager reaccumulates from nearest persisted AccReward

blukat29:
## Proposed changes

- Address the issue where an archive node not serving `kaia_getTotalSupply` for non-128-multiple blocks.
- Cause
  - SupplyManager writes `AccReward` every BlockInterval. The intention was to make full nodes write every `BlockInterval` whereas archive nodes write every `1` block.
  - In reality, archive nodes may set `BlockInterval == 128 (default)` because it didn't matter so far. The [writeStateTrie loop](https://github.com/kaiachain/kaia/blob/v1.0.1/blockchain/blockchain.go#L1363) ignores BlockInterval when gcmode is archive.
  - Therefore, SupplyManager didn't store `AccReward` every `1` block. Rather it stored every 128 blocks in archive nodes with default settings.
- Fix
  - Have SupplyManager calculate `AccReward` for any given block, by re-accumulating from the nearest persisted `AccReward`.
  - Note that the re-accumulation only requires blocks and receipts which should be available in an archive node.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Further comments

- Prepare a Kairos archive node
```js
> admin.nodeConfig.TrieBlockInterval
128
> num = 157740851
> num % 128
51
```
- Before PR (1afbbfb3)
```js
> eth.getBalance("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", num)
49924765500000000000
> kaia.getTotalSupply(num)
Error: accumulated reward not stored
	at web3.js:6812:9(39)
	at send (web3.js:5223:62(29))
	at <eval>:1:20(3)
```
- After PR (ee14c8b)
```js
> eth.getBalance("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", num)
49924765500000000000
> kaia.getTotalSupply(num)
{
  burntFee: "0x899be299255df6283350",
  deadBurn: "0x35a29f8b3e27ba28cd",
  kip103Burn: "0x6a15eef7e3a354cf657913",
  kip160Burn: "-0xbf82646906be407e42a4800",
  number: "0x966ef33",
  totalBurnt: "-0xb8d8029a1f5dc275c85f9f6",
  totalMinted: "0x446c3b15f9926687d2c87232559657244007b80000",
  totalSupply: "0x446c3b15f9926687d2d3ffb27f384d0067643df9f6",
  zeroBurn: "0x65c7159b3ef9a5c78da"
}
```

### DIFF
--- a/reward/supply_manager.go
+++ b/reward/supply_manager.go
@@ -40,6 +40,7 @@ import (
 var (
 	supplyCacheSize   = 86400          // A day; Some total supply consumers might want daily supply.
 	supplyLogInterval = uint64(102400) // Periodic total supply log.
+	supplyReaccLimit  = uint64(1024)   // Re-accumulate from the last accumulated block.
 	zeroBurnAddress   = common.HexToAddress("0x0")
 	deadBurnAddress   = common.HexToAddress("0xdead")
 
@@ -143,9 +144,9 @@ func (sm *supplyManager) GetAccReward(num uint64) (*database.AccReward, error) {
 		return nil, errNoAccReward
 	}
 
-	accReward := sm.db.ReadAccReward(num)
-	if accReward == nil {
-		return nil, errNoAccReward
+	accReward, err := sm.getAccRewardUncached(num)
+	if err != nil {
+		return nil, err
 	}
 
 	sm.accRewardCache.Add(num, accReward.Copy())
@@ -318,7 +319,7 @@ func (sm *supplyManager) catchup() {
 	for lastNum < headNum {
 		logger.Info("Total supply big step catchup", "last", lastNum, "head", headNum, "minted", lastAccReward.Minted.String(), "burntFee", lastAccReward.BurntFee.String())
 
-		accReward, err := sm.accumulateReward(lastNum, headNum, lastAccReward)
+		accReward, err := sm.accumulateReward(lastNum, headNum, lastAccReward, true)
 		if err != nil {
 			if err != errSupplyManagerQuit {
 				logger.Error("Total supply accumulate failed", "from", lastNum, "to", headNum, "err", err)
@@ -341,7 +342,7 @@ func (sm *supplyManager) catchup() {
 		case head := <-sm.chainHeadChan:
 			headNum = head.Block.NumberU64()
 
-			supply, err := sm.accumulateReward(lastNum, headNum, lastAccReward)
+			supply, err := sm.accumulateReward(lastNum, headNum, lastAccReward, true)
 			if err != nil {
 				if err != errSupplyManagerQuit {
 					logger.Error("Total supply accumulate failed", "from", lastNum, "to", headNum, "err", err)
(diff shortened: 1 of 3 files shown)


reward/supply_manager.go
```diff
@@ -379,9 +380,35 @@ func (sm *supplyManager) totalSupplyFromState(num uint64) (*big.Int, error) {
...
+			fromAcc = accReward
+			break
+		}
+	}
+	if fromAcc == nil {
+		return nil, errNoAccReward
+	}
+
+	logger.Trace("on-demand reaccumulating rewards", "from", fromNum, "to", num)
+	return sm.accumulateReward(fromNum, num, fromAcc, false)
```
hyeonLewis, line 405: I think `write` is meaningless for this since in most cases there's no checkpoint interval between `fromNum` and `num`. And even if there's a case, I think writing missing `accReward` at checkpoint to db isn't needed to be prevented.

reward/supply_manager.go
```diff
@@ -143,9 +144,9 @@ func (sm *supplyManager) GetAccReward(num uint64) (*database.AccReward, error) {
 		return nil, errNoAccReward
```
hyeonLewis, line 144: Currently this branch isn't effective in production since usually it has cache for the recent blocks. But if caches are gone, it may return error even if it can calculate the latest `accReward` thanks to this PR (re-accumulate from `lastNum` to `num`). Do you think it'll be still effective after this PR?

reward/supply_manager.go
```diff
@@ -379,9 +380,35 @@ func (sm *supplyManager) totalSupplyFromState(num uint64) (*big.Int, error) {
...
+func (sm *supplyManager) getAccRewardUncached(num uint64) (*database.AccReward, error) {
+	accReward := sm.db.ReadAccReward(num)
+	if accReward != nil {
+		return accReward, nil
+	}
+
+	// Trace back to the last stored accumulated reward.
+	var fromNum uint64
+	var fromAcc *database.AccReward
+	for i := uint64(1); i < supplyReaccLimit; i++ {
```
hyeonLewis, line 392: Why don't we just use `sm.db.ReadLastAccRewardBlockNumber`?

blukat29: LastAccRewardBlockNumber is _the highest_ number ever persisted. When API requested 257, we need to find 256 but Last... is like 161234567.

reward/supply_manager.go
blukat29: To prevent overwriting Last number, and to keep database size from growing via API.

hyeonLewis: Oh, sure. I got the point.

reward/supply_manager.go
hyeonLewis: I think we don't need to consider unessential case. Sorry for the hussle.

hyeonLewis (approved):

reward/supply_manager.go
blukat29: Sounds fair. I think I intended block unwinding, but unwinding can be done by DeleteAccReward instead.

reward/supply_manager.go
```diff
@@ -379,9 +380,35 @@ func (sm *supplyManager) totalSupplyFromState(num uint64) (*big.Int, error) {
 	return totalSupply, nil
 }
 
+func (sm *supplyManager) getAccRewardUncached(num uint64) (*database.AccReward, error) {
+	accReward := sm.db.ReadAccReward(num)
+	if accReward != nil {
+		return accReward, nil
+	}
+
```
ian0371, line 388: maybe introduce a fast path by trying `num - num % checkpointInterval`?

blukat29: In theory, checkpointInterval can be different after a node restart because it originates from the node config TrieBlockInterval (`--state.block-interval`). I don't want this to rely on the 128-by-default assumption. Same logic applied in https://github.com/kaiachain/kaia/blob/v1.0.1/node/cn/state_accessor.go#L87.

reward/supply_manager.go
blukat29: Yes the branch is harmless. Let's keep it this way.

reward/supply_manager.go
ian0371: yeah I understand why the logic below must exist, but I think it's very likely to hit the branch if we introduced one before going down all the blocks. It can cause one additional DB read overhead, but a high chance of hitting.

blukat29: Nice idea. Added in c1401375.

hyeonLewis: it's not related to this PR, but ian's comment gives me a thing to consider about block interval. It may occur some problems if node operator changes block interval to unexpected values. For example, it's not safe to use 128-by-default assumption and even TrieBlockInterval since when node startup with updated `TrieBlockInterval`, it might not have states at `num - num%updatedTrieBlockInterval`. Maybe we also need to consider this for our modular architecture.

ian0371 (approved):

reward/supply_manager.go
blukat29: @hyeonLewis Absolutely. We shall not rely on the "fixed block interval" as it is not a universal concept.

hyunsooda (approved):

PR #67, opened 2024-08-13
reward: SupplyManager correctly handles block rewinds

blukat29:
## Proposed changes

- The GetTotalSupply API had a bug that over-accumulates TotalMinted and TotalBurnt after a manual block rewinding via `debug_setHead`. See https://app.circleci.com/pipelines/github/kaiachain/kaia/355/workflows/a330fe31-08f5-4771-bd3a-3459d4849a70/jobs/1776 and `reward/supply_manager_test.go:TestCatchupRewind` for the test reproducing it.

1. SupplyManager catchup thread is guarded against out-of-order ChainHeadEvent that might be emitted after block rewind (i.e. debug_setHead)
2. Block rewind (i.e. BlockChain.setHeadBeyondRoot) involves deletion of SupplyCheckpoints, and restarting the catchup thread.
3. Invalidate existing supply checkpoints (dubbed "accRewards"), triggers every node upgrading to v1.0.2 re-accumulate from the genesis. Because of the bug, nodes v1.0.1 or earlier which undergone setHead at least once might have corrupt accRewards data.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -611,6 +611,11 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 			// to low, so it's safe the update in-memory markers directly.
 			bc.currentFastBlock.Store(newHeadFastBlock)
 		}
+
+		// Rewind the supply checkpoint
+		newLastSupplyCheckpointNumber := header.Number.Uint64() - (header.Number.Uint64() % params.SupplyCheckpointInterval)
+		bc.db.WriteLastSupplyCheckpointNumber(newLastSupplyCheckpointNumber)
+
 		return bc.CurrentBlock().Number().Uint64(), nil
 	}
 
@@ -629,6 +634,7 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 		if bc.Config().Istanbul.ProposerPolicy == params.WeightedRandom && !bc.Config().IsKaiaForkEnabled(new(big.Int).SetUint64(num)) && params.IsStakingUpdateInterval(num) {
 			bc.db.DeleteStakingInfo(num)
 		}
+		bc.db.DeleteSupplyCheckpoint(num)
 	}
 
 	// If SetHead was only called as a chain reparation method, try to skip
--- a/node/cn/api_backend.go
+++ b/node/cn/api_backend.go
@@ -100,6 +100,8 @@ func (b *CNAPIBackend) SetHead(number uint64) error {
 	b.cn.protocolManager.Downloader().Cancel()
 	b.cn.protocolManager.SetSyncStop(true)
 	defer b.cn.protocolManager.SetSyncStop(false)
+	b.cn.supplyManager.Stop()
+	defer b.cn.supplyManager.Start()
 	return doSetHead(b.cn.blockchain, b.cn.engine, b.cn.governance, b.gpo, number)
 }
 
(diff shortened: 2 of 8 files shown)


2dvorak (approved):

ian0371 (approved):

blockchain/blockchain.go
```diff
@@ -611,6 +611,11 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 			// to low, so it's safe the update in-memory markers directly.
 			bc.currentFastBlock.Store(newHeadFastBlock)
 		}
+
+		// Rewind the supply checkpoint
+		newLastSupplyCheckpointNumber := header.Number.Uint64() - (header.Number.Uint64() % params.SupplyCheckpointInterval)
+		bc.db.WriteLastSupplyCheckpointNumber(newLastSupplyCheckpointNumber)
```
hyunsooda, line 617: 1. This unconditionally stores the nearest check point number. Is it still safe if the rewinding number is beyond the nearest check pointer number because check point key may be deleted more than this range in `delFn`.

2. It works with a case below
------------ c1(exist) ------------ c2(exist) ------------ c3(exist) ------------ headerNumber ------------

is it safe?
------------ c1(exist) ------------ c2(exist) ------------ c3(non-exist) ------------ headerNumber ------------

* c = checkpoint

maybe we do not need to care much about this because `accumulateReward` will be triggered?

blukat29:
Fixed api_backend_test unit test and linter.

blockchain/blockchain.go
blukat29: a. About the feasibility of the situation,
  Suppose c1=128, c2=256, c3=384.
  How can we arrive that situation? start with currentHeader=400, SetHead(270), then immediately SetHead(380)?

b. For recovery,
  - `catchup()` is always (re)started after sethead
    - When setHead is performed at node startup (via --start-block-num or hardfork config mismatch, etc) then LastNum is rewinded by `updateFn` then `catchup()` starts.
    - When setHead is triggered by debug_setHead API, then APIBackend.SetHead Stop() and Start() the SupplyManager.
  - When `catchup()` starts, it checks that the SupplyCheckpoint(num = LastCheckpointNumber()) exists and if not, the goroutine will restart from the genesis.

hyunsooda: a. Given the situation, I thought that last supply checkpoint number would be 384, but some of keys were deleted more than 14. However, the `updateFn` is called for every block number (I confused the delete function was only invoked for every block number without looking the detail). So, it was false alarm. Sorry for noise.

b.  Starting from genesis may be heavy work if it does not occur rarely. For example, we did `debug_setHead(klay.blockNumber - 1)`, which involves to modify the last checkpoint to `lastcheckpoint -  (headerNumber % lastchecjpoint)` and if there's no corresponding key with this number, then reading last checkpoint will be zero, result in building from scratch, but I 'm not sure where this case does exist (e.g., c1, c2 exists, c3 non-exist, c4 exist).

blukat29: a. Yes that's right.
b. If you see the upper part of `updateFn`, if there state missing the header it rewinds to genesis because you encountered an unexpected "gap". I'd employ the same strategy.

hyunsooda (approved):

jeongkyun-oh (approved):

PR #107, opened 2024-10-10
blockchain: Implement EIP-2028 calldata pricing

hyunsooda:
## Proposed changes

The gas calculation for EIP-2028 has been integrated into IntrinsicGasPayload(). The expected gas values are also being adjusted with precise HF state.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


blockchain/types/transaction_test.go
```diff
@@ -633,13 +635,21 @@ func TestIntrinsicGas(t *testing.T) {
 		assert.Equal(t, tc.expectGas2, gas)
 		assert.Equal(t, nil, err)
 
-		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsIstanbul: false})
+		gas, err = IntrinsicGas(data, nil, false, params.Rules{IsPrague: true})
```
blukat29, line 638: Just to be safe..
```suggestion
		gas, err = IntrinsicGas(data, nil, false, params.Rules{IsIstanbul: true, IsPrague: true})
```

blockchain/types/transaction_test.go
```diff
@@ -633,13 +635,21 @@ func TestIntrinsicGas(t *testing.T) {
...
 		assert.Equal(t, tc.expectGas3, gas)
 		assert.Equal(t, nil, err)
 
-		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsIstanbul: true})
+		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsIstanbul: false})
```
blukat29, line 642:
```suggestion
		gas, err = IntrinsicGas(data, nil, true)
```

blockchain/types/transaction_test.go
```diff
@@ -633,13 +635,21 @@ func TestIntrinsicGas(t *testing.T) {
...
 		assert.Equal(t, tc.expectGas4, gas)
 		assert.Equal(t, nil, err)
+
+		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsIstanbul: true})
+		assert.Equal(t, tc.expectGas5, gas)
+		assert.Equal(t, nil, err)
+
+		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsPrague: true})
```
blukat29, line 650:
```suggestion
		gas, err = IntrinsicGas(data, nil, true, params.Rules{IsIstanbul: true, IsPrague: true})
```

blockchain/types/tx_internal_data.go
```diff
@@ -610,14 +608,7 @@ func IntrinsicGas(data []byte, accessList AccessList, contractCreation bool, r p
...
 
-	var gasPayloadWithGas uint64
-	var err error
-	if r.IsIstanbul {
-		gasPayloadWithGas, err = IntrinsicGasPayload(gas, data, contractCreation, r)
-	} else {
-		gasPayloadWithGas, err = IntrinsicGasPayloadLegacy(gas, data)
-	}
-
+	gasPayloadWithGas, err := IntrinsicGasPayload(gas, data, contractCreation, rules)
```
blukat29, line 611: Can we make it more clearer?

```suggestion
	payloadGas, err := IntrinsicGasPayload(data, isContractCreation, rules)
	if err != nil {
		return 0, err
	}
	gas += payloadGas
```

ian0371 (commented):
lgtm except why did you change `tests/b1.rlp` and `tests/b2.rlp`?

blockchain/types/transaction_test.go
ian0371: To be safer, how bout `params.Rules{IsIstanbul: true, IsShanghai: true, IsPrague: true}`

hyunsooda: @blukat29
The numbers were exactly fitted for `Prague` HF test.
```
nonZeroGas = params.TxDataNonZeroGasEIP2028
zeroGas = params.TxDataZeroGas
```

Can you explain in more detail how adding `IsIstanbul` contributes to the context safe?


@ian0371 The `expectedGas*` numbers did not consider the `Shanghai` value. Do you mean to add this case additionally?

blockchain/types/transaction_test.go
hyunsooda: I didn't touch the `expecteGas4`. Additionally, what does this suggestion mean? Fourth parameter is not optional.

blockchain/types/tx_internal_data.go
hyunsooda: Some tx types have predefined gas values, so as I understand it, `IntrinsicGasPayload` is used to retain the gas value at its call site.

hyunsooda:
@ian0371 The file `https://github.com/kaiachain/kaia/blob/dev/tests/hard_fork_test.go#L52`  has a test which read previously stored test files (`genesis.json`, `b1.rlp` and `b2.rlp`). The `Istanbul` HF has been absent to `genesis.json`. So, I add it first into it and regenerate `b1.rlp` and `b2.rlp`. This modification was necessary because the semantic of `IntrinsicGasPayload()` has been changed in this PR, from unconditionally payment of 100gas per data to EIP2028. Thus, without this change, the test was failed because it reaches to [here](https://github.com/kaiachain/kaia/pull/107/files#diff-36960dea371fb7a7722f48b0905c890030f61801a29c132b5ade154357f0e1dbR575-R576)

blockchain/types/transaction_test.go
ian0371: It's because if a block is Prague, it implies it's Istanbul and Shanghai as well.
In other words, a block can't be Prague if it's not Istanbul and not Shanghai.

Hence, these are only valid rules:
```
params.Rules{IsIstanbul: true}
params.Rules{IsIstanbul: true, IsShanghai: true}
params.Rules{IsIstanbul: true, IsShanghai: true, IsPrague: true}
```

hyunsooda: That makes sense. Then, do you suggest refining the `expectGas*` values? FYI. Without changing this, the test fails if other HFs are added, as you mentioned.

ian0371:
@hyunsooda Got it. Although it seems the new `genesis.json` also doesn't have `istanbulCompatibleBlock`. Is this ok?

blockchain/types/transaction_test.go
blukat29: Then please adjust the testcases to reflect the reality

blockchain/types/transaction_test.go
blukat29: Oh I misread the code. Nevermind.

blockchain/types/tx_internal_data.go
blukat29: Because the IntrinsicGas() function is calculating a sum of three components,
```
IntrinsicGas = (TxGasContractCreation or TxGas) + PayloadGas + (AcesssLisstGas)
```
I thought we can make this function look like roughly
```
gas = TxGasContractCreation or TxGas
gas += PayloadGas
gas += AccessListGas
```
But for the uint64 overflow checks, now I see that it's inevitable to pass in the `gas` value into `IntrinsicGasPayload` function.

hyunsooda:
I thought that I've modified it, but did not actually and tried to understand how it was twisted. 

1. The `Istanbul` or `Shanghai` fork number had not been specified to `genesis.json`.
2. `IntrinsicGasPayload()` unconditionally have been subtracting 100gas per data.
3. both `b1.rlp` and `b2.rlp` were generated 5y ago (based on the current logic (logically similar))


Given the three facts above, `IntrinsicGasPayload()` should subtract 100gas per data, but the PR changes it. It was the root cause and now I reverted `genesis.json` which is not necessarily one to be modified. Instead, I regenerated `b1.rlp` and `b2.rlp` so that the generated files would take the current behavior of `IntrinsicGasPayload()`.

blockchain/types/transaction_test.go
hyunsooda: Updated https://github.com/kaiachain/kaia/pull/107/commits/c94555e18d200cd1eafbf5beec078c2991aff76d

ian0371:
> `IntrinsicGasPayload()` unconditionally have been subtracting 100gas per data.

Could you show me where?

hyunsooda:
@ian0371 blockchain/types/tx_internal_data_account_creation.go

ian0371:
I don't see any "subtraction" in the link or in this PR. What am I missing?

In addition, do you mean that `IntrinsicGas()` behavior is different now compared? I'm thinking that `b1.rlp` is a legacy block, and this PR's `IntrinsicGas()` cannot handle it (it's a pre-Istanbul case, right?). Doesn't this mean that it would break the backward compatibility? Sorry but I'm a little confused.

hyunsooda:
I'm sorry. `Subtraction` seems corrupt you. Let me correct this word to `payment`.

> In addition, do you mean that IntrinsicGas() behavior is different now compared? I'm thinking that b1.rlp is a legacy block, and this PR's IntrinsicGas() cannot handle it (it's a pre-Istanbul case, right?). 

Yes.


> Doesn't this mean that it would break the backward compatibility?

Yes, it calculates differently. Can you refer to the first problem in the PR description? Directly calling the `IntrinsicGasPayload()` does not check `Istanbul` HF and directly pays 100gas per data. However, it is different if the `IntriscGas()` is called. This checks HF state and calls `IntrinsicGasPayload()` if `Istanbul` HF is on. 

Now I think that I was wrong. This direction was started from the thought that want to resolve the wrong behavior implemented initially. But for the backward compatibility, I should not touch them damn.. It's inevitable to leave as it is. If this is confirmed, let me rework to keep the backward compatability.

ian0371:
Ohh I see.
I think calling `IntrinsicGasPayload()` in the first place was a very big mistake.... only `IntrinsicGas()` must've been used..
It feels like a headache now to refactor it

ian0371:
How about we leave the ingrate function as in the PR, unify the gas calculation after prague, and adjust (subtract) gas from the callers pre-Prague for those who call `IntrinsicGasPayload()` directly?

as-is
```
gas = IntrinsicGasPayload()
```

to-be
```
gas = IntrinsicGas()
if prePrague:
	gas -= 100
```

Let's discuss with @blukat29 in the next scrum.

hyunsooda:
@ian0371 @blukat29 Simplified the logic by inserting EIP-2028 calculation into `IntrinsicGasPayload()` and updated the PR description.

blockchain/types/tx_internal_data.go
```diff
@@ -553,16 +553,35 @@ func toWordSize(size uint64) uint64 {
 }
 
 func IntrinsicGasPayload(gas uint64, data []byte, isContractCreation bool, rules params.Rules) (uint64, error) {
-	// Bump the required gas by the amount of transactional data
+	// // Bump the required gas by the amount of transactional data
```
ian0371, line 556:
```suggestion
	// Bump the required gas by the amount of transactional data
```

ian0371 (commented):
Could you add the discussed comments? (for `IntrinsicGas`)

blockchain/types/tx_internal_data.go
hyunsooda: Uhm, I thought that I had added the comment requests. Can you let me know which statements were absent for comments in `IntrinsicGas()` header?

ian0371: I interpreted the meeting minutes like this:

```
func IntrinsicGas() {
	...
	if r.IsIstanbul {
		// tx types 1,2 only exist after istanbul; so they take this path.
		// tx types 8+ take this path as well.
		gasPayloadWithGas, err = IntrinsicGasPayload(gas, data, contractCreation, r)
	} else {
		// only for tx type 0 before istanbul.
		gasPayloadWithGas, err = IntrinsicGasPayloadLegacy(gas, data)
	}
}

// Klaytn-TxTypes since genesis, and EthTxTypes since istanbul use this.
func IntrinsicGasPayload() {...}

// Eth-TxTypes before istanbul use this. Before istanbul, there's only tx type 0.
func IntrinsicGasPayloadLegacy() {...}
```

ian0371: > Could you add the discussed comments? (for IntrinsicGas)

The above [comment](https://github.com/kaiachain/kaia/pull/107#discussion_r1798726357) is about this.


This [suggested change](https://github.com/kaiachain/kaia/pull/107#discussion_r1798699207) is an orthogonal issue. (`// //` should be `//`)

hyunsooda: > This https://github.com/kaiachain/kaia/pull/107#discussion_r1798699207 is an orthogonal issue. (// // should be //)

Yes I know. Just wanted to proceed conversation in a thread. Thank you for your kindness.


The comment below seems the opposite case. For now, I added it. Do you see this is a redundant one?
```
// tx types 1,2 only exist after istanbul; so they take this path.
		// tx types 8+ take this path as well.
```

ian0371: I'm ok without it

ian0371 (approved):

PR #114, opened 2024-10-22
kaiax/staking: Implement Preload feature

blukat29:
## Proposed changes

- Add the "Preload" feature in the kaiax/staking. Serves the same purpose with the existing PreloadStakingInfo.
- Usage
	```
	mStaking.AllocPreloadRef() -> RefId
	mStaking.PreloadFromState(RefId, StateDB)
	mStaking.FreePreloadRef(RefId)
	```
- Data structure
	```
	PreloadBuffer: [StateRoot] -> { StakingInfo, [RefId] -> struct{} }
	
	Suppose there are two concurrent state reexec threads.
	First thread is assigned RefId=1, and reexec through three blocks of which state roots are H1, H2, H3.
	Second thread is assigned RefId=2, and reexec through three blocks of which state roots are H3, H4, H5.
	The resulting SideStates looks like this:
	
	H1 -> { S1, [1] }
	H2 -> { S2, [1] }
	H3 -> { S3, [1,2] }
	H4 -> { S4, [2] }
	H5 -> { S5, [2] }
	```
- The GetStakingInfo() also refers to the PreloadBuffer.

The SideState feature has two usages.
- Before CreateSnapshot in `cn.New()` -- this usage will be removed as we introduce kaiax/gov and kaiax/valset. So I simply modified the existing reward.PreloadStakingInfo as a temporary fix.
- In `StateAtBlock` - properly using StakingModule.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -396,12 +396,12 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	if headers, err = cn.Engine().GetKaiaHeadersForSnapshotApply(cn.blockchain, currBlock.NumberU64(), currBlock.Hash(), nil); err != nil {
 		logger.Error("Failed to get headers to apply", "err", err)
 	} else {
-		sideStateRef, err := reward.PreloadStakingInfo(headers, mStaking)
+		preloadRef, err := reward.PreloadStakingInfo(headers, mStaking)
 		if err != nil {
 			logger.Error("Preload staking info failed", "err", err)
 		}
 		defer func() {
-			mStaking.FreeSideStateRef(sideStateRef)
+			mStaking.FreePreloadRef(preloadRef)
 		}()
 	}
 	if err := cn.Engine().CreateSnapshot(cn.blockchain, currBlock.NumberU64(), currBlock.Hash(), headers); err != nil {
--- a/node/cn/backend.go
+++ b/node/cn/backend.go
@@ -384,16 +384,22 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 		currBlock = cn.blockchain.CurrentBlock()
 		headers   []*types.Header
 	)
+	// Temporarily supply blockchain for `Finalize`, staking module for `snapshot`.
+	cn.blockchain.Engine().(consensus.Istanbul).SetChain(cn.blockchain)
+	mStaking := staking_impl.NewStakingModule()
+	mStaking.Init(&staking_impl.InitOpts{
+		ChainKv:     cn.chainDB.GetMiscDB(),
+		ChainConfig: cn.chainConfig,
+		Chain:       cn.blockchain,
+	})
+	cn.blockchain.Engine().(consensus.Istanbul).RegisterStakingModule(mStaking)
 	if headers, err = cn.Engine().GetKaiaHeadersForSnapshotApply(cn.blockchain, currBlock.NumberU64(), currBlock.Hash(), nil); err != nil {
 		logger.Error("Failed to get headers to apply", "err", err)
 	} else {
-		// Temporarily supply blockchain for `Finalize`.
-		cn.blockchain.Engine().(consensus.Istanbul).SetChain(cn.blockchain)
 		preloadNums, err := reward.PreloadStakingInfo(headers)
 		if err != nil {
 			logger.Error("Preload staking info failed", "err", err)
 		}
-		cn.blockchain.Engine().(consensus.Istanbul).SetChain(nil)
 		defer func() {
 			for _, num := range preloadNums {
 				reward.UnloadStakingInfo(num)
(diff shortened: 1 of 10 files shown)


blukat29:
@hyeonLewis PTAL. Without this PR, the CreateSnapshot at node startup might break.

hyeonLewis (commented):
This is just a question: Is there any reason to store stateDB, not stakingInfo directly same as before?

reward/staking_manager.go
```diff
@@ -279,51 +280,47 @@ func PreloadStakingInfo(headers []*types.Header) ([]uint64, error) {
...
-		}
+		stakingModule.AddSideState(sideStateRef, statedb)
 		if current.NumberU64() == target {
 			break
 		}
 		// Retrieve the next block to regenerate and process it
 		next := current.NumberU64() + 1
 		if current = bc.GetBlockByNumber(next); current == nil {
-			return nil, fmt.Errorf("block #%d not found", next)
+			return 0, fmt.Errorf("block #%d not found", next)
```
hyeonLewis, line 311: Same issue here as state_accessor.

node/cn/state_accessor.go
```diff
@@ -162,11 +156,8 @@ func (cn *CN) stateAtBlock(block *types.Block, reexec uint64, base *state.StateD
 		if cn.config.DisableUnsafeDebug && time.Since(start) > cn.config.StateRegenerationTimeLimit {
 			return nil, nil, fmt.Errorf("this request has queried old states too long since it exceeds the state regeneration time limit(%s)", cn.config.StateRegenerationTimeLimit.String())
 		}
-		// Preload StakingInfo from the current block and state. Needed for next block's engine.Finalize() post-Kaia.
-		preloadedStakingBlockNums = append(preloadedStakingBlockNums, current.NumberU64())
-		if err := reward.PreloadStakingInfoWithState(current.Header(), statedb); err != nil {
-			return nil, nil, fmt.Errorf("preloading staking info from block %d failed: %v", current.NumberU64(), err)
-		}
+		// Make StakingModule remember the current block state. Needed for next block's engine.Finalize() post-Kaia.
+		cn.stakingModule.AddSideState(sideStateRef, statedb)
```
hyeonLewis, line 160: If any error occurs below, `sideStateRef` can't be released. It seems previous implementation also had same issue.

blukat29: Let me work on it.

blukat29:
@hyeonLewis I thought the Preload feature may be inaccurate because it stores StakingInfo keyed by block numbers. When there is a bad block then block rewind, the block state may differ on the same number. In the process of moving from number-keyed to roothash-keyed, I got to store the state.
But on a second thought, while I will keep the roothash-key, but resource-wise it's better to store the StakingInfo than the whole StateDB.

blukat29:
@hyeonLewis Modified to store pre-calculated StakingInfo instead and tests pass. Let me polish it and fix the "can't be released" issues.

node/cn/state_accessor.go
ian0371: suggestion: use named return for error, and a defer func checks if it is not nil, in which case free the ref.

reward/staking_manager.go
blukat29: Fixed in 22ea3e7c.

node/cn/state_accessor.go
blukat29: Fixed in 22ea3e7c.

By the way, Deference() is also missing in some error cases. But because geth has the same issue I won't fix that here. https://github.com/ethereum/go-ethereum/blob/v1.14.11/eth/state_accessor.go

blukat29: cc @ian0371 @yoomee1313

ian0371: The state root is always dereferenced regardless of error. If no error, the caller dereferences it. Other cases is done here: https://github.com/kaiachain/kaia/blob/1fe135d9e12ab4ec8057aa5e71b6423353eb7834/node/cn/state_accessor.go#L188-L190

So our case is different. The releaser won't be run by the caller when there's error, so the callee must call it in defer function.

node/cn/backend.go
```diff
@@ -384,25 +384,31 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
...
 	} else {
-		// Temporarily supply blockchain for `Finalize`.
-		cn.blockchain.Engine().(consensus.Istanbul).SetChain(cn.blockchain)
-		preloadNums, err := reward.PreloadStakingInfo(headers)
+		preloadRef, err := reward.PreloadStakingInfo(headers, mStaking)
 		if err != nil {
 			logger.Error("Preload staking info failed", "err", err)
 		}
-		cn.blockchain.Engine().(consensus.Istanbul).SetChain(nil)
 		defer func() {
```
ian0371, line 403: We should move this defer func to L399

node/cn/state_accessor.go
blukat29: Oh, I see. the `root` is quite short-lived.

The release() function doing the FreePreloadRef was an attempt to replace `reward.PreloadStakingInfo` with `stateAtBlock` so we have only one implementation. But I later learned that merging them was non-trivial.

I'll go back to the original way of calling `defer FreePreloadRef()` inside stateAtBlock.

ian0371 (approved):

hyeonLewis (approved):

PR #118, opened 2024-10-30
kaiax/supply: Add SupplyModule

blukat29:
## Proposed changes

- Describe your changes to communicate to the maintainers why we should accept this pull request.
- If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kaiax/supply/impl/checkpoint.go
```diff
@@ -0,0 +1,148 @@
...
+		BurntFee: new(big.Int).Set(sc.BurntFee),
+	}
+}
+
+func nearestCheckpointInterval(num uint64) uint64 {
+	return num - (num % checkpointInterval)
+}
+
+// loadLastCheckpoint loads the last supply checkpoint from the database.
+func (s *SupplyModule) loadLastCheckpoint() error {
```
ian0371, line 50: Perhaps move SupplyModule methods to `init.go`?

kaiax/supply/impl/getter_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"math/big"
+
+	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/kaiax/supply"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test individual getters
+func (s *SupplyTestSuite) TestFromState() {
```
ian0371, line 29: These tests are not run. Please follow the format [here](https://pkg.go.dev/testing)

kaiax/supply/impl/checkpoint.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"math/big"
+	"sync/atomic"
+	"time"
+
+	"github.com/kaiachain/kaia/kaiax/supply"
+)
+
+var (
+	checkpointInterval    = uint64(128)
```
ian0371, line 28: This gives a false impression that checkpoint is updated every 128 blocks; i.e., `getCheckpoint(128)=...=getCheckpoint(255)`.

blukat29: Sure. Will `checkpointCommitInterval` clarify it?

ian0371: That's better. btw, why did we name it checkpoint in the first place?
Maybe we're just saving TotalSupply in DB but with most fields removed just to reduce storage size.

blukat29: I think we were coming up with a different name to force re-accumulation after fixing a rewind-related bug. But AccReward sounds more accurate for the in-memory structure that increments every block. PTAL 8e14a9a.

kaiax/supply/impl/checkpoint.go
blukat29: Let me move it after we're settle with the SupplyCheckpoint naming.

blukat29: Moved into init.go and getter.go according to their usage.

kaiax/supply/impl/checkpoint.go
blukat29: Modified to ReadAccReward(), WriteAccReward(), lastAccReward, etc.

kaiax/supply/impl/execution.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (s *SupplyModule) RewindTo(newBlock *types.Block) {
+	// Soft reset to the nearest checkpoint interval less than the new block number,
+	// so that the next accumulation will start below the rewound block.
+	newLastNum := nearestCheckpointNumber(newBlock.NumberU64())
+	WriteLastAccRewardNumber(s.ChainKv, newLastNum)
+}
+
+func (s *SupplyModule) RewindDelete(hash common.Hash, num uint64) {
+	if num%checkpointInterval != 0 {
```
ian0371, line 51: Delete all accReward after num?

kaiax/supply/impl/execution.go
```diff
@@ -0,0 +1,54 @@
...
+		DeleteAccReward(s.ChainKv, num)
```
ian0371, line 52:
```suggestion
		DeleteAccReward(s.ChainKv, num)
		WriteLastAccRewardNumber(s.ChainKv, num)
```

kaiax/supply/impl/init.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (s *SupplyModule) Stop() {
+	atomic.StoreUint32(&s.quit, 1)
+	s.quitCh <- struct{}{}
+	s.wg.Wait()
+}
+
+// loadLastCheckpoint loads the last supply checkpoint from the database.
+func (s *SupplyModule) loadLastCheckpoint() error {
```
ian0371, line 118:
```suggestion
// loadLastAccReward loads the last supply checkpoint from the database.
func (s *SupplyModule) loadLastAccReward() error {
```

kaiax/supply/impl/init.go
```diff
@@ -0,0 +1,188 @@
...
+	mu            sync.RWMutex
+	lastAccNum    uint64            // Last AccReward number
+	lastAccReward *supply.AccReward // Last AccReward
+
+	// Stops long-running tasks.
+	quit   uint32         // stops the synchronous loop in accumulateCheckpoint
+	quitCh chan struct{}  // stops the goroutine in select loop
+	wg     sync.WaitGroup // wait for the goroutine to finish
+
+	supplyCache *lru.ARCCache // (number uint64) -> (totalSupply *TotalSupply)
```
hyeonLewis, line 71: What's the purpose of `supplyCache`? It only stores `totalSupply` when there's API request, which means full node can't get the `totalSupply` from cache if hadn't. So would we expect the node operator to call `GetTotalSupply` for every block to cache the data? If not, cache won't be filled for block after `TrieInMemory` in case of full node.

Also, I think introducing `accRewardCache` can be effective.

kaiax/supply/impl/execution.go
blukat29: LastAccRewardNumber should be overwritten in RewindTo which is called by updateFn.

In setHeadBeyondRoot (https://github.com/kaiachain/kaia/blob/9ec35a0050193b17c5f2dfd0f2acdc070779ff01/blockchain/blockchain.go#L663-L675), it's either `updateFn` or `updateFn+delFun`.

kaiax/supply/impl/init.go
blukat29: Fixed by 134b5080

kaiax/supply/impl/init.go
blukat29: It is for caching the API result. I found out that caching accReward won't help full nodes anyway because CanonicalBurn data is unavailable after TriesInMemory. So I removed accRewardCache.

blukat29: Having accRewardCache might speed up GetTotalSupply(128k + r) cases, but even without cache it only iterates up to 128 headers which won't take too long.

hyeonLewis: Ok. I also think 127 iterations won't be matter.

I just wanted to confirm that it was intended and it seems so. And for full node, it can still serve blocks with `num%128==0`, which only makes acceptable accuracy errors.

blukat29: You're right. Full nodes can serve blocks with `num%128=0` in which case there exists ReadAccReward(). Because DB has its own cache, we don't need separate cache.

hyeonLewis (commented):
Will the following legacy lines be deleted in another PR?

https://github.com/kaiachain/kaia/blob/9ec35a0050193b17c5f2dfd0f2acdc070779ff01/blockchain/blockchain.go#L627

https://github.com/kaiachain/kaia/blob/9ec35a0050193b17c5f2dfd0f2acdc070779ff01/blockchain/blockchain.go#L647

kaiax/supply/impl/execution.go
hyeonLewis, line 51: Is it feasible to delete `accReward` at non-interval blocks? We only store `accReward` at the interval block. Am I missing something?

blukat29: Oops, fixed in cacb1cb9.

kaiax/supply/impl/execution.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return err
+		}
+		s.lastAccNum = newNum
+		s.lastAccReward = newCheckpoint
+	}
+	return nil
+}
+
+func (s *SupplyModule) RewindTo(newBlock *types.Block) {
```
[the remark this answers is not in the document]
hyunsooda: Then how about adding a branch and debug log if the block is nil without function sig modification?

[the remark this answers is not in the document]
blukat29: RewindTo is called from exactly one place, in setHeadBeyondRoot in which case the newBlock is decisively not nil. Let me skip nil check here.

hyeonLewis:
> Will the following legacy lines be deleted in another PR?
> 
> https://github.com/kaiachain/kaia/blob/9ec35a0050193b17c5f2dfd0f2acdc070779ff01/blockchain/blockchain.go#L627
> 
> https://github.com/kaiachain/kaia/blob/9ec35a0050193b17c5f2dfd0f2acdc070779ff01/blockchain/blockchain.go#L647

@blukat29 Could you check my previous comment?

blukat29:
@hyeonLewis I deleted them in https://github.com/kaiachain/kaia/commit/cacb1cb9f3dfff5c183040c02bce54f3d87a5c98 and https://github.com/kaiachain/kaia/pull/118/commits/f7071e941a1461163dc330e5db81f3dc9f96d5fc

hyeonLewis (approved):

ian0371 (approved):

Issue #106, opened 2024-10-10
Introduce EIP-7702 (Set EOA account code)


Mdaiki0730:
**Is your request related to a problem? Please describe.**
Kaia has an account abstraction in the form of ERC-4337. 
We can receive the benefits of AA through this, but if you already have an EOA account, you will need to create a new contract account and transfer assets.

**Describe the solution you'd like**
Introducing EIP-7702 can solve this problem.
EIP-7702 allows you to treat an EOA like a contract account by setting a contract in the code field of the EOA.
A new `SET_CODE_TX_TYPE` is going to be defined, which will work as follows:
![EIP7702_detail_process](https://github.com/user-attachments/assets/19f07b06-2853-4500-9d74-0bd617183a2e)

**Describe alternatives you've considered**
- EIP-3074
  - EIP-7702 was additionally proposed as an alternative.
- RIP-7560
  - Because this is a proposal for Rollup, it seems a bit difficult to introduce it to L1 right away.

**Additional context**
The following changes are expected:

##### Add Objects
- Add SetCodeTxType, which implements TxInternalData. This will be treated as the EthTypedTransaction.
- Define an AuthorizationList([]Authorization) for the authorization tuple.

##### Changes
- /api
  - Change the interface of api to correspond to `SetCodeTxType`
- /blockchain
  - state_transition
    - Add the main code for adding set code and authority to `StateTransition.TransitionDb`
  - types/tx_internal_data
    - Add `TxTypeSetCode`
    - Extende the `StateDB` interface (for validation)
  - tx_pool
    - There may be changes to `validateTx`
  - state/statedb
    - Implement `ResolveCode` in `StateDB`(also resolveStateObject)
  - vm
    - Add 7702 to eips in VM
    - Changed `GetCode` to `ResolveCode` in evm and instructions (If code is requested)
    - Updated interface of `StateDB`
    - Call `enable7702` in jump_table
    - Define gas around EIP7702 in `operations_acl`
- /params
  - Define of gas fee for authorization

Mdaiki0730:
We have found some additional changes that we need to make.
#### Points
- Since [EOA](https://docs.kaia.io/learn/accounts/#externally-owned-accounts-eoas-) does not have a code hash field, how do we retain the EOA after the setcode?
- A change needs to be made to Kaia's Account RPC API related to setcode.

Mdaiki0730:
@kaiachain/core
[...]

PR #105, opened 2024-10-10
params: Declare Prague hardfork configuration

hyunsooda:
## Proposed changes

Followed the previous process from:
- Cancun: https://github.com/klaytn/klaytn/pull/1969/files
- Kaia(Dragon): https://github.com/klaytn/klaytn/pull/2114/files
- Geth(Prague): https://github.com/ethereum/go-ethereum/pull/26481/files

Please let me know if anything is missing.

```shell
> kaia.getChainConfig()
{
  ...
  ...
  istanbulCompatibleBlock: 0,
  kaiaCompatibleBlock: 0,
  kip103CompatibleBlock: 0,
  kip103ContractAddress: "0x0000000000000000000000000000000000000000",
  kip160CompatibleBlock: 0,
  kip160ContractAddress: "0x0000000000000000000000000000000000000000",
  koreCompatibleBlock: 0,
  londonCompatibleBlock: 0,
  magmaCompatibleBlock: 0,
  pragueCompatibleBlock: 0,
  randaoCompatibleBlock: 0,
  shanghaiCompatibleBlock: 0,
}
```

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/kaiachain/kaia/blob/main/CONTRIBUTING.md) doc
- [ ] I have read the [CLA](https://gist.github.com/kaiachain-dev/bbf65cc330275c057463c4c94ce787a6) and signed by comment `I have read the CLA Document and I hereby sign the CLA` in first time contribute
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #119, opened 2024-10-31
blockchain: Add PragueSigner and EIP-7702 SetCode transaction

Mdaiki0730:
## Proposed changes

- Define the `TxTypeEthereumSetCode` to set the code for EOA.
- Serialization changes with the addition of `TxTypeEthereumSetCode`
- Add PragueSigner
  - Although it has been newly created, the process is no different from that of existing signers.
- Add test case for gas cost calculation of `TxTypeEthereumSetCode`
- Add set code processing to state_transition
  - This isn't perfect yet, there is a PR to add a code hash field to EOA that will get it working perfectly, see the comments.
- This PR is adapting the ongoing go-eth PR for kaia.
  - https://github.com/ethereum/go-ethereum/pull/30078

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

- [x] I have added necessary documentation (if appropriate)

## Related issues

- #106 

## Further comments

The responsibilities of this PR are to add SetCodeTxType and PragueSigner, and to add the expected processing of SetCodeTxType to state_transition (Functional testing is not possible yet because the code field has not been added to EOA).
We are planning to submit three additional PRs later.
- Add code hash field to EOA
- Add a method to set code for EOA and functional test of SetCodeTxType
- Integrate test

### DIFF
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -1413,7 +1413,7 @@ func EthDoCall(ctx context.Context, b Backend, args EthTransactionArgs, blockNrO
 	} else {
 		baseFee = new(big.Int).SetUint64(params.ZeroBaseFee)
 	}
-	intrinsicGas, err := types.IntrinsicGas(args.data(), nil, args.To == nil, b.ChainConfig().Rules(header.Number))
+	intrinsicGas, err := types.IntrinsicGas(args.data(), nil, nil, args.To == nil, b.ChainConfig().Rules(header.Number))
 	if err != nil {
 		return nil, err
 	}
@@ -1570,7 +1570,7 @@ func AccessList(ctx context.Context, b Backend, blockNrOrHash rpc.BlockNumberOrH
 	precompiles := vm.ActivePrecompiles(rules)
 
 	toMsg := func() (*types.Transaction, error) {
-		intrinsicGas, err := types.IntrinsicGas(args.data(), nil, args.To == nil, rules)
+		intrinsicGas, err := types.IntrinsicGas(args.data(), nil, nil, args.To == nil, rules)
 		if err != nil {
 			return nil, err
 		}
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -771,7 +771,7 @@ type ethTxJSON struct {
 	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
 	AccessList *types.AccessList `json:"accessList,omitempty"`
 
-	// Authorization list transaction (SetCodeTx) fields:
+	// Set code transaction fields:
 	AuthorizationList *types.AuthorizationList `json:"authorizationList,omitempty"`
 
 	// Only used for encoding:
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -722,25 +722,26 @@ func (api *EthereumAPI) GetBlockTransactionCountByHash(ctx context.Context, bloc
 // RPCTransaction in go-ethereum has been renamed to EthRPCTransaction.
 // RPCTransaction is defined in go-ethereum's internal package, so RPCTransaction is redefined here as EthRPCTransaction.
 type EthRPCTransaction struct {
-	BlockHash        *common.Hash      `json:"blockHash"`
-	BlockNumber      *hexutil.Big      `json:"blockNumber"`
-	From             common.Address    `json:"from"`
-	Gas              hexutil.Uint64    `json:"gas"`
-	GasPrice         *hexutil.Big      `json:"gasPrice"`
-	GasFeeCap        *hexutil.Big      `json:"maxFeePerGas,omitempty"`
-	GasTipCap        *hexutil.Big      `json:"maxPriorityFeePerGas,omitempty"`
-	Hash             common.Hash       `json:"hash"`
-	Input            hexutil.Bytes     `json:"input"`
-	Nonce            hexutil.Uint64    `json:"nonce"`
-	To               *common.Address   `json:"to"`
-	TransactionIndex *hexutil.Uint64   `json:"transactionIndex"`
-	Value            *hexutil.Big      `json:"value"`
-	Type             hexutil.Uint64    `json:"type"`
-	Accesses         *types.AccessList `json:"accessList,omitempty"`
-	ChainID          *hexutil.Big      `json:"chainId,omitempty"`
-	V                *hexutil.Big      `json:"v"`
-	R                *hexutil.Big      `json:"r"`
-	S                *hexutil.Big      `json:"s"`
+	BlockHash        *common.Hash             `json:"blockHash"`
+	BlockNumber      *hexutil.Big             `json:"blockNumber"`
+	From             common.Address           `json:"from"`
+	Gas              hexutil.Uint64           `json:"gas"`
+	GasPrice         *hexutil.Big             `json:"gasPrice"`
+	GasFeeCap        *hexutil.Big             `json:"maxFeePerGas,omitempty"`
+	GasTipCap        *hexutil.Big             `json:"maxPriorityFeePerGas,omitempty"`
+	Hash             common.Hash              `json:"hash"`
+	Input            hexutil.Bytes            `json:"input"`
+	Nonce            hexutil.Uint64           `json:"nonce"`
+	To               *common.Address          `json:"to"`
+	TransactionIndex *hexutil.Uint64          `json:"transactionIndex"`
+	Value            *hexutil.Big             `json:"value"`
+	Type             hexutil.Uint64           `json:"type"`
+	Accesses         *types.AccessList        `json:"accessList,omitempty"`
+	Authorizations   *types.AuthorizationList `json:"authorizationList,omitempty"`
+	ChainID          *hexutil.Big             `json:"chainId,omitempty"`
+	V                *hexutil.Big             `json:"v"`
+	R                *hexutil.Big             `json:"r"`
+	S                *hexutil.Big             `json:"s"`
 }
 
 // ethTxJSON is the JSON representation of Ethereum transaction.
@@ -767,8 +768,9 @@ type ethTxJSON struct {
 	To                   *common.Address `json:"to"`
 
 	// Access list transaction fields:
-	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
-	AccessList *types.AccessList `json:"accessList,omitempty"`
+	ChainID           *hexutil.Big             `json:"chainId,omitempty"`
+	AccessList        *types.AccessList        `json:"accessList,omitempty"`
+	AuthorizationList *types.AuthorizationList `json:"authorizationList,omitempty"`
 
 	// Only used for encoding:
 	Hash common.Hash `json:"hash"`
@@ -863,6 +865,20 @@ func newEthRPCTransaction(block *types.Block, tx *types.Transaction, blockHash c
 			// transaction is not processed yet
 			result.GasPrice = (*hexutil.Big)(tx.EffectiveGasPrice(nil, nil))
 		}
+	case types.TxTypeEthereumSetCode:
+		al := tx.AccessList()
+		aul := tx.AuthorizationList()
+		result.Accesses = &al
+		result.Authorizations = &aul
+		result.ChainID = (*hexutil.Big)(tx.ChainId())
+		result.GasFeeCap = (*hexutil.Big)(tx.GasFeeCap())
+		result.GasTipCap = (*hexutil.Big)(tx.GasTipCap())
+		if block != nil {
+			result.GasPrice = (*hexutil.Big)(tx.EffectiveGasPrice(block.Header(), config))
+		} else {
+			// transaction is not processed yet
+			result.GasPrice = (*hexutil.Big)(tx.EffectiveGasPrice(nil, nil))
+		}
 	}
 	return result
 }
@@ -915,6 +931,14 @@ func formatTxToEthTxJSON(tx *types.Transaction) *ethTxJSON {
 	case types.TxTypeEthereumDynamicFee:
 		al := tx.AccessList()
 		enc.AccessList = &al
+		enc.ChainID = (*hexutil.Big)(tx.ChainId())
+		enc.MaxFeePerGas = (*hexutil.Big)(tx.GasFeeCap())
+		enc.MaxPriorityFeePerGas = (*hexutil.Big)(tx.GasTipCap())
+	case types.TxTypeEthereumSetCode:
+		al := tx.AccessList()
+		aul := tx.AuthorizationList()
+		enc.AccessList = &al
+		enc.AuthorizationList = &aul
 		enc.ChainID = (*hexutil.Big)(tx.ChainId())
 		enc.MaxFeePerGas = (*hexutil.Big)(tx.GasFeeCap())
 		enc.MaxPriorityFeePerGas = (*hexutil.Big)(tx.GasTipCap())
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -770,6 +770,8 @@ type ethTxJSON struct {
 	// Access list transaction fields:
 	ChainID           *hexutil.Big             `json:"chainId,omitempty"`
 	AccessList        *types.AccessList        `json:"accessList,omitempty"`
+
+	// Authorization list transaction (SetCodeTx) fields:
 	AuthorizationList *types.AuthorizationList `json:"authorizationList,omitempty"`
 
 	// Only used for encoding:
--- a/api/api_ethereum.go
+++ b/api/api_ethereum.go
@@ -768,8 +768,8 @@ type ethTxJSON struct {
 	To                   *common.Address `json:"to"`
 
 	// Access list transaction fields:
-	ChainID           *hexutil.Big             `json:"chainId,omitempty"`
-	AccessList        *types.AccessList        `json:"accessList,omitempty"`
+	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
+	AccessList *types.AccessList `json:"accessList,omitempty"`
 
 	// Authorization list transaction (SetCodeTx) fields:
 	AuthorizationList *types.AuthorizationList `json:"authorizationList,omitempty"`
--- a/blockchain/state_transition.go
+++ b/blockchain/state_transition.go
@@ -399,13 +399,16 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 			if exists := st.state.Exist(authority); exists {
 				st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
 			}
-			st.state.SetNonce(authority, auth.Nonce+1)
+			st.state.IncNonce(authority)
 			delegation := types.AddressToDelegation(auth.Address)
 			if auth.Address == common.ZeroAddress {
 				// If the delegation is for the zero address, completely clear all
 				// delegations from the account.
 				delegation = []byte{}
 			}
+			// TODO: Change to a method that adds code to EOA
+			// This is currently expected to be an error. This will be addressed in a PR that adds a code hash field to the EOA.
+			// https://github.com/kaiachain/kaia/blob/v1.0.3/blockchain/state/state_object.go#L542
 			st.state.SetCode(authority, delegation)
 
 			// Usually the transaction destination and delegation target are added to
--- a/blockchain/state_transition.go
+++ b/blockchain/state_transition.go
@@ -410,7 +410,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 				// delegations from the account.
 				delegation = []byte{}
 			}
-			// TODO: Change to a method that adds code to EOA
+			// TODO-7702: Change to a method that adds code to EOA
 			// This is currently expected to be an error. This will be addressed in a PR that adds a code hash field to the EOA.
 			// https://github.com/kaiachain/kaia/blob/v1.0.3/blockchain/state/state_object.go#L542
 			st.state.SetCode(authority, delegation)
--- a/blockchain/state_transition.go
+++ b/blockchain/state_transition.go
@@ -123,6 +123,7 @@ type Message interface {
 	Execute(vm types.VM, stateDB types.StateDB, currentBlockNumber uint64, gas uint64, value *big.Int) ([]byte, uint64, error)
 
 	AccessList() types.AccessList
+	AuthorizationList() types.AuthorizationList
 }
 
 // ExecutionResult includes all output after executing given evm
@@ -291,6 +292,16 @@ func (st *StateTransition) preCheck() error {
 			return ErrNonceTooLow
 		}
 	}
+
+	// Check that EIP-7702 authorization list signatures are well formed.
+	for i, auth := range st.msg.AuthorizationList() {
+		switch {
+		case auth.R.BitLen() > 256:
+			return fmt.Errorf("%w: address %v, authorization %d", ErrAuthSignatureVeryHigh, st.msg.ValidatedSender(), i)
+		case auth.S.BitLen() > 256:
+			return fmt.Errorf("%w: address %v, authorization %d", ErrAuthSignatureVeryHigh, st.msg.ValidatedSender(), i)
+		}
+	}
 	return st.buyGas()
 }
 
@@ -347,10 +358,64 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 
 	rules := st.evm.ChainConfig().Rules(st.evm.Context.BlockNumber)
 
+	// Verify authorization list is not empty.
+	if msg.AuthorizationList() != nil && len(msg.AuthorizationList()) == 0 {
+		return nil, fmt.Errorf("%w: address %v", ErrEmptyAuthList, msg.ValidatedSender())
+	}
+
 	// Execute the preparatory steps for state transition which includes:
 	// - prepare accessList(post-berlin)
 	// - reset transient storage(eip 1153)
 	st.state.Prepare(rules, msg.ValidatedSender(), msg.ValidatedFeePayer(), st.evm.Context.Coinbase, msg.To(), vm.ActivePrecompiles(rules), msg.AccessList())
+
+	// Check authorizations list validity.
+	if msg.AuthorizationList() != nil {
+		for _, auth := range msg.AuthorizationList() {
+			// Verify chain ID is 0 or equal to current chain ID.
+			if auth.ChainID != big.NewInt(0) && st.evm.ChainConfig().ChainID != auth.ChainID {
+				continue
+			}
+			// Limit nonce to 2^64-1 per EIP-2681.
+			if auth.Nonce+1 < auth.Nonce {
+				continue
+			}
+			// Validate signature values and recover authority.
+			authority, err := auth.Authority()
+			if err != nil {
+				continue
+			}
+			// Check the authority account 1) doesn't have code or has exisiting
+			// delegation 2) matches the auth's nonce
+			st.state.AddAddressToAccessList(authority)
+			code := st.state.GetCode(authority)
+			if _, ok := types.ParseDelegation(code); len(code) != 0 && !ok {
+				continue
+			}
+			if have := st.state.GetNonce(authority); have != auth.Nonce {
+				continue
+			}
+			// If the account already exists in state, refund the new account cost
+			// charged in the initrinsic calculation.
+			if exists := st.state.Exist(authority); exists {
+				st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
+			}
+			st.state.SetNonce(authority, auth.Nonce+1)
+			delegation := types.AddressToDelegation(auth.Address)
+			if auth.Address == common.ZeroAddress {
+				// If the delegation is for the zero address, completely clear all
+				// delegations from the account.
+				delegation = []byte{}
+			}
+			st.state.SetCode(authority, delegation)
+
+			// Usually the transaction destination and delegation target are added to
+			// the access list in statedb.Prepare(..), however if the delegation is in
+			// the same transaction we need add here as Prepare already happened.
+			if *msg.To() == authority {
+				st.state.AddAddressToAccessList(auth.Address)
+			}
+		}
+	}
 
 	// Check whether the init code size has been exceeded.
 	if rules.IsShanghai && msg.To() == nil && len(st.data) > params.MaxInitCodeSize {
--- a/blockchain/state_transition.go
+++ b/blockchain/state_transition.go
@@ -405,7 +405,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 			}
 			st.state.IncNonce(authority)
 			delegation := types.AddressToDelegation(auth.Address)
-			if auth.Address == common.ZeroAddress {
+			if common.EmptyAddress(auth.Address) {
 				// If the delegation is for the zero address, completely clear all
 				// delegations from the account.
 				delegation = []byte{}
--- a/blockchain/state_transition.go
+++ b/blockchain/state_transition.go
@@ -397,6 +397,10 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 			// If the account already exists in state, refund the new account cost
 			// charged in the initrinsic calculation.
 			if exists := st.state.Exist(authority); exists {
+				// If the account is not AccountKeyTypeLegacy, setcode is not allowed.
+				if !st.state.GetKey(authority).Type().IsLegacyAccountKey() {
+					continue
+				}
 				st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
 			}
 			st.state.IncNonce(authority)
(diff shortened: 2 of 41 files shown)


Mdaiki0730:

Mdaiki0730:
recheck

blockchain/state_transition.go
```diff
@@ -347,11 +358,68 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+			}
+			if have := st.state.GetNonce(authority); have != auth.Nonce {
+				continue
+			}
+			// If the account already exists in state, refund the new account cost
+			// charged in the initrinsic calculation.
+			if exists := st.state.Exist(authority); exists {
+				st.state.AddRefund(params.CallNewAccountGas - params.TxAuthTupleGas)
+			}
+			st.state.IncNonce(authority)
```
ian0371, line 402: Please check if `authority` is not account-updated (e.g., authority is not `AccountKeyWeightedMultiSig`).
If we allow such accounts to be set-coded, then the account can be stolen when the original private key is leaked, even if it is no longer used.

ian0371: On a second thought, it is also the case for Ethereum if the original private key is leaked.
The difference is that the users may think that the original key is "deprecated" when updating the account.

However, I still suggest we block set-code tx on account-updated accounts.
For example, [our secretary account](https://kaiascope.com/tx/0xb17fa3a6de3af1048a8f2a7789e3458e9830f12f428f666dcf2459e5ad1c4930?tabId=accountKeyInfo) (which is the owner of the Voting contract) has been account-updated as a multisig account (3-of-5).

- address: 0x815643D2C645e4cCC39a7Bab3BdB735bEc451899
- owners:
  - 0x77ADe79c884306E0CFBB43F1F28e0ed09a117BdD
  - 0x815643D2C645e4cCC39a7Bab3BdB735bEc451899 (same as address)
  - 0x43C5f274f75aad4C3Bf40CF008e60c96F65dAE1e
  - ...

You can see that the second owner has the original private key of the account, so he can subvert the multisig scheme by himself with EIP-7702.

This is just off the top of my head, so it may need further discussion.

Mdaiki0730: That's a great opinion.
Since the signer of Tx is EthereumType Tx, only AccountKeyTypeLegacy is allowed, but the authority KeyType was not restricted.
I'll add validation to allow setcode only for AccountKeyTypeLegacy. Thank you!

da1suk8:
Where is the calculation of gas costs for the authorization list length being done?

**FYI**
- https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#gas-costs
- https://github.com/ethereum/go-ethereum/pull/30078/files#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92eR117-R119

blockchain/types/tx_internal_data_ethereum_set_code.go
```diff
@@ -0,0 +1,573 @@
...
+		reflect.DeepEqual(t.AccessList, ta.AccessList) &&
+		reflect.DeepEqual(t.AuthorizationList, ta.AuthorizationList) &&
+		t.V.Cmp(ta.V) == 0 &&
+		t.R.Cmp(ta.R) == 0 &&
+		t.S.Cmp(ta.S) == 0
+}
+
+func (t *TxInternalDataEthereumSetCode) IntrinsicGas(currentBlockNumber uint64) (uint64, error) {
+	gas, err := IntrinsicGas(t.Payload, t.AccessList, t.Recipient == nil, *fork.Rules(big.NewInt(int64(currentBlockNumber))))
+	return gas + uint64(len(t.AuthorizationList))*params.CallNewAccountGas, err
```
blukat29, line 310: @Mdaiki0730 Please move this logic into the IntrinsicGas function like `IntrinsicGas(payload, accesslist, authlist, ...)`. Ideally, `TxInternalDataEthereum*.IntrinsicGas` should be a proxy for the global `func IntrinsicGas()`. cc @tnasu

blockchain/state_transition.go
```diff
@@ -347,11 +358,72 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+			delegation := types.AddressToDelegation(auth.Address)
+			if auth.Address == common.ZeroAddress {
+				// If the delegation is for the zero address, completely clear all
+				// delegations from the account.
+				delegation = []byte{}
+			}
+			// TODO: Change to a method that adds code to EOA
```
tnasu, line 413: When TODO fix it?
I think we don't have to leave TODO comments.

tnasu (commented):
Questions:
* Is Kaia's `TxTypeEthereumDynamicFee` the same as Ethereum's `types.BlobTxType`?

blockchain/state_transition.go
blukat29: @tnasu Currently it could be inevitable to have TODOs because this feature comprises of multiple pull requests. I guess @shiki-tak will prepare another PR to fix it?

How about marking it TODO-7702 or something so we can later search for it?

common/types.go
```diff
@@ -49,6 +49,8 @@ var (
 	hashT    = reflect.TypeOf(Hash{})
 	extHashT = reflect.TypeOf(ExtHash{})
 	addressT = reflect.TypeOf(Address{})
+	// ZeroAddress represents the zero address value.
+	ZeroAddress = Address{}
```
blukat29, line 53: Use `common.EmptyAddress()` instead please.

blockchain/vm/operations_acl.go
```diff
@@ -243,3 +244,102 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
...
+		warmAccess := evm.StateDB.AddressInAccessList(addr)
+		// The WarmStorageReadCostEIP2929 (100) is already deducted in the form of a constant cost, so
+		// the cost to charge for cold access, if any, is Cold - Warm
+		coldCost := params.ColdAccountAccessCostEIP2929 - params.WarmStorageReadCostEIP2929
+		if !warmAccess {
+			evm.StateDB.AddAddressToAccessList(addr)
+			// Charge the remaining difference here already, to correctly calculate available
+			// gas for call
+			if !contract.UseGas(coldCost) {
+				return 0, ErrOutOfGas
```
blukat29, line 268:
```suggestion
				return 0, kerrors.ErrOutOfGas
```
Let's refactor it later along with other uses of `kerrors.ErrOutOfGas`.

blockchain/vm/operations_acl.go
```diff
@@ -243,3 +244,102 @@ func makeSelfdestructGasFn(refundsEnabled bool) gasFunc {
...
+		}
+		// In case of a cold access, we temporarily add the cold charge back, and also
+		// add it to the returned gas. By adding it to the return, it will be charged
+		// outside of this function, as part of the dynamic gas, and that will make it
+		// also become correctly reported to tracers.
+		contract.Gas += coldCost
+
+		var overflow bool
+		if gas, overflow = math.SafeAdd(gas, coldCost); overflow {
+			return 0, ErrGasUintOverflow
```
blukat29, line 303:
```suggestion
			return 0, errGasUintOverflow
```
Let's refactor it later along with other uses of `errGasUintOverflow`.

blockchain/types/tx_internal_data.go
```diff
@@ -58,7 +58,11 @@ const (
 	TxTypeKaiaLast, _, _
 	TxTypeEthereumAccessList = TxType(0x7801)
 	TxTypeEthereumDynamicFee = TxType(0x7802)
-	TxTypeEthereumLast       = TxType(0x7803)
+	// Kaia has determined that BLOB_TX defined in EIP4844 is not necessary.
+	// https://github.com/ethereum/EIPs/blob/70471d02d48a81ca963407abe9c48706059dc8e8/EIPS/eip-4844.md
```
blukat29, line 62:
```suggestion
	// EIP-4844 BLOB_TX_TYPE not supported in Kaia.
```

blockchain/vm/errors.go
```diff
@@ -36,6 +36,8 @@ var (
 	ErrTotalTimeLimitReached             = errors.New("reached the total execution time limit for txs in a block")
 	ErrOpcodeComputationCostLimitReached = errors.New("reached the opcode computation cost limit")
 	ErrFailedOnSetCode                   = errors.New("failed on setting code to an account")
+	ErrOutOfGas                          = errors.New("out of gas")
```
shiki-tak, line 39: How about using kerrors?
https://github.com/kaiachain/kaia/blob/dev/kerrors/kerrors.go#L29

blockchain/vm/errors.go
```diff
@@ -36,6 +36,8 @@ var (
...
+	ErrGasUintOverflow                   = errors.New("gas uint64 overflow")
```
shiki-tak, line 40: What about adding it to kerrors?

blockchain/state_transition.go
ulbqb: First of all, can legacy transactions be executed by account-updated accounts? I think this is one of legacy transactions. If so, should this feature be limited for only legacy accounts?

Mdaiki0730:
@tnasu 
> Is Kaia's TxTypeEthereumDynamicFee the same as Ethereum's types.BlobTxType?

No, it isn't. It is a Type 2 transaction of EIP1559.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

blockchain/state_transition.go
Mdaiki0730: > How about marking it TODO-7702 or something so we can later search for it?

As @blukat29 says, the current plan does not allow for TODOs to be avoided. I agree with this idea.

api/api_ethereum.go
```diff
@@ -767,8 +768,9 @@ type ethTxJSON struct {
 	To                   *common.Address `json:"to"`
 
 	// Access list transaction fields:
-	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
-	AccessList *types.AccessList `json:"accessList,omitempty"`
+	ChainID           *hexutil.Big             `json:"chainId,omitempty"`
+	AccessList        *types.AccessList        `json:"accessList,omitempty"`
+	AuthorizationList *types.AuthorizationList `json:"authorizationList,omitempty"`
```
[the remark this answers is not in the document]
Mdaiki0730: I don't think there's any problem with adding fields unless we have a specific reason to do so.
Do you mean create a new struct and inherit EthRPCTransaction?

blockchain/vm/errors.go
Mdaiki0730: As @blukat29 pointed out, that would be a good idea.
https://github.com/kaiachain/kaia/pull/119#discussion_r1830259252

blockchain/vm/errors.go
Mdaiki0730: There seems to be an existing function, so we'll use that.
https://github.com/kaiachain/kaia/pull/119#discussion_r1830259459

blockchain/types/tx_internal_data_ethereum_set_code.go
Mdaiki0730: There was a plan to remove the subclass. Ok, I'll move it.
Is the reason for removing because there are often duplicates?

blukat29: Exactly. Duplication is the reason. It's easy to miss something when we have 20+ files implementing subclasses.

blockchain/state_transition.go
Mdaiki0730: > The difference is that the users may think that the original key is "deprecated" when updating the account.

Just to confirm, if we update the key in Kaia, we will no longer be able to sign with the original private key, won't we?

api/api_ethereum.go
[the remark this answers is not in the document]
Mdaiki0730: I committed. Thank you.

blockchain/state_transition.go
ian0371: > Just to confirm, if we update the key in Kaia, we will no longer be able to sign with the original private key, won't we?

If we update the key in Kaia, we will no longer be able to *send transaction* with the original key.

> In this case, I think what is common to both Ethereum and Kaia is that users must be aware of the importance of keeping their private keys even after the code is set.

I think so too. I think we used to tout account-update feature can be used in case of private key theft.
Perhaps from now on, if a private key is stolen, we should recommend to users that they need to migrate to a new account rather than using account-update feature.

blukat29 (approved):

blockchain/state_transition.go
Mdaiki0730: This seems like a documentation issue, where do you keep track of tasks related to adding things to the Kaia docs?

blukat29: Perhaps we need a KIP document outlining the difference from Ethereum's EIP-7702?
- e.g. KIP-71 vs. EIP-1559 https://kips.kaia.io/KIPs/kip-71
- e.g. KIP-xxx vs. EIP-2028 https://github.com/kaiachain/kips/pull/23

Mdaiki0730: Also adding it to the Kaia doc may be necessary.
e.g) https://docs.kaia.io/learn/transactions/ethereum/#txtypeethereumdynamicfee-

Mdaiki0730: This seems to be a little off topic from the PR content, so shall we discuss it somewhere else?

ian0371: Sure. I just noticed that this has been resolved by e314ba1

blockchain/state_transition.go
tnasu: I'm just afraid to miss fixing it. And we don't have the power to forcibly correct it, so I think it's pointless. I believe that this part of the modification will be resolved through human communication rather than by looking at the TODOs in the source code.

blockchain/types/transaction_signing_test.go
```diff
@@ -37,6 +37,89 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestPragueSigningWithoutChainID(t *testing.T) {
```
hyunsooda, line 40: How about refactoring the common code in three functions?

blockchain/types/transaction_signing.go
```diff
@@ -59,7 +59,7 @@ func MakeSigner(config *params.ChainConfig, blockNumber *big.Int) Signer {
 	var signer Signer
 
 	if config.IsEthTxTypeForkEnabled(blockNumber) {
-		signer = NewLondonSigner(config.ChainID)
+		signer = NewPragueSigner(config.ChainID)
```
ian0371, line 62: I think the logic should be:
```
	if config.IsPragueForkEnabled(blockNumber) {
		signer = NewPragueSigner(config.ChainID)
	} else if config.IsEthTxTypeForkEnabled(blockNumber) {
		signer = NewLondonSigner(config.ChainID)
	} else {
		signer = NewEIP155Signer(config.ChainID)
	}
```

blockchain/types/transaction_signing.go
```diff
@@ -78,7 +78,7 @@ func LatestSigner(config *params.ChainConfig) Signer {
 	// Be aware that it checks whether EthTxTypeCompatibleBlock is set,
 	// but doesn't check whether it is enabled on a specific block number.
 	if config.EthTxTypeCompatibleBlock != nil {
-		return NewLondonSigner(config.ChainID)
+		return NewPragueSigner(config.ChainID)
```
ian0371, line 81: I think the logic should be:

```
	if config.IsPragueForkEnabled(blockNumber) {
		return NewPragueSigner(config.ChainID)
	} else if config.IsEthTxTypeForkEnabled(blockNumber) {
		return NewLondonSigner(config.ChainID)
	}
```

ian0371 (changes requested):

api/api_ethereum.go
```diff
@@ -770,6 +771,9 @@ type ethTxJSON struct {
 	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
 	AccessList *types.AccessList `json:"accessList,omitempty"`
 
+	// Authorization list transaction (SetCodeTx) fields:
```
tnasu, line 774:
```suggestion
	// SetCode transaction fields:
```

hyunsooda (commented):
Is there usage of this implementation? or other member is trying to apply this code setting to EOA with erc-4337 ecosystem?

tnasu:
@Mdaiki0730 

> > Is Kaia's TxTypeEthereumDynamicFee the same as Ethereum's types.BlobTxType?
> 
> No, it isn't. It is a Type 2 transaction of EIP1559.
> https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

I'm not sure whether it's enough that this PR modification depends on EIP1559-based. Because the modification of EIP7702 of `geth` depends on BlobTxType-based.

blockchain/types/transaction_signing.go
Mdaiki0730: > // Be aware that it checks whether EthTxTypeCompatibleBlock is set,
// but doesn't check whether it is enabled on a specific block number.

Following the comments above, this might be better.

```
        if config.PragueCompatibleBlock != nil {
		return NewPragueSigner(config.ChainID)
	} else if config.EthTxTypeCompatibleBlock != nil {
		return NewLondonSigner(config.ChainID)
	}
```

api/api_ethereum.go
Mdaiki0730: @tnasu cc @blukat29 
If we follow convention it would probably look something like this, but what do you think?
```
	// Set code transaction fields:
```

Mdaiki0730:
> Is there usage of this implementation? or other member is trying to apply this code setting to EOA with erc-4337 ecosystem?

Unfortunately there are no examples of this yet.
However, practical examples are being considered that take the EIP4337 format into account.
Core development is given priority due to schedule.

Mdaiki0730:
@tnasu 

> I'm not sure whether it's enough that this PR modification depends on EIP1559-based. Because the modification of EIP7702 of geth depends on BlobTxType-based.

Since kaia doesn't have `BlobTxType`, it seems best to base it on DynamicFee(Type2).
I don't think this would break the logic of EIP7702.

What do you think? @kaiachain/core

api/api_ethereum.go
blukat29: Yes that looks good!

api/api_ethereum.go
```diff
@@ -918,6 +936,14 @@ func formatTxToEthTxJSON(tx *types.Transaction) *ethTxJSON {
 		enc.ChainID = (*hexutil.Big)(tx.ChainId())
 		enc.MaxFeePerGas = (*hexutil.Big)(tx.GasFeeCap())
 		enc.MaxPriorityFeePerGas = (*hexutil.Big)(tx.GasTipCap())
+	case types.TxTypeEthereumSetCode:
+		al := tx.AccessList()
+		aul := tx.AuthorizationList()
```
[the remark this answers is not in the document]
Mdaiki0730: Yes, this is the function used when creating an EthTransaction from a Kaia transaction, so in the case of set code tx, you need to inherit aul as is.

api/api_ethereum.go
```diff
@@ -863,6 +867,20 @@ func newEthRPCTransaction(block *types.Block, tx *types.Transaction, blockHash c
 			// transaction is not processed yet
 			result.GasPrice = (*hexutil.Big)(tx.EffectiveGasPrice(nil, nil))
 		}
+	case types.TxTypeEthereumSetCode:
+		al := tx.AccessList()
+		aul := tx.AuthorizationList()
```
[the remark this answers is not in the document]
Mdaiki0730: Yes, this is the function used when creating an EthTransaction from a Kaia transaction, so in the case of set code tx, you need to inherit aul as is.

blockchain/types/tx_internal_data.go
```diff
@@ -58,7 +58,10 @@ const (
 	TxTypeKaiaLast, _, _
 	TxTypeEthereumAccessList = TxType(0x7801)
 	TxTypeEthereumDynamicFee = TxType(0x7802)
-	TxTypeEthereumLast       = TxType(0x7803)
+	// EIP-4844 BLOB_TX_TYPE not supported in Kaia.
+	_                     = TxType(0x7803)
+	TxTypeEthereumSetCode = TxType(0x7804)
+	TxTypeEthereumLast    = TxType(0x7805)
```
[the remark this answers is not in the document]
Mdaiki0730: As far as I can tell from investigating the usage, it is not related to state, so there should be no problem.
It is used as an exit flag when looping through all EthTxTypes and validation.
https://github.com/search?q=repo%3Akaiachain%2Fkaia%20TxTypeEthereumLast&type=code

blockchain/types/tx_internal_data_ethereum_set_code.go
```diff
@@ -0,0 +1,572 @@
...
+}
+
+// TxInternalDataEthereumSetCode represents a set code transaction.
+type TxInternalDataEthereumSetCode struct {
+	ChainID           *big.Int
+	AccountNonce      uint64
+	GasTipCap         *big.Int // a.k.a. maxPriorityFeePerGas
+	GasFeeCap         *big.Int // a.k.a. maxFeePerGas
+	GasLimit          uint64
+	Recipient         *common.Address `rlp:"nil"` // nil means contract creation
```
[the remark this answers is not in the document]
Mdaiki0730: This is the same as contract deployment.
EIP7702 mentions [not executing the init code](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#no-initcode), but this only means that the init code is not executed when the code is set for the authority. When the Recipient is nil, the contract will be deployed as usual.

e.g. eth impl) https://github.com/ethereum/go-ethereum/pull/30078/files#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92eR524-R526
e.g. erigon impl) https://github.com/erigontech/erigon/pull/10812/files#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92eL375-L381

blockchain/types/transaction_signing.go
Mdaiki0730: When adding IsPragueForkEnabled, we must specify PragueCompatibleBlock.
https://github.com/kaiachain/kaia/blob/f329a688e17b7741d35f2b4736c05d3d7a77d430/params/config.go#L54

Has the next hard fork block been decided? If not, would you like to temporarily enter the maximum value of big.Int?

ian0371: Oh actually we have the field in [dev branch](https://github.com/kaiachain/kaia/blob/dev/params/config.go#L62).
You can add the field and assign "nil", which indicates that the fork block is undetermined (similar to max value of big.Int).

Mdaiki0730: Thank you for letting me know. 
I would implement it so that conflict resolution is easy.

blockchain/types/tx_internal_data_ethereum_set_code.go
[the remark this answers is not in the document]
blukat29: > EIP7702 of geth depends on BlobTxType-based.

In [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md), a SetCodeTx comprises 13 fields. First 9 are said to "follow the same semantics as EIP-4844"
- same as 4844 (9): `chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, value, data, access_list`
- specific to 7702 (4): `authorization_list, signature_y_parity, signature_r, signature_s`

In [EIP-4844](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md), a BlobTx comprises 14 fields. 8 of them are said to "follow the same semantics as EIP-1559"
- same as 1559 (8): `chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, value, data, access_list`
- slightly different (1): `to`
  - "The field `to` deviates slightly from the semantics with the exception that it MUST NOT be `nil` and therefore must always represent a 20-byte address. This means that blob transactions cannot have the form of a create transaction."
- specific to 4844 (5): `max_fee_per_blob_gas, blob_versioned_hashes, y_parity, r, s`

Therefore, 8 fields of SetCodeTx are the same as DynamicFeeTx. According to EIP-7702 and EIP-4844, `to` field must not be `nil`. Refer to https://github.com/lightclient/go-ethereum/blob/eip-7702/core/types/tx_setcode.go#L59

Mdaiki0730:
The [PR105](https://github.com/kaiachain/kaia/pull/105) have been incorporated to incorporate Prague's config implementation. They are cherry-picked to avoid force-push. These will be conflict-resolved before being merged into dev.

blockchain/types/tx_internal_data_ethereum_set_code.go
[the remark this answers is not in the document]
Mdaiki0730: @tnasu @blukat29 
I had a wrong understanding of the recipient. Thank you for pointing that out.
It was fixed by https://github.com/kaiachain/kaia/pull/119/commits/f3352e4da73eb25c26d917a8972519a77e0e9535

blockchain/types/tx_internal_data_ethereum_set_code.go
```diff
@@ -0,0 +1,567 @@
...
+}
+
+// TxInternalDataEthereumSetCode represents a set code transaction.
+type TxInternalDataEthereumSetCode struct {
+	ChainID           *big.Int
+	AccountNonce      uint64
+	GasTipCap         *big.Int // a.k.a. maxPriorityFeePerGas
+	GasFeeCap         *big.Int // a.k.a. maxFeePerGas
+	GasLimit          uint64
+	Recipient         *common.Address
```
blukat29, line 64: Can you just make it `common.Address` type? That's what they did with go-ethereum https://github.com/lightclient/go-ethereum/blob/eip-7702/core/types/tx_setcode.go#L59

Mdaiki0730: Yes, I've corrected it.

blukat29 (approved):

api/api_ethereum.go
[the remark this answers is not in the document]
Mdaiki0730: Ah, I misunderstood the question.
You meant "Why don’t we need to initial/set for aul in this?".

I believe that authorization is a set code specific field and does not need to be initialized in other cases.

ian0371 (approved):

api/api_ethereum.go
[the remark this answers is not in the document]
Mdaiki0730: Ah, I misunderstood the question.
You meant "Why don’t we need to initial/set for aul in this?".

I believe that authorization is a set code specific field and does not need to be initialized in other cases.

blockchain/types/transaction_signing_test.go
Mdaiki0730: Thank you for your comment!
It was refoctored by https://github.com/kaiachain/kaia/pull/119/commits/aec484d2a77cc5bc9d28c10f82182cb88b6c84df.

api/api_ethereum.go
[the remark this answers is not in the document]
tnasu: I don't ask about `TxTypeEthereumSetCode`. I ask about the case of `TxTypeEthereumAccessList/TxTypeEthereumDynamicFee`. 

But I got it. Thanks

api/api_ethereum.go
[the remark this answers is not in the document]
tnasu: I don't ask about `TxTypeEthereumSetCode`. I ask about the case of `TxTypeEthereumAccessList/TxTypeEthereumDynamicFee`. 

But I got it. Thanks

blockchain/types/tx_internal_data_serializer_test.go
```diff
@@ -287,6 +289,26 @@ func genDynamicFeeTransaction() TxInternalData {
...
+func genSetCodeTransaction() TxInternalData {
+	tx, err := NewTxInternalDataWithMap(TxTypeEthereumSetCode, map[TxValueKeyType]interface{}{
+		TxValueKeyNonce:             nonce,
+		TxValueKeyTo:                to,
+		TxValueKeyAmount:            amount,
+		TxValueKeyGasLimit:          gasLimit,
+		TxValueKeyGasFeeCap:         gasFeeCap,
+		TxValueKeyGasTipCap:         gasTipCap,
+		TxValueKeyData:              []byte("1234"),
+		TxValueKeyAccessList:        accesses,
```
tnasu, line 301: Do we need the `accessList` for the test?
I've also check the Ethereum test codes.

https://github.com/ethereum/go-ethereum/pull/30078/files#diff-d78b498ba524953c810b53b2f1e866245d052a063863ed820c9b4852f8762d69R4306-R4314

```
		txdata := &types.SetCodeTx{
			ChainID:   gspec.Config.ChainID.Uint64(),
			Nonce:     0,
			To:        addr1,
			Gas:       500000,
			GasFeeCap: uint256.MustFromBig(newGwei(5)),
			GasTipCap: uint256.NewInt(2),
			AuthList:  []*types.Authorization{auth1, auth2},
		}
```

tnasu (commented):
I think `AccessList` can set empty list

https://github.com/ethereum/go-ethereum/pull/30078/files#diff-77719ae57e7e6c3e0cac05fa12c6c5f2e5f9bc810c034d1446de2414c36d9210R464-R466

```
		if dec.AccessList != nil {
			itx.AccessList = *dec.AccessList
		}
```

Mdaiki0730:
@tnasu Thank you for your comment.
> I think AccessList can set an empty list and the most simple test case is to set an empty list

Since this is not a test file, it also contains an access_list.
Because EIP7702 tx inherits the semantics of EIP1559, including gas cost calculation and fields, setting an empty list with  UnmarshalJSON is likely to cause problems.
- https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#set-code-transaction
- https://github.com/ethereum/EIPs/blob/2bb4314865ecb8f41b7fd950612cef584a5d1c2a/EIPS/eip-7702.md#gas-costs

blockchain/types/transaction_signing.go
```diff
@@ -251,6 +255,116 @@ type Signer interface {
...
+		// id, add 27 to become equivalent to unprotected Homestead signatures.
+		V := new(big.Int).Add(v, big.NewInt(27))
+		return V
+	})
+}
+
+// SenderFeePayer returns the public key derived from tx signature and txhash.
+func (s pragueSigner) SenderFeePayer(tx *Transaction) ([]*ecdsa.PublicKey, error) {
+	// EIP-7702(Set code transaction) tx don't supported fee-delegation.
+	// EIP-1559(Dynamic fee transaction) tx don't supported fee-delegation.
```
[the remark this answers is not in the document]
Mdaiki0730: Thank you! It was removed by https://github.com/kaiachain/kaia/pull/119/commits/009398e7da934af4dbcc8b8ab5168cb67ea3d51c

blockchain/types/transaction_test.go
```diff
@@ -834,6 +834,18 @@ func TestTransactionCoding(t *testing.T) {
 				GasTipCap:    big.NewInt(10),
 				AccessList:   accesses,
 			}
+		case 8:
+			// Tx with non-zero authorization list.
+			txData = &TxInternalDataEthereumSetCode{
+				ChainID:           big.NewInt(1),
+				AccountNonce:      i,
+				Recipient:         recipient,
```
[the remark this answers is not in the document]
Mdaiki0730: Similar to the ethereum implementation, we decided to treat the recipient field in SET_CODE_TX as common.Address.
ref: https://github.com/kaiachain/kaia/pull/119#discussion_r1836169314

blockchain/types/transaction_signing_test.go
```diff
@@ -37,6 +37,69 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestPragueSigning(t *testing.T) {
+	key, _ := crypto.GenerateKey()
+	addr := crypto.PubkeyToAddress(key.PublicKey)
+	accessList := AccessList{{Address: common.HexToAddress("0x0000000000000000000000000000000000000001"), StorageKeys: []common.Hash{{0}}}}
```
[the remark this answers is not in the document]
Mdaiki0730: I don't think it matters either.
These can be considered an extension of the existing test, so I think we can include access_list as well.
https://github.com/kaiachain/kaia/blob/178185beb4c576c6efbed3baf2171ae2db8ac599/blockchain/types/transaction_signing_test.go#L40-L115

blockchain/types/tx_internal_data_serializer_test.go
Mdaiki0730: I don't think it matters either.
These can be considered an extension of the existing test, so I think we can include access_list as well.
https://github.com/kaiachain/kaia/blob/178185beb4c576c6efbed3baf2171ae2db8ac599/blockchain/types/tx_internal_data_serializer_test.go#L271-L288

blockchain/types/transaction_test.go
```diff
@@ -834,6 +834,18 @@ func TestTransactionCoding(t *testing.T) {
...
+				GasLimit:          123457,
+				GasFeeCap:         big.NewInt(10),
+				GasTipCap:         big.NewInt(10),
+				AccessList:        accesses,
```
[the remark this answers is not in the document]
Mdaiki0730: I have a different opinion on this one and would like to suggest adding a case to test for an empty accessList.
The existing test cases also test an empty accessList for each tx type.
https://github.com/kaiachain/kaia/blob/178185beb4c576c6efbed3baf2171ae2db8ac599/blockchain/types/transaction_test.go#L814-L846

It was added by https://github.com/kaiachain/kaia/pull/119/commits/2cabab85181491c3dd9bd1e0539ae87d68973b84

PR #121, opened 2024-10-31
blockchain: implement EIP-2935

hyunsooda:
## Proposed changes

The main function has been implemented, but a task has been queued to simplify the history store callsite.

- Draft commit: `c1714134` (implement the main function)
- Final commit: `b38c8f2` (simplify the historical store callsite)

### Contract Deployment Process

According to EIP-2935 deployment transaction artifact, the code below generate the exact same system history storage address.
```js
const ethers = require('ethers');
const tx = {
  "from": "0xe473f7e92ba2490e9fcbbe8bb9c3be3adbb74efc",
  "type": "0x00",
  "nonce": "0x",
  "to": "0x",
  "gas": "0x03d090",
  "gasPrice": "0xe8d4a51000",
  "maxPriorityFeePerGas": null,
  "maxFeePerGas": null,
  "value": "0x",
  "input": "0x60648060095f395ff33373fffffffffffffffffffffffffffffffffffffffe1460575767ffffffffffffffff5f3511605357600143035f3511604b575f35612000014311604b57611fff5f3516545f5260205ff35b5f5f5260205ff35b5f5ffd5b5f35611fff60014303165500",
  "v": "0x1b",
  "r": "0x0539",
  "s": "0x1b9b6eb1f0",
  "hash": "0x3c769a03d6e2212f1d26ab59ba797dce0900df29ffd23c1dd391fd6b217973ad",
}

const signedEncodedTransaction = ethers.encodeRlp([tx.nonce, tx.gasPrice, tx.gas, tx.to, tx.value, tx.input, tx.v, tx.r, tx.s])
console.log(signedEncodedTransaction)

// generated rlp raw transcation bytes: 0xf8868085e8d4a510008303d0908080b86d60648060095f395ff33373fffffffffffffffffffffffffffffffffffffffe1460575767ffffffffffffffff5f3511605357600143035f3511604b575f35612000014311604b57611fff5f3516545f5260205ff35b5f5f5260205ff35b5f5ffd5b5f35611fff600143031655001b820539851b9b6eb1f0
```

```
> eth.sendRawTransaction("0xf8868085e8d4a510008303d0908080b86d60648060095f395ff33373fffffffffffffffffffffffffffffffffffffffe1460575767ffffffffffffffff5f3511605357600143035f3511604b575f35612000
014311604b57611fff5f3516545f5260205ff35b5f5f5260205ff35b5f5ffd5b5f35611fff600143031655001b820539851b9b6eb1f0")
"0x3c769a03d6e2212f1d26ab59ba797dce0900df29ffd23c1dd391fd6b217973ad"
> kaia.getTransactionReceipt("0x3c769a03d6e2212f1d26ab59ba797dce0900df29ffd23c1dd391fd6b217973ad")
{
  blockHash: "0x9af795d26881f60aafce80051434a4f06a8209bbeb5512bc037f2cb0fcfd574e",
  blockNumber: 19560,
  contractAddress: "0x0aae40965e6800cd9b1f4b05ff21581047e3f91e", // history storage address
  ...
  ...
}
```

### Usage
```solidity
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity 0.8.24;

 contract EIP2935 {
     address constant HISTORY_STORAGE_ADDR = 0x0AAE40965E6800cD9b1f4b05ff21581047E3F91e;

     function getHash(uint256 blockNumber) public returns (uint256) {
         (bool ok, bytes memory output) = HISTORY_STORAGE_ADDR.call(abi.encode(blockNumber));
         require(ok, "historical retreival error");
         return uint256(bytes32(output));
     }
 }
```

Once deployed, retrieval works like below:

```
> kaia.call({to:"0x0eF71b533f074383A14696f7EeBD94446b10F679", data:"0x6b2fafa90000000000000000000000000000000000000000000000000000000000004f6d"}) // getHash(20333)
> 0x52105353978f50ba403e770e7a920d70c3ee9a2df991a428fd5fc6aebaacf4f5 // Hash of block 20333
```

### Implementation details

- Add `consensus.Engine.Initialize()` that performs pre-block state transitions.
- TraceChain: traces blocks `(start, end]`
```diff
  for num = start; num < end; num++ {
-   statedb = StateAtBlock(num)
+   statedb = StateAtTransaction(num+1, 0)
    for tx in block(num+1)
      traceTx(tx)
  }
```
- TraceBlock: traces block `num`
```diff
- statedb = StateAtBlock(num-1)
+ statedb = StateAtTransaction(num, 0)
  for tx in block(num)
    traceTx(tx)
```
- TraceTransaction: traces the `tx` in block `num` -- unchanged
```diff
  statedb = StateAtTransaction(num, idx)
  traceTx(tx)
```
- TraceCall: traces the `tx` at the end of the block `num`, not the beginning of `num+1` -- unchanged
```
  statedb = StateAtBlock(num)
  traceTx(tx)
```

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/blockchain/evm.go
+++ b/blockchain/evm.go
@@ -45,8 +45,6 @@ type ChainContext interface {
 
 	// GetHeader returns the hash corresponding to their hash.
 	GetHeader(common.Hash, uint64) *types.Header
-
-	Config() *params.ChainConfig
 }
 
 // NewEVMBlockContext creates a new context for use in the EVM.
--- a/blockchain/evm.go
+++ b/blockchain/evm.go
@@ -45,6 +45,8 @@ type ChainContext interface {
 
 	// GetHeader returns the hash corresponding to their hash.
 	GetHeader(common.Hash, uint64) *types.Header
+
+	Config() *params.ChainConfig
 }
 
 // NewEVMBlockContext creates a new context for use in the EVM.
--- a/blockchain/state_processor.go
+++ b/blockchain/state_processor.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kaiachain/kaia/blockchain/state"
 	"github.com/kaiachain/kaia/blockchain/types"
 	"github.com/kaiachain/kaia/blockchain/vm"
+	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/consensus"
 	"github.com/kaiachain/kaia/params"
 )
@@ -74,6 +75,12 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		processStats     ProcessStats
 	)
 
+	if p.config.IsPragueForkEnabled(block.Number()) {
+		context := NewEVMBlockContext(header, p.bc, nil)
+		vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, &cfg)
+		ProcessParentBlockHash(header, vmenv, statedb, p.config.Rules(header.Number))
+	}
+
 	// Extract author from the header
 	author, _ := p.bc.Engine().Author(header) // Ignore error, we're past header validation
 
@@ -98,4 +105,38 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	processStats.AfterFinalize = time.Now()
 
 	return receipts, allLogs, *usedGas, internalTxTraces, processStats, nil
+}
+
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
+// as per EIP-2935.
+func ProcessParentBlockHash(header *types.Header, vmenv *vm.EVM, statedb vm.StateDB, rules params.Rules) error {
+	var (
+		from     = params.SystemAddress
+		data     = header.ParentHash.Bytes()
+		gasLimit = uint64(30_000_000)
+	)
+
+	intrinsicGas, err := types.IntrinsicGas(data, nil, false, rules)
+	if err != nil {
+		return err
+	}
+
+	msg := types.NewMessage(
+		from,
+		&params.HistoryStorageAddress,
+		0,
+		common.Big0,
+		gasLimit,
+		common.Big0,
+		data,
+		false,
+		intrinsicGas,
+		nil,
+	)
+
+	vmenv.Reset(NewEVMTxContext(msg, header, vmenv.ChainConfig()), statedb)
+	statedb.AddAddressToAccessList(params.HistoryStorageAddress)
+	vmenv.Call(vm.AccountRef(from), *msg.To(), msg.Data(), gasLimit, common.Big0)
+	statedb.Finalise(true, true)
+	return nil
 }
--- a/blockchain/state_processor.go
+++ b/blockchain/state_processor.go
@@ -75,11 +75,7 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		processStats     ProcessStats
 	)
 
-	if p.config.IsPragueForkEnabled(block.Number()) {
-		context := NewEVMBlockContext(header, p.bc, nil)
-		vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, p.config, &cfg)
-		ProcessParentBlockHash(header, vmenv, statedb, p.config.Rules(header.Number))
-	}
+	p.engine.Initialize(p.bc, header, statedb)
 
 	// Extract author from the header
 	author, _ := p.bc.Engine().Author(header) // Ignore error, we're past header validation
--- a/blockchain/vm/interface.go
+++ b/blockchain/vm/interface.go
@@ -104,4 +104,6 @@ type StateDB interface {
 	GetTxHash() common.Hash
 
 	GetKey(address common.Address) accountkey.AccountKey
+
+	Finalise(bool, bool)
 }
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -102,7 +102,7 @@ type Engine interface {
 	Prepare(chain ChainReader, header *types.Header) error
 
 	// Initialize runs any pre-transaction state modifications (e.g., EIP-2539)
-	Initialize(chain ChainContext, header *types.Header, state *state.StateDB)
+	Initialize(chain ChainReader, header *types.Header, state *state.StateDB)
 
 	// Finalize runs any post-transaction state modifications (e.g. block rewards)
 	// and assembles the final block.
@@ -187,10 +187,4 @@ type ConsensusInfo struct {
 	Committee      []common.Address
 	Committers     []common.Address
 	Round          byte
-}
-
-type ChainContext interface {
-	Config() *params.ChainConfig
-	Engine() Engine
-	GetHeader(hash common.Hash, number uint64) *types.Header
 }
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -101,6 +101,9 @@ type Engine interface {
 	// rules of a particular engine. The changes are executed inline.
 	Prepare(chain ChainReader, header *types.Header) error
 
+	// Initialize runs any pre-transaction state modifications (e.g., EIP-2539)
+	Initialize(chain ChainContext, header *types.Header, state *state.StateDB)
+
 	// Finalize runs any post-transaction state modifications (e.g. block rewards)
 	// and assembles the final block.
 	// Note: The block header and state database might be updated to reflect any
@@ -184,4 +187,10 @@ type ConsensusInfo struct {
 	Committee      []common.Address
 	Committers     []common.Address
 	Round          byte
+}
+
+type ChainContext interface {
+	Config() *params.ChainConfig
+	Engine() Engine
+	GetHeader(hash common.Hash, number uint64) *types.Header
 }
--- a/node/cn/state_accessor.go
+++ b/node/cn/state_accessor.go
@@ -214,6 +214,13 @@ func (cn *CN) stateAtTransaction(block *types.Block, txIndex int, reexec uint64)
 	if err != nil {
 		return nil, vm.BlockContext{}, vm.TxContext{}, nil, nil, err
 	}
+	// If prague hardfork, insert parent block hash in the state as per EIP-2935.
+	if cn.chainConfig.IsPragueForkEnabled(block.Number()) {
+		header := block.Header()
+		context := blockchain.NewEVMBlockContext(header, cn.blockchain, nil)
+		vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, cn.blockchain.Config(), &vm.Config{})
+		blockchain.ProcessParentBlockHash(header, vmenv, statedb, cn.chainConfig.Rules(header.Number))
+	}
 	if txIndex == 0 && len(block.Transactions()) == 0 {
 		return nil, vm.BlockContext{}, vm.TxContext{}, statedb, release, nil
 	}
--- a/node/cn/state_accessor.go
+++ b/node/cn/state_accessor.go
@@ -215,12 +215,8 @@ func (cn *CN) stateAtTransaction(block *types.Block, txIndex int, reexec uint64)
 		return nil, vm.BlockContext{}, vm.TxContext{}, nil, nil, err
 	}
 	// If prague hardfork, insert parent block hash in the state as per EIP-2935.
-	if cn.chainConfig.IsPragueForkEnabled(block.Number()) {
-		header := block.Header()
-		context := blockchain.NewEVMBlockContext(header, cn.blockchain, nil)
-		vmenv := vm.NewEVM(context, vm.TxContext{}, statedb, cn.blockchain.Config(), &vm.Config{})
-		blockchain.ProcessParentBlockHash(header, vmenv, statedb, cn.chainConfig.Rules(header.Number))
-	}
+	cn.engine.Initialize(cn.blockchain, block.Header(), statedb)
+
 	if txIndex == 0 && len(block.Transactions()) == 0 {
 		return nil, vm.BlockContext{}, vm.TxContext{}, statedb, release, nil
 	}
--- a/node/cn/state_accessor.go
+++ b/node/cn/state_accessor.go
@@ -198,7 +198,7 @@ func (cn *CN) stateAtBlock(block *types.Block, reexec uint64, base *state.StateD
 }
 
 // stateAtTransaction returns the execution environment of a certain transaction.
-func (cn *CN) stateAtTransaction(block *types.Block, txIndex int, reexec uint64) (blockchain.Message, vm.BlockContext, vm.TxContext, *state.StateDB, tracers.StateReleaseFunc, error) {
+func (cn *CN) stateAtTransaction(block *types.Block, txIndex int, reexec uint64, base *state.StateDB, readOnly bool, preferDisk bool) (blockchain.Message, vm.BlockContext, vm.TxContext, *state.StateDB, tracers.StateReleaseFunc, error) {
 	// Short circuit if it's genesis block.
 	if block.NumberU64() == 0 {
 		return nil, vm.BlockContext{}, vm.TxContext{}, nil, nil, errors.New("no transaction in genesis")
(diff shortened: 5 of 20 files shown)


blockchain/vm/interface.go
```diff
@@ -104,4 +104,6 @@ type StateDB interface {
 	GetTxHash() common.Hash
 
 	GetKey(address common.Address) accountkey.AccountKey
+
+	Finalise(bool, bool)
```
blukat29, line 108: What is it for?

blockchain/state_processor.go
```diff
@@ -99,3 +102,37 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 
 	return receipts, allLogs, *usedGas, internalTxTraces, processStats, nil
 }
+
+// ProcessParentBlockHash stores the parent block hash in the history storage contract
+// as per EIP-2935.
+func ProcessParentBlockHash(header *types.Header, vmenv *vm.EVM, statedb vm.StateDB, rules params.Rules) error {
```
blukat29, line 108: Can we move this function into `consensus/misc/eip2935.go`? as we adopt Initialize(), the 2935 is more like a part of Istanbul consensus engine. Thus having it under consensus directory.

blockchain/vm/interface.go
blukat29, line 108: If we move ProcessParentBlockHash into consensus directory, check if we still need this addition. engine.Initialize receives the concrete type *StateDB, not this interface vm.StateDB.

consensus/consensus.go
```diff
@@ -185,3 +188,9 @@ type ConsensusInfo struct {
 	Committers     []common.Address
 	Round          byte
 }
+
+type ChainContext interface {
```
blukat29, line 192: You don't need separate ChainContext type. ChainReader already has these.

hyunsooda: There are three callsites in api package. `api.backend` does not satisfy the interface of `ChianReader`, so I must create a minimal interface type.

ian0371, line 192: Can we reuse `ChainReader` instead of introducing a new type?

ian0371: oh, it's been mentiond.

ian0371: How about adding missing methods here instead?

https://github.com/kaiachain/kaia/blob/8fb6105b4c0455caefc67caf7b314b40b7333912/blockchain/evm.go#L42-L50

node/cn/tracers/api.go
```diff
@@ -451,6 +455,9 @@ func (api *CommonAPI) traceChain(start, end *types.Block, config *TraceConfig, n
 				failed = err
 				break
 			}
+			// Insert parent hash in history contract.
+			api.backend.Engine().Initialize(newChainContext(localctx, api.backend), next.Header(), statedb)
```
ian0371, line 459: Is this necesesary here? I think this is called by the call chain:
```
api.go:453 StateAtBlock()
api_backend.go:404 stateAtBlock()
state_accessor.go:167 Process()
```

blockchain/state_processor.go
hyunsooda: The statement `vmenv.Reset(blockchain.NewEVMTxContext(msg, header, vmenv.ChainConfig()), statedb)` matters. `NewEVMTxContext()` is defined in the `blockchain` package, and the `misc` cannot be imported by the cycle.

Of course I can pass it `any` type, but it is fragile if the change is made in the signature of  `NewEVMTxContext` without considering the `ProcessParentBlockHash` function because its dependency is not duty of compiler. Also, adding function signature comparison at runtime, but still not effective method I think.

blockchain/vm/interface.go
hyunsooda: Waiting for the [prerequisit](https://github.com/kaiachain/kaia/pull/121#discussion_r1825432048)

node/cn/tracers/api.go
hyunsooda: I think `Process()` might not be touched. Thus, marking the blockhash at this point as well even if the `readOnly` is false.

node/cn/tracers/api.go
```diff
@@ -667,7 +679,6 @@ func (api *CommonAPI) traceBlock(ctx context.Context, block *types.Block, config
 				}
 
 				txCtx := blockchain.NewEVMTxContext(msg, block.Header(), api.backend.ChainConfig())
-				blockCtx := blockchain.NewEVMBlockContext(block.Header(), newChainContext(ctx, api.backend), nil)
```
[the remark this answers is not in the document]
hyunsooda: The loop handles transactions in given a block and no parameter changes made in the loop. You can refer [geth](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go#L607)

consensus/consensus.go
hyunsooda: Uhm,, really? The `Backend` seems too bloated to match the defined function names. I believe that's why the `ChainContext` type exists. I'm okay with extending one, but the impact of changes seems to be considered before making a change.

node/cn/tracers/api.go
ian0371: `StateAtBlock` seems to only return statedb only when the "correct" one is generated.
If `Initialize` wasn't called by `Process`, then `Finalize` would not be called as well, in which case `StateAtBlock` will return an error.

btw, is `Initialize` idempotent? (I think it is now, but perhaps can change later)

hyunsooda: > If Initialize wasn't called by Process, then Finalize would not be called as well

The `Finalize` of state database(not engine) is done in `ProcessParentBlockHash`. Does it answer indirectly?

> btw, is Initialize idempotent? (I think it is now, but perhaps can change later)

If you mean the idempotent in a unit of single block, then yes. I'm sure you were asking in this context.
Of course, calling two times with the same parameter at different block numbers(e.g., `n` and `n+1`) differs, which stores the same parameter(hash) into `block.number - 1` and `block.number`, respectively.

ian0371: I'm not talking about `statedb.Finalise` (with an "s").
`Initialize` always goes in pair with `engine.Finalize`, right? (please correct me if I'm wrong)
`Process` either (1) calls both `Initialize` and `Finalize`, or (2) calls neither.
`StateAtBlock` does not return an error only in the first case.

I'm concerning if later in the future when `Initialize` stops being idempotent, this line's additional call can be problematic.
Maybe I'm over concerning, so please tell me your opinion.

Also, I just found out if we're going to leave this code, I think it should go to the bottom (after statedb is sent to the channel):
https://github.com/kaiachain/kaia/blob/8fb6105b4c0455caefc67caf7b314b40b7333912/node/cn/tracers/api.go#L469-L478

blockchain/state_processor.go
blukat29: Oh I see. I'm fine with it then. Thank you for thoughts.

consensus/consensus.go
blukat29: I confirmed that tracers/api.go:backend does not satisfy ChainReader and there is a wrapper class chainContext specifically for calling NewEVMBlockContext. I think having consensus.ChainContext is our best bet in the current situation. After this PR, let me propose another PR that simplifies tracer.

node/cn/tracers/api.go
```diff
@@ -647,7 +654,12 @@ func (api *CommonAPI) traceBlock(ctx context.Context, block *types.Block, config
 
 		pend = new(sync.WaitGroup)
 		jobs = make(chan *txTraceTask, len(txs))
+
+		header   = block.Header()
+		blockCtx = blockchain.NewEVMBlockContext(header, newChainContext(ctx, api.backend), nil)
 	)
+	api.backend.Engine().Initialize(newChainContext(ctx, api.backend), header, statedb)
```
blukat29, line 661: In L643, you can replace StateAtBlock with StateAtTransaction(... txIndex=0 ...) to eliminate the need to Initialize(). The same goes for traceChain.

Then the whole discussion about ChainContext can be voided.

node/cn/tracers/api.go
ian0371: I wanted all Initialize from `api.go` be removed, but I didn't think we could.
After reading ollie's comment, I took another look, and I think we can.

node/cn/state_accessor.go
```diff
@@ -214,6 +214,9 @@ func (cn *CN) stateAtTransaction(block *types.Block, txIndex int, reexec uint64)
 	if err != nil {
 		return nil, vm.BlockContext{}, vm.TxContext{}, nil, nil, err
 	}
+	// If prague hardfork, insert parent block hash in the state as per EIP-2935.
+	cn.engine.Initialize(cn.blockchain, block.Header(), statedb)
```
ian0371, line 218: I think this should be removed as well, just like `Initialize` is not called from `cn.stateAtBlock`.

node/cn/tracers/api.go
blukat29: As-is
```
traceBlock(100)
  stateAtBlock(99) -> returns the state after block 99
  for tx in txs(100)
    ApplyTransaction(tx)
```

To-be
```
traceBlock(100)
  stateAtTransaction(100, txIndex=0)
    stateAtBlock(99)
    engine.Initialize()
  for tx in txs(100)
    ApplyTransaction(tx)
```

Note that with EIP-2953, `state after block 99 = state before block 100` is no longer true. Tracer APIs want the `state before block 100` so using `stateAtTransaction` seems more appropriate.

node/cn/state_accessor.go
ian0371: Sorry, I was mistaken. Never mind

node/cn/tracers/api.go
ian0371: Sorry, I didn't know it was for tracing N+1 block where StateAtBlock(N) + Initialize is required.

node/cn/tracers/api.go
hyunsooda: To verify my understanding, an outline below was your suggestion?

```
traceBlock(100)
  stateAtTransaction(100, txIndex=0)
  for tx in txs(100)
    ApplyTransaction(tx)
```

which contributes to removing the explicit `Initialize()`  call. Additionally, system transactions do not make dirty the state before executing a simulation. Correct?

hyunsooda: PTAL https://github.com/kaiachain/kaia/pull/121/commits/20c2c1c181573ea210211fa347040019c575e347

node/cn/tracers/api.go
```diff
@@ -429,34 +429,12 @@ func (api *CommonAPI) traceChain(start, end *types.Block, config *TraceConfig, n
...
 			}
-			// Prepare the statedb for tracing. Don't use the live database for
-			// tracing to avoid persisting state junks into the database. Switch
-			// over to `preferDisk` mode only if the memory usage exceeds the
-			// limit, the trie database will be reconstructed from scratch only
-			// if the relevant state is available in disk.
-			var preferDisk bool
-			if statedb != nil {
-				s1, s2, s3 := statedb.Database().TrieDB().Size()
-				preferDisk = s1+s2+s3 > defaultTracechainMemLimit
```
blukat29, line 447: @hyunsooda Can you revive this logic? arguments `StateAtBlock(..., base *state.StateDB, readOnly bool, preferDisk bool(` might have a huge impact on performance. I'm okay with adding the arguments to StateAtTransaction.

node/cn/tracers/api.go
```diff
@@ -743,7 +715,7 @@ func (api *CommonAPI) standardTraceBlockToFile(ctx context.Context, block *types
 	if config != nil && config.Reexec != nil {
 		reexec = *config.Reexec
 	}
-	statedb, release, err := api.backend.StateAtBlock(ctx, parent, reexec, nil, true, false)
+	_, _, _, statedb, release, err := api.backend.StateAtTransaction(ctx, parent, 0, reexec)
```
blukat29, line 718: Maybe we should do the same for the TraceCall.
1. For correctness: TraceCall should simulate a transaction execution after Initialize().
2. For future extensibility: https://github.com/ethereum/go-ethereum/blob/da17f2d65b8922a20f86740644b2d7f6ef688e9b/eth/tracers/api.go#L948-L952 geth eth_traceCall RPC takes optional TxIndex argument. If we're going to adopt the feature, it has to be implemented with StateAtTransaction.

node/cn/tracers/api.go
hyunsooda: Sorry, I don't understand what you are concerned about. I replaced `StateAtBlock` to `StateAtTransaction`. With the argument tx index zero, the loop cost would be zero. Could you explain more about this?

node/cn/tracers/api.go
hyunsooda: 1. I'll replace `StateAtBlock` to `StateAtTransaction` with tx index zero.
2. This seems to be an independent feature in this PR. Do you want me to add that feature together within this PR?

hyunsooda:
@hyeonLewis Thank you. Typo fixed.

node/cn/tracers/api.go
blukat29: Sorry I didn't mean to add the feature now. It might be nice to replace all call sites of StateAtBlock with StateAtTransaction so later we can be more comfortable with introducing the feature.

node/cn/tracers/api.go
blukat29: In StateAtTransaction(block, txIndex, reexec), it calls `StateAtBlock(parent, reexec, nil, true, false)`. The last 3 arguments are fixed. But it seems TraceChain wants to customize those 3 arguments to be `base, false, preferDisk`. By replacing StateAtBlock with StateAtTransaction, the 3 arguments are now fixed to be `nil, true, false`. I think the performance might degrade by the change.

So I suggest you to allow StateAtTransaction to receive the 3 arguments and pass it to StateAtBlock.

node/cn/tracers/api.go
hyunsooda: PTAL https://github.com/kaiachain/kaia/pull/121/commits/a9f4650f134984439abfbbd8650a9c44d8616e70

node/cn/tracers/api.go
hyunsooda: PTAL https://github.com/kaiachain/kaia/pull/121/commits/b886f3f020ac4ca2adbeb68308b6b40b54e0ac8a

blockchain/evm.go
```diff
@@ -45,6 +45,8 @@ type ChainContext interface {
 
 	// GetHeader returns the hash corresponding to their hash.
 	GetHeader(common.Hash, uint64) *types.Header
+
+	Config() *params.ChainConfig
```
blukat29, line 49: Is it unnecessary because we now use ChainReader?

node/cn/tracers/api.go
```diff
@@ -159,6 +159,10 @@ func (context *chainContext) GetHeader(hash common.Hash, number uint64) *types.H
 	return header
 }
 
+func (context *chainContext) Config() *params.ChainConfig {
```
blukat29, line 162: If blockchain.ChainContext.Config() becomes unnecessary, please remove it as well.

node/cn/tracers/api.go
```diff
@@ -891,7 +891,18 @@ func (api *CommonAPI) TraceCall(ctx context.Context, args kaiaapi.CallArgs, bloc
...
 		reexec = *config.Reexec
 	}
-	statedb, release, err := api.backend.StateAtBlock(ctx, block, reexec, nil, true, false)
+
+	var (
+		statedb *state.StateDB
+		release StateReleaseFunc
+	)
+	// if next block does exist, use `StateAtTransaction` with transaction index zero. Otherwise, fallback to use `StateAtBlock` with the input block number
+	if next, err := api.blockByNumber(ctx, rpc.BlockNumber(block.NumberU64())); err != nil {
```
blukat29, line 900: What is this `next`? isn't it the same as `block` local variable? The geth branch (https://github.com/ethereum/go-ethereum/blob/da17f2d65b8922a20f86740644b2d7f6ef688e9b/eth/tracers/api.go#L948-L952) is about TxIndex and we don't take TxIndex, so in our code single StateAtTransaction(block, 0, ... ) should suffice.

blockchain/evm.go
hyunsooda: Thank you fixed. https://github.com/kaiachain/kaia/pull/121/commits/5eeba8bf39278d306526d4f23dd0524a820c436b

node/cn/tracers/api.go
hyunsooda: https://github.com/kaiachain/kaia/pull/121#discussion_r1830537169

node/cn/tracers/api.go
blukat29: Yes `StateAtTransaction(101, 0, ...) equals to StateAtBlock(100, ...)+Initialize(101)`.

Oh, to preserve the semantics, we should do either `StateAtBlock(num)` or `StateAtTransaction(num+1)`...
In that case, can you revert to `StateAtBlock(num)`? Sorry I was confused.

blukat29 (approved):

ian0371 (approved):

PR #129, opened 2024-11-12
test: Add Shanghai tests by `ethereum/execution-spec-tests`

ulbqb:
## Proposed changes

This PR adds tests by `ethereum/execution-spec-tests`. Currently, this supports testing only Shanghai EIPs against each forks (only Cancun and Prague). I'm going to add tests for other forks in other PRs.

Simulating Ethereum behavior is needed because of the difference of spec between Kaia and Ethereum to use `ethereum/execution-spec-tests` as is. This simulate following behavior:

- Gas price - Simulate Ethereum gas prices and set the values ​​directly on the txContext
- Intrinsic gas - Ethereum’s intrinsic gas after Istanbul is the same as Kaia’s Prague
- Op code gas - Change constant gas for specific op code.
- Mining reward - Simualte Ethereum mining reward and set the value directly to the state db.
- State root - Fetch all data from the state db and set the data to a new state db as LegacyAccount

Main code modifications:

- Add SetLegacyAccountForTest to StateDB
- Add relaxPrecompileRangeForTest
- Add ChangeGasCostForTest

Remarks:

- TestExecutionSpecState should be tested separately because of changing global variable (relaxPrecompileRangeForTest)

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


ulbqb:
recheck

ulbqb:

ulbqb:
recheck

blockchain/vm/jump_table.go
```diff
@@ -75,7 +75,6 @@ func newCancunInstructionSet() JumpTable {
 	enable5656(&instructionSet) // EIP-5656 MCOPY opcode
 	enable6780(&instructionSet) // EIP-6780 SELFDESTRUCT only in same transaction
 	enable1153(&instructionSet) // EIP-1153 TLOAD, TSTORE opcode
-	enable1052(&instructionSet) // EIP-1052 EXTCODEHASH fix
```
ian0371, line 78: It seems that you removed this line because EIP-1052 is never activated on Ethereum.  
However, we cannot simply remove this. This will break the backward compatibility.

One way to solve this is override `vm.CancunInstructionSet` during this test.
btw, why is Shanghai test affected by Cancun? Shanghai is older.

Makefile
```diff
@@ -28,10 +28,12 @@ abigen:
 	@echo "Run \"$(BIN)/abigen\" to launch abigen."
 
 test:
-	$(GORUN) build/ci.go test
+	$(GORUN) build/ci.go test -skip ^TestExecutionSpecState$$
```
ian0371, line 31: Is this for testing? If so, you can simply run `cd tests; go test -run=TestExecutionSpecState -v`

tests/state_test.go
```diff
@@ -29,10 +29,14 @@ import (
 	"testing"
 
 	"github.com/kaiachain/kaia/blockchain/vm"
+	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/params"
 )
 
 func TestState(t *testing.T) {
+	common.RelaxPrecompileRangeForTest(false)
+	enableTestExecutionSpecState(false)
```
ian0371, line 38: To not affect other tests:

```suggestion
	common.RelaxPrecompileRangeForTest(false)
	enableTestExecutionSpecState(false)
	defer common.RelaxPrecompileRangeForTest(true)
	defer enableTestExecutionSpecState(true)
```

tests/state_test.go
```diff
@@ -58,21 +62,64 @@ func TestState(t *testing.T) {
...
-				if subtest.Fork == "Constantinople" {
-					t.Skip("constantinople not supported yet")
+		execStateTest(t, st, test, name, []string{"Constantinople"})
+	})
+}
+
+// TestExecutionSpecState runs the test fixtures from execution-spec-tests.
+func TestExecutionSpecState(t *testing.T) {
+	common.RelaxPrecompileRangeForTest(true)
+	enableTestExecutionSpecState(true)
```
ian0371, line 72: To not affect other tests:

```suggestion
	common.RelaxPrecompileRangeForTest(true)
	enableTestExecutionSpecState(true)
	defer common.RelaxPrecompileRangeForTest(false)
	defer enableTestExecutionSpecState(false)
```

tests/state_test_util.go
```diff
@@ -115,6 +118,12 @@ type stTransactionMarshaling struct {
 	PrivateKey hexutil.Bytes
 }
 
+var isTestExecutionSpecState bool
```
ian0371, line 121: Is this correct?
If `isTestExecutionSpecState = false`, test runs `kaiachain/kaia-core-tests`
If `isTestExecutionSpecState = true`, test runs ethereum tests.

ulbqb: Yes. Only if `isTestExecutionSpecState = true`, simulating Ethereum stuff is run.

blockchain/vm/jump_table.go
ian0371: Got it. Thanks.

ulbqb: > It seems that you removed this line because EIP-1052 is never activated on Ethereum.
However, we cannot simply remove this. This will break the backward compatibility.

How about skipping tests related to this problem? I think it's better to test Kaia as is.

Only one test could be skipped.
- `shanghai/eip3651_warm_coinbase/warm_coinbase/warm_coinbase_gas_usage.json/tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_gas_usage[fork_Cancun-state_test-EXTCODEHASH]`

tests/state_test.go
ulbqb: This doesn't seem to work because defer is run before all tests are finished.

Makefile
ulbqb: This is for a CI test. Separating tests is needed because `relaxPrecompileRangeForTest` and `isTestExecutionSpecState` affect overall.

blockchain/vm/jump_table.go
ian0371: You're right. The test should be skipped.

tests/state_test_util.go
```diff
@@ -289,3 +330,117 @@ func rlpHash(x interface{}) (h common.Hash) {
 	hw.Sum(h[:0])
 	return h
 }
+
+// simulate gas price for Ethereum
+func simulateEthGasPrice(config *params.ChainConfig, json *stJSON) error {
```
hyunsooda, line 335: 1. The function seems unpack the `gasPrice`. I was confused because the function name starts with `simluate`. What's the correct semantic you intended?
2. The `gasPrice` is set to `LowerBoundBaseFee`. Can you explain how this is effective?

tests/state_test_util.go
```diff
@@ -115,6 +119,12 @@ type stTransactionMarshaling struct {
 	PrivateKey hexutil.Bytes
 }
 
+var isTestExecutionSpecState bool
```
hyunsooda, line 122: Can you explain which problem was encountered(precompiled contract range mismatching?) or is expected if `isTestExecutionSpecState` disappears?

tests/init.go
```diff
@@ -35,12 +35,43 @@ var Forks = map[string]*params.ChainConfig{
...
+	},
+	"Paris": {
+		ChainID:                  big.NewInt(1),
+		IstanbulCompatibleBlock:  new(big.Int),
+		LondonCompatibleBlock:    new(big.Int),
+		EthTxTypeCompatibleBlock: new(big.Int),
+		MagmaCompatibleBlock:     new(big.Int),
+		KoreCompatibleBlock:      new(big.Int),
+	},
+	"Merge": {
```
hyunsooda, line 67: It seems not related to Kaia. Can this be omited?

common/types.go
```diff
@@ -353,14 +353,24 @@ func BigToAddress(b *big.Int) Address { return BytesToAddress(b.Bytes()) }
...
 func HexToAddress(s string) Address { return BytesToAddress(FromHex(s)) }
 
+var relaxPrecompileRangeForTest bool
+
 // IsPrecompiledContractAddress returns true if the input address is in the range of precompiled contract addresses.
 func IsPrecompiledContractAddress(addr Address) bool {
+	if relaxPrecompileRangeForTest {
+		return false
+	}
 	if bytes.Compare(addr.Bytes(), lastPrecompiledContractAddressHex) > 0 || addr == (Address{}) {
```
hyunsooda, line 363: @blukat29 If the `Validate`(`tx_pool.go`) interface signature is changed to accept chain rule, the condition L363 may disappear as well as `relaxPrecompileRangeForTest`. What do you think?

tests/state_test_util.go
```diff
@@ -289,3 +330,117 @@ func rlpHash(x interface{}) (h common.Hash) {
...
+			return 0, types.ErrGasUintOverflow
+		}
+		gas += z * params.TxDataZeroGas
+	}
+
+	return gas, nil
+}
+
+// simulate staking reward for Ethereum
+func simulateEthStakingReward(statedb *state.StateDB, evm *vm.EVM, tx *stTransaction, baseFee *big.Int, usedGas uint64, gasPrice *big.Int) {
```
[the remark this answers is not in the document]
ulbqb: It seems better.

tests/state_test_util.go
ulbqb: Kaia limits address range of precompiled contracts by [this function](https://github.com/kaiachain/kaia/blob/08f4f7037822b1c255d40c90bdbde38da5382cf4/common/types.go#L354-L359). But some test cases set contracts whose address is like 0x00…00100 on setting pre state. So it is recognized as a reserved precompiled address on executing a transaction and then [this error](https://github.com/kaiachain/kaia/blob/08f4f7037822b1c255d40c90bdbde38da5382cf4/kerrors/kerrors.go#L36) is returned.

tests/init.go
ulbqb: Merge is used. On the other hand, Paris wasn't used so I'll remove Paris (AFAIK, Paris is equivalent to Merge.).

common/types.go
blukat29: 1. Where in tx_pool.go? IsPrecompiledContractAddress is used in tx_internal_data_*.go:Validate. e.g. 
https://github.com/kaiachain/kaia/blob/178185beb4c576c6efbed3baf2171ae2db8ac599/blockchain/types/tx_internal_data_legacy.go#L362-L368

2. We don't plan to change the boundary after a hardfork. It only changes during testing. `relaxPrecompileRangeForTest` toggles between Kaia and Eth. Not between, say, Prague or Cancun.
```go
// NOT going to do this
func (t *TxInternalDataLegacy) Validate(stateDB StateDB, num uint64) error {
	if t.Recipient != nil {
		if common.IsPrecompiledContractAddress(*t.Recipient) && !fork.Rules(num).IsPrague {
			return kerrors.ErrPrecompiledContractAddress
		}
	}
	return t.ValidateMutableValue(stateDB, currentBlockNumber)
}
```

tests/state_test.go
```diff
@@ -29,10 +29,14 @@ import (
 	"testing"
 
 	"github.com/kaiachain/kaia/blockchain/vm"
+	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/params"
 )
 
 func TestState(t *testing.T) {
```
blukat29, line 36:
```suggestion
// TestCoreSpecState runs the StateTests fixtures from kaia-core-tests
func TestKaiaSpecState(t *testing.T) {
```

tests/state_test.go
```diff
@@ -58,21 +62,67 @@ func TestState(t *testing.T) {
...
-			key := fmt.Sprintf("%s/%d", subtest.Fork, subtest.Index)
-			name := name + "/" + key
-			t.Run(key, func(t *testing.T) {
-				if subtest.Fork == "Constantinople" {
-					t.Skip("constantinople not supported yet")
+		execStateTest(t, st, test, name, []string{"Constantinople"})
+	})
+}
+
+// TestExecutionSpecState runs the test fixtures from execution-spec-tests.
```
blukat29, line 69:
```suggestion
// TestExecutionSpecState runs the state_test fixtures from execution-spec-tests.
```

tests/state_test_util.go
blukat29: Maybe we should call it something like `isEthSpecCompatibleMode` or something.. cc @hyunsooda

common/types.go
hyunsooda: 1. I mispointed it, `tx_internal_data.go`. 
2. The boundary check condition seems unnecessary, and it has been exploited in various tx types like adhoc because this does not involve any prerequisite and is too simple to call it. Once I compared it with the Geth implementation, I came up with a way to replace the `Validate` interface and remove the boundary condition, which was originally the right way to validate a precompiled contract address.
Plus, `relaxPrecompileRangeForTest` can be removed. This was my initial view.

tests/state_test.go
ian0371: Then can you pass the variable via argument?
I think there needs to be two entrypoints:
```
func TestStateWithKaiaCoreTests() { execStateTest(..., simulate=false) }
func TestStateWithEthereum() { execStateTest(..., simulate=true) }
```

tests/state_test_util.go
```diff
@@ -361,12 +361,12 @@ func simulateEthGasPrice(config *params.ChainConfig, json *stJSON) error {
 
 // simulate intrinsic gas amount for Ethereum
 func simulateEthIntrinsicGas(intrinsicGas uint64, data []byte, r params.Rules) (uint64, error) {
-	if r.IsIstanbul {
+	if r.IsIstanbul && !r.IsPrague {
```
hyunsooda, line 364: It seems a contradiction because L369 sets `Prague`. I didn't intend to change this initially. Can you explain how it was derived?

blockchain/vm/jump_table.go
blukat29: FYI we patched 1052 in Kaia's Cancun version https://github.com/klaytn/klaytn/releases/tag/v1.12.0

Makefile
blukat29: I think we are not running in parallel, you can run them altogether, of you defer SetRelax...(false)

tests/init.go
hyunsooda: Did you mean `Merge` as Ethereum's 'The Merge'? If so, I wonder if those tests are effective in Kaia.

common/types.go
hyunsooda: For example,

```diff
-func (t *TxInternalDataFeeDelegatedSmartContractDeploy) Validate(stateDB StateDB, currentBlockNumber uint64) error {
+func (t *TxInternalDataFeeDelegatedSmartContractDeploy) Validate(stateDB StateDB, currentBlockNumber uint64, rules params.Rules) error {
        var to common.Address
        if t.Recipient != nil {
                return kerrors.ErrInvalidContractAddress
        } else {
                to = crypto.CreateAddress(t.From, t.AccountNonce)
        }
-       if common.IsPrecompiledContractAddress(to) {
+       // if common.IsPrecompiledContractAddress(to) {
+       //      return kerrors.ErrPrecompiledContractAddress
+       // }
+       if vm.IsPrecompiledContractAddress(to, rules) {
                return kerrors.ErrPrecompiledContractAddress
        }
        if t.HumanReadable {
```

```diff
+func IsPrecompiledContractAddress(addr common.Address, rules params.Rules) bool {
+       evm := EVM{chainRules: rules}
+       precompiles := evm.GetPrecompiledContractMap(addr)
+       _, ok := precompiles[addr]
+       return ok
+}
+
```

Even though the simple pseudo-code has an import cycle, eventually, this seems to remove the ad-hoc comparison and retrieve the precompiled contract address map, compare with it, and return the `true` or `false` that Geth does.

tests/state_test_util.go
ulbqb: If Kaia is operating as the equivalent of Prague, this simulating can be skip because `kaiaPayloadGas == ethPayloadGas`. 

https://github.com/ethereum/go-ethereum/blob/v1.14.11/core/state_transition.go#L91-L93
`Prague` is set in L369 in order to calculate the payload gas of Ethereum after Istanbul.

tests/init.go
ulbqb: Yes.

Actually, this is just the config in the case where Kaia behaves as Ethereum Merge. This means that we can say Kaia has Ethereum Merge compatibility if tests pass using this config.

tests/state_test_util.go
ulbqb: > The function seems unpack the gasPrice. I was confused because the function name starts with simluate. What's the correct semantic you intended?

Here sets the config so that tx senders pay as much fee as Ethereum. AFAIK, Kaia's gas price is base on only chain config not tx contains. Therefore gas price is calculated in Ethereum context, and the value is set to the config.

common/types.go
blukat29: If `IsPrecompiledContractAddress` intends to detect [currently existing precompiles], then your approach is the accurate one. I inferred (not sure though) that the function intends to protect a 'reserved precompile range 0x1~0x3ff'. See https://github.com/kaiachain/kaia/blob/178185beb4c576c6efbed3baf2171ae2db8ac599/blockchain/vm/evm.go#L530 where contract creation in the range is prohibited.

Similar EIPs has been proposed https://eips.ethereum.org/EIPS/eip-1352, though not explicitly enforced in Eth.

I think we need to ratify if we want to protect the range, or protect the concrete precompile instances.

tests/state_test_util.go
hyunsooda: I think `intrinsicGas - kaiaPayloadGas` always becomes zero. Before Prague, so, only the ether logic can be considered maybe.

ulbqb: I see. How about this? I think this is more intrinsic.

```golang
func simulateEthIntrinsicGas(data []byte, contractCreation bool, r params.Rules) (uint64, error) {
	if r.IsIstanbul {
		r.IsPrague = true
	}
	return types.IntrinsicGas(data, nil, contractCreation, r)
}
```

hyunsooda: Looks nice, but can you consider changing the function name such as `useEthIntrinsicGas` because `simulate` seems bloated?

Makefile
ulbqb: https://github.com/kaiachain/kaia/pull/129#discussion_r1841401749
I couldn't use defer because of this.

tests/state_test_util.go
ulbqb: https://github.com/ulbqb/kaia/blob/04126d29cbe605ae574d1272761bbab947adcafb/tests/state_test_util.go#L187
It's difficult to reproduce the gas price of Ethereum from the chain config so I'm going to directly set the gas price to txContext.

common/types.go
hyunsooda: Understood. The IsPrecompiledContractAddress function enforces the allowed range for precompiled contracts, rather than relying on an ad-hoc comparison. Let me close it now, and another issue can be opened if needed.

Makefile
hyunsooda: If we use mutex, can this be reverted?

hyunsooda: I tested it below and successfully done:

```diff
-  $(GORUN) build/ci.go test -skip ^TestExecutionSpecState$$
- $(GORUN) build/ci.go test -run ^TestExecutionSpecState$$

+ $(GORUN) build/ci.go test    You, 24/11/18 • Uncommitted changes
```

What error leads you to branching with two executions?

ulbqb:
## Update

- Enable to change opcode gas cost only during tests (`useEthOpCodeGas`) - This is needed if Kaia adopts different gas cost of opcode from Ethereum such as EXTCODEHASH.
- Enable to running node as only Cancun and Prague only during tests - It's meaningless to test Kaia as old forks from now. And Randao, Kip103, and Kaia forks are run as Cancun.

Makefile
ulbqb: I don't know why but TestKaiaSpecState failed. Some of the errors is following:

```
--- FAIL: TestKaiaSpecState (2.14s)
    --- FAIL: TestKaiaSpecState/stSolidityTest/TestCryptographicFunctions.json (0.00s)
        --- FAIL: TestKaiaSpecState/stSolidityTest/TestCryptographicFunctions.json/Byzantium/0 (0.00s)
            state_test.go:132: post state root mismatch: got 349489d4e80a0456a178264c7bcf402dba239a4ac6d5aa62b03bb6b645409e6c, want 79f49f0fa14e21191f7513eed2df22be22474565fbe230776b8fc8b46d7e7fd4
            state_test.go:134: gas limit too high for EVM trace
    --- FAIL: TestKaiaSpecState/stReturnDataTest/create_callprecompile_returndatasize.json (0.01s)
        --- FAIL: TestKaiaSpecState/stReturnDataTest/create_callprecompile_returndatasize.json/Byzantium/0 (0.00s)
            state_test.go:132: post state root mismatch: got e988bab3dc9bbbe1eb5d415237f26ed6152a22e83ca77510c77b5b6c5440e34a, want 5546052355a4c86deacb9fe9e0d6ac9f6ff7ae16b3c2900ca87bad7926109d4d
            state_test.go:134: gas limit too high for EVM trace
    --- FAIL: TestKaiaSpecState/stReturnDataTest/call_ecrec_success_empty_then_returndatasize.json (0.00s)
        --- FAIL: TestKaiaSpecState/stReturnDataTest/call_ecrec_success_empty_then_returndatasize.json/Byzantium/0 (0.00s)
            state_test.go:132: post state root mismatch: got 019e9af82544c1354885801c7ad431353abcdd58fdc5aaee141ef6c6d9bccc8a, want 21be96a714bbd3db07c2e3d063b4ed5e3470190b7fa576e45353e83d7f33830e
            state_test.go:134: gas limit too high for EVM trace
```

ulbqb: This problem can be solved by removing `t.Parallel()` at L40 in `state_test.go`.

hyunsooda: @blukat29 `t.Parallel()` seems not much effective and shows no difference in my local. What do you think?

tests/state_test_util.go
```diff
@@ -366,6 +371,13 @@ func useEthGasPrice(config *params.ChainConfig, json *stJSON) (*big.Int, error)
 	return gasPrice, nil
 }
 
+func useEthOpCodeGas(r params.Rules, evm *vm.EVM) {
+	if r.IsCancun {
+		// EIP-1052 must be retained for backward compatibility.
+		vm.ChangeGasCostForTest(&evm.Config.JumpTable, vm.EXTCODEHASH, params.WarmStorageReadCostEIP2929)
+	}
+}
+
```
hyunsooda, line 380: Could you explain how it is introduced?

Makefile
ulbqb: Do we need to ensure that other tests are not running in parallel? If so, it's not easy...

ulbqb: @hyunsooda 
Can you reproduce it by removing cache?

tests/state_test_util.go
ulbqb: https://github.com/kaiachain/kaia/pull/129#discussion_r1841399589

This is related to this comment. The constant gas of EXTCODEHASH is different because EIP-1052 is never activated on Ethereum. The relevant tests were being skipped, but I fixed it so that they can be run with a constant gas change.

Makefile
hyunsooda: > Do we need to ensure that other tests are not running in parallel? If so, it's not easy...

I belive that `t.Parallel()` is effective in the scope of a specific test. If so, `TestKaiaSpecState` is only the target  of the sequential execution.

> Can you reproduce it by removing cache?

Do you mean `build/cache` directory? If so, I removed it and did `make test` with comment out `t.Parallel()`.

tests/state_test_util.go
hyunsooda: Thanks for the information.

> // EIP-1052 must be retained for backward compatibility.

Then, the comment seems like it needs to be clarified. What do you think about this?
`// EIP-1052 must be inactivated for backward compatibility.`

Makefile
ulbqb: > I belive that t.Parallel() is effective in the scope of a specific test. If so, TestKaiaSpecState is only the target of the sequential execution.

When I tried the following code, Test_Func2 could be finished after Test_Func3 started.

```golang
package tests

import (
	"fmt"
	"testing"
)

func trace(name string) func() {
	fmt.Printf("%s entered\n", name)
	return func() {
		fmt.Printf("%s returned\n", name)
	}

}

func Test_Func1(t *testing.T) {
	defer trace("Test_Func1")()

	// ...
}

func Test_Func2(t *testing.T) {
	defer trace("Test_Func2")()
	t.Parallel()

	// ...
}

func Test_Func3(t *testing.T) {
	defer trace("Test_Func3")()

	// ...
}
```

> Do you mean build/cache directory? If so, I removed it and did make test with comment out t.Parallel().

Please use `go clean -testcache`.

blukat29: > Do we need to ensure that other tests are not running in parallel?

I vaguely remember that most tests are not parallel.. We can simply search for `t.Parallel`

hyunsooda: I'm a little bit confused. If the test is executed in parallel, the function execution order does matter because this PR uses and updates a global variable, but I thought that we were commenting out `t.Parallel()` and expected that `TestExecutionSpecState` and `TestKaiaSpecState` would not corrupt each other. Can you pinpoint the current situation you are encountering so that I can understand more?

tests/state_test_util.go
ulbqb: I'll add the details.

Makefile
ulbqb: @hyunsooda 
The problem is running `TestKaiaSpecState` with `relaxPrecompileRangeForTest = true`. This is caused by starting `TestExecutionSpecState` before finishing `TestKaiaSpecState` due to `t.Parallel()`.

I think dropping `t.Parallel()` in all tests and resetting `relaxPrecompileRangeForTest` after finishing `TestExecutionSpecState` are needed so that `TestKaiaSpecState` (`common.RelaxPrecompileRangeForTest(true)`) doesn't affect other tests.

ulbqb: I understood the spec of `t.Parallel()`. 

The test pauses when `t.Parallel()` is run and then next test starts. All tests which pause by `t.Parallel()` resume in parallel after all tests are done.

Current problem:
1. run `common.RelaxPrecompileRangeForTest(false)` in `TestKaiaSpecState`
2. run `t.Parallel()` and `TestKaiaSpecState` pauses
3. run `TestExecutionSpecState` and `common.RelaxPrecompileRangeForTest(true)`
4. resume `TestKaiaSpecState` with `relaxPrecompileRangeForTest = true`

Simple solution is running `common.RelaxPrecompileRangeForTest(false)` after `t.Parallel()` in `TestKaiaSpecState`. But the root cause is not running `common.RelaxPrecompileRangeForTest(false)` at the end of `TestExecutionSpecState`. So I'll implement this using Test Suite.

tests/state_test_util.go
ulbqb: I fixed this at https://github.com/kaiachain/kaia/pull/129/commits/0afb6440cf1ae0c894ae10398a31cfa61b0903a7.

Makefile
hyunsooda: Got it, thank you. The other tests seem to have been still omitted in the `Makefile`, don't they?

ulbqb: I fixed it at https://github.com/kaiachain/kaia/pull/129/commits/e98fb5f4f5c73976f32baf88f1f1a6e8851eca0b.

tests/state_test_util.go
ulbqb: Is your proposal about a variable name?

blockchain/vm/jump_table.go
ulbqb: I add changing gas cost of opcode to enable to run the skipped test.
commit: https://github.com/kaiachain/kaia/pull/129/commits/831c67d787c6f9ecc866b99da0c462e69d36329e

Makefile
hyunsooda: @blukat29 @ian0371 This thread can be closed as `Makefile` is reverted and remains unchanged.

tests/state_test_util.go
```diff
@@ -274,9 +313,17 @@ func (tx *stTransaction) toMessage(ps stPostState, r params.Rules) (blockchain.M
...
+	if isTestExecutionSpecState {
+		intrinsicGas, err = useEthIntrinsicGas(data, to == nil, r)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		intrinsicGas, err = types.IntrinsicGas(data, nil, to == nil, r)
+		if err != nil {
+			return nil, err
+		}
```
hyunsooda, line 326:
```suggestion
	var intrinsicGas uint64
	if isTestExecutionSpecState {
		intrinsicGas, err = useEthIntrinsicGas(data, to == nil, r)
	} else {
		intrinsicGas, err = types.IntrinsicGas(data, nil, to == nil, r)
         }
         if err != nil {
		return nil, err
	}
```

hyunsooda: As-Is:
if ether's gas calc fails:
  set error to `err` and return
if kaia's gas calc fails:
  set error to `err` and return

I thought that this is equiv with the suggested change.
if ether's gas calc fails:
  set error to `err`
if kaia's gas calc fails:
  set error to `err`
if err is not nil return

Thus, I think there is no missing information with this change. Please teach me if there's another point .

ulbqb: It was my mistake. This applies.
commit: https://github.com/kaiachain/kaia/pull/129/commits/11899f7495a691917fbc01e031a213e6ef88a205

blukat29 (approved):

hyunsooda (approved):

ian0371 (approved):

PR #147, opened 2024-11-26
cmd: Add flags for tx and receipt pruning

2dvorak:
## Proposed changes

- This PR adds flags for enabling tx/receipt pruning and setting their pruning retention
- Actual logics for the pruning will be added in the follow-up PR

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

none


cmd/utils/config.go
```diff
@@ -604,6 +604,18 @@ func (kCfg *KaiaConfig) SetKaiaConfig(ctx *cli.Context, stack *node.Node) {
...
+		cfg.TxPruningRetention = ctx.Uint64(TxPruningRetentionFlag.Name)
+	} else {
+		cfg.TxPruningRetention = 0
+	}
+	cfg.ReceiptPruning = ctx.Bool(ReceiptPruningFlag.Name)
+	if cfg.ReceiptPruning {
+		cfg.ReceiptPruningRetention = ctx.Uint64(ReceiptPruningRetentionFlag.Name)
+	} else {
+		cfg.ReceiptPruningRetention = 0
+	}
```
hyunsooda, line 618: [About value constraint]

- Even if both retention values differ, I guess no issue will be found. Do you think another point of view about it?
- Is it okay if both values are lower than the `LivePruningRetention` value?
- A plan of upper and lower bound?

blukat29: - I think they can differ. At least I know an extreme case where deleting all receipts and block syncs well. In pruning terms, TxPruningRetention = infinity and ReceiptPruningRetention = 0.
- Maybe? The Retention value only affects the [pruneTrieNodeLoop](https://github.com/kaiachain/kaia/blob/35e43b859b658732323d3f67264c46a1475016a0/blockchain/blockchain.go#L1509) and it is unrelated to txs or receipts in the database.
- No need for the upper bound. You can set 9999999999 to signify infinity. However, a lower bound could be a good safety measure. e.g. any arbitrary number like 128.

@2dvorak Can we have the lower bound for all 3 (tx, receipt, state live pruning) retentions?

blukat29: @2dvorak @hyunsooda  I think separate lower-bound PR is better. So I can mention it in the Release note.

blukat29:
@2dvorak Rebase to the latest dev to fix linter.

cmd/utils/config.go
2dvorak: @hyunsooda @blukat29 OK, I'll create another PR for lower-bounds. Thanks!

blukat29 (approved):

ian0371 (approved):

cmd/utils/config.go
blukat29: @hyunsooda PTAL

hyeonLewis (approved):

hyunsooda (approved):

PR #229, opened 2025-01-24
Set completion mark in correct timing

hyeonLewis:
## Proposed changes

When state migration has been finished, it removes old DB and dbm for state migration. But setting the completion mark `inMigration = false` is done after `dbm.dbs[StateTrieMigrationDB] = nil` in separate go-routine (no lock held), which leads to potential nil dereference due to timing issue while opening a new trie (since it accesses to `dbm.dbs[StateTrieMigrationDB]` if `inMigration = true`).

I strongly suspect it's the root cause of the random test failure of `TestMigration_StartMigrationByMiscDBOnRestart` in CI.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

The CI failure always happens in #225.

### DIFF
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -831,10 +831,6 @@ func (dbm *databaseManager) FinishStateMigration(succeed bool) chan struct{} {
 	dbm.setDBDir(StateTrieDB, dbDirToBeUsed)
 	dbm.dbs[StateTrieDB] = dbToBeUsed
 
-	// Set the completion mark.
-	// The old database will be completely removed (possibly not be removed if error occurs)
-	dbm.setStateTrieMigrationStatus(0)
-	dbm.dbs[StateTrieMigrationDB] = nil
 	dbm.setDBDir(StateTrieMigrationDB, "")
 
 	dbPathToBeRemoved := filepath.Join(dbm.config.Dir, dbDirToBeRemoved)
@@ -850,6 +846,10 @@ func (dbm *databaseManager) FinishStateMigration(succeed bool) chan struct{} {
 // Remove old database. This is called once migration(copy) is done.
 func (dbm *databaseManager) removeOldDB(dbPath string, endCheck chan struct{}) {
 	defer func() {
+		// Set the completion mark.
+		// The old database will be completely removed (possibly not be removed if error occurs)
+		dbm.setStateTrieMigrationStatus(0)
+		dbm.dbs[StateTrieMigrationDB] = nil
 		if endCheck != nil {
 			close(endCheck)
 		}
--- a/storage/database/db_manager.go
+++ b/storage/database/db_manager.go
@@ -831,6 +831,9 @@ func (dbm *databaseManager) FinishStateMigration(succeed bool) chan struct{} {
 	dbm.setDBDir(StateTrieDB, dbDirToBeUsed)
 	dbm.dbs[StateTrieDB] = dbToBeUsed
 
+	// Set the completion mark.
+	// The old database will be completely removed (possibly not be removed if error occurs)
+	dbm.setStateTrieMigrationStatus(0)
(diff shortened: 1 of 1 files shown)


ian0371 (commented):
Based on PR description, before accessing `dbm.dbs[StateTrieMigrationDB]`, isn't the read lock checked? Could you show me which line?
+ this PR seems correct because the lock is not held in `removeOldDB`.

ian0371 (dismissed):

hyeonLewis:
@ian0371 https://github.com/kaiachain/kaia/blob/dev/storage/database/db_manager.go#L1841. And the lock is not held in `removeOldDB`.

hyunsooda:
If the completion mark is forward back, previous bad ux may remain (https://github.com/klaytn/klaytn/issues/1412). I think these three lines https://github.com/kaiachain/kaia/pull/229/files#diff-3f4e98143300d776da515f7e8784f500929242d2fd1705aa237cdb76aa84d638R836-R838 should be invoked in one operation. What do you think if the other two lines are moved together in L851?

hyeonLewis:
@hyunsooda Thanks for supplying contexts, and I think moving all three lines into defer in `removeOldDB` also works. @ian0371 What do you think?

ian0371:
We should figure out the dependencies of each resource, and figure out the deallocation order.

1. Invalidate memory `dbm.inMigration`; `dbm.inMigration = false`
2. Invalidate memory; `dbm.dbs[StateTrieMigrationDB] = nil`. Ensure that it is not accessed when `dbm.inMigration` is false.
3. Remove DB; `removeOldDB()`
4. Invalidate MigrationDB path; `dbm.setDBDir(StateTrieMigrationDB, "")`. Ensure that it is not accessed when `dbm.inMigration` is false.
5. Write migration status; `dbm.setStateTrieMigrationStatus(0)`

So the invariant is that MigrationDB can be accessed only when `dbm.inMigration` is true.

If the program halts during step 3, resume from 3 on restart. Maybe we need to add "ShouldRemoveOldDB" to indicate that removing DB is not finished.

hyeonLewis:
@ian0371 @hyunsooda Currently, the node restarts the state migration if the program exits/halts before deleting the old DB. And if we call `setStateTrieMigrationStatus(0)` early, the issue lake solved occurred again. 

The fundamental solution would be storing deletion info into DB as Ian's suggestion, and it can be handled in separate PR. I'll remain `setDBDir` to maintain the same [behavior](https://github.com/kaiachain/kaia/blob/dev/storage/database/db_manager.go#L520) when restarting the node.

Before that, I patched to make dbs nil after `inMigration = false`. PTAL.

hyeonLewis:
@hyunsooda Reading state trie from `StateTrieMigrationDB` must be prevented after `isMigration = false`. Could you provide which code has an inconsistency issue?

hyunsooda (approved):
Thank you, got the point. Following PR expects that recorgnize the end of migration work and directly jump to remove database if it has not been deleted.

yoomee1313 (approved):

ian0371:
I think reverting back to old file is correct (both `FinishStateMigration` and `removeOldDB`).
https://github.com/klaytn/klaytn/blob/5b050ee5243962da863b8bd627401f3d72d47a8f/storage/database/db_manager.go#L749-L751

If `inMigration` is true, reading and writing trie node happens on two DBs, trieDB and migrateDB.
At `FinishStateMigration()`, it means that migrateDB is ready to use and can replace trieDB. Thus, after calling `FinishStateMigration()`, there's no reason for `inMigration` flag to be true. That's why nil dereference took place in the first place.

All we need to do is remove old database in the `removeOldDB` goroutine, and make it continue on restart.

ian0371 (approved):

PR #216, opened 2025-01-21
feat: Update KIP-228 (EIP-7702) 

tnasu:
## Proposed changes

After introducing [KIP-228(EIP-7702)](https://github.com/kaiachain/kaia/pull/172), the reference of Geth is updated with the latest EIP spec

Changes:
- Update to call `AddAddressToAccessList` once only
- Use `GetCodeHash` when creating contract to check deployable 
- Should rm dynamic gas cost for reading opcode (`EXTCODE*`) for EOA with code
- Updated spec only return `0xef01`, not whole delegation designatior
  - It may be revert 
- Refactor the codes

Remaining updates:
- ChainId
  - https://github.com/ethereum/EIPs/pull/9143
  - https://github.com/ethereum/EIPs/pull/9152
  - https://github.com/ethereum/go-ethereum/pull/30982
  - https://github.com/ethereum/go-ethereum/pull/31032
  - https://github.com/ethereum/go-ethereum/pull/31054
  - https://github.com/ethereum/go-ethereum/pull/30937
- EXTCODE
  - https://github.com/ethereum/EIPs/pull/9250

## Types of changes

- [ ] Bugfix
- [x] New feature or enhancement
- [ ] Others

## Checklist

- [x] I have read the [CONTRIBUTING GUIDELINES](https://github.com/kaiachain/kaia/blob/main/CONTRIBUTING.md) doc
- [x] I have read the [CLA](https://gist.github.com/kaiachain-dev/bbf65cc330275c057463c4c94ce787a6) and signed by comment `I have read the CLA Document and I hereby sign the CLA` in first time contribute
- [x] Lint and unit tests pass locally with my changes (`$ make test`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- https://github.com/kaiachain/kaia/issues/106

PR #233, opened 2025-01-24
feat: support uint256 for rlp, log, hex, vm, abi

tnasu:
## Proposed changes

Support uint256 for RLP, Hex, etc. These changes also help to update EIP-7702, which uses uint256

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- https://github.com/kaiachain/kaia/issues/106
- https://github.com/kaiachain/kaia/pull/216


blukat29 (dismissed):

rlp/rlpgen/gen.go
```diff
@@ -635,6 +678,9 @@ func (bctx *buildContext) makeOp(name *types.Named, typ types.Type, tags rlpstru
 		if isBigInt(typ) {
 			return bigIntOp{}, nil
 		}
+		if isUint256(typ) {
+			return uint256Op{pointer: true}, nil
```
Mdaiki0730, line 682: Why is this pointer true?

ulbqb:
Why is this needed? Doesn't the other txs use uint256?

rlp/rlpgen/gen.go
tnasu: Thanks, I've missed it! I'll fix

https://github.com/ethereum/go-ethereum/pull/26898/files#diff-1bfd65d008fe8358562b840df33d80ef9210d68f9f295fc9558d644a45f6396cR682

tnasu:
@ulbqb

>Doesn't the other txs use uint256?

What are the other txs? If the other txs isn't `SET_CODE_TX(EIP-7702)`, then there is no plan to use `uinit256` for the other txs. If we want to use `uint256`, we should support `uint256` like this PR.
On the other hand, I can also say this PR isn't needed since we can use `big.Int` instead of `uint256`.

But, I think using `uint256` is popular usage in Geth and the others since it is better performance than `big.Int`. Why don't we use it?

What is your opinion?

Thanks

rlp/rlpgen/gen.go
Mdaiki0730: By introduce [this fix](https://github.com/kaiachain/kaia/pull/233/commits/4680c5eb46b3ad0fe256a999e4bc5bce83729531), it seems we need to fix [uint256.out.txt](https://github.com/kaiachain/kaia/pull/233/files#diff-782511bf28abf012ba50c201bef716f02aaa5a2558633755f9f2743382565ab4).

ulbqb:
> On the other hand, I can also say this PR isn't needed since we can use big.Int instead of uint256.

I see. I was curious about if type 1 or 2 have already use uint256 fields. IMO, uint256 should be applied for the other txs unless there is huge impact. If it will be done by geth, there is no reason why we don't apply.

ulbqb (approved):

Mdaiki0730 (approved):

blukat29 (approved):

tnasu:
@ulbqb 

> I was curious about if type 1 or 2 have already used uint256 fields
We don't use uint256 for existing transaction types because of need this PR.

> uint256 should be applied for the other txs unless there is huge impact
It's hard to say whether it has a huge impact or not. I'm making this PR by searching Geth PRs.
https://github.com/search?q=repo%3Aethereum%2Fgo-ethereum+uint256++is%3Amerged&type=pullrequests&p=1

It's hard to introduce using uint256, and we have to maintain related uint256 after merging this PR.

> If it will be done by geth, there is no reason why we don't apply.
Geth is using uint256 for txs after blob_tx. The other txs are using bigInt
* https://github.com/ethereum/go-ethereum/blob/master/core/types/tx_blob.go
* https://github.com/ethereum/go-ethereum/blob/master/core/types/tx_setcode.go

Issue #152, opened 2024-11-28
Add Ethereum compatibility test


ulbqb:
## Problem

Kaia isn't tested about Ethereum compatibility.

## Solution

Kaia can be tested using ethereum/tests` and `ethereum/execution-spec-test` with some modification.

## Todo

- [x] Add stateTest from `ethereum/execution-spec-test`
  - [x] Prague
    - [x] except 7702: #154
    - [x] 7702: #172
  - [x] Cancun: #154
  - [x] Shanghai: #129
  - [x] Berlin - Frontier: #164
- [x] Investigating the difference between blockchainTest & stateTest: https://github.com/kaiachain/kaia/issues/152#issuecomment-2530402638
- [x] Add blockchainTests from `ethereum/execution-spec-test`
  - [x] post state test: #238
  - [ ] other tests
- [x] Investigating the difference between `ethereum/tests` & `ethereum/execution-spec-test`: https://github.com/kaiachain/kaia/issues/152#issuecomment-2720118377
- [x] ~~Add tests using `ethereum/tests` if it's needed~~
- [ ] Add the detail to README #241

sub tasks:

- [x] Replace execution-spec-test file with latest major version #245
- [x] too many isTestExecutionSpecState branches, separate test runner like Run and runLegacy
- [ ] move useEth functions to eest_utils.go
- [x] share the difference processing between state and block, and use the useEth function for all
- [ ] enable CancunToPragueAtTime15k on block tests
- [ ] gaslimit issue

ulbqb:
# Investigating the difference between blockchainTest & stateTest

TBU

Mdaiki0730:
I will report the problem I encountered when running ETH blockchain testing on kaia.

## Blockchain test flow
1. Calculate genesis from json
2. Compare the genesis block hash and state root with the expected in json
3. Create a Kaia blockchain object and insert a json block RLP into it.
4. After inserting the block, compare the expected state (post state) in the JSON with the actual state.
5. Validate the header of the inserted block

## How adopt kaia
There are two solutions to this problem.
**1. prepare following components**
- A converter that converts `blocks fields` to Kaia blocks
  - A component that replicates the kaia state root
  - A component that calculates the Kaia header from the ETH header.
- A converter that converts Kaia blocks to eth blocks
  - A component that replicates the eth state root (can we use state_test?)
  - A component that calculates the eth header from the kaia header.
[...]

PR #238, opened 2025-01-28
tests: Enable EEST blockchainTests

Mdaiki0730:
## Proposed changes

Modify the eest blockchain test flow to run it on all forks (except 2537 and 7702). This time, enable state and storage checks, and comment out header validation.

### Test flow
1. Calculate genesis from json
2. Compare the genesis block hash and state root with the expected in json
3. Create a Kaia blockchain object and insert blocks created from RLP data.
4. After inserting the block, compare the expected state (post state) in the JSON with the actual state.
5. Validate the part of header of the inserted block

### Changes
- ./tests
  - Add a test engine that reproduces Ethereum gas fees and rewards. 
  - Process to decode Eth Block RLP to Kaia TX
  - Change IsPrecompiledContractAddress to a global variable
  - Divide useEthGasPrice into a json modification part and a calculation part
  - Divide useEthMiningReward into a json modification part and a calculation part
- ./blockchain
  - Add a call to BeforeApplyMessage in blockchain.ApplyTransaction
  - Add a function to chain maker to add Tx to a block regardless of the occurrence of an error.

## Impact to the main logic

- blockchain.ApplyTransaction
  - If the engine implements the BeforeApplyMessage method, it will be called immediately before ApplyMessage with blockchain.ApplyTransaction. Existing engines do not implement this method and so are not affected.
- IsPrecompiledContractAddress
  - Some tx that use IsPrecompiledContractAddress for validation will provide an additional rules argument to use it.
  - If you write a process to update this during main code execution, it could cause problems with your validation logic.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

- #152
  - Add blockchainTests from ethereum/execution-spec-test for new EIPs if it's needed
  - Add blockchainTest from ethereum/execution-spec-test for old EIPs if it's needed (except 2537 and 7702)

## Further comments


ulbqb:
Please describe impact to the main logic (IsPrecompiledContractAddress and BeforeApplyMessage).

Mdaiki0730:
> Please describe impact to the main logic (IsPrecompiledContractAddress and BeforeApplyMessage).

@ulbqb I added it to PR's description, please check it!

ulbqb:
@Mdaiki0730 
How about describing it in another section? It's not that all modification affects the main logic.

Mdaiki0730:
That's good. Changed it to new section.

blockchain/vm/evm.go
```diff
@@ -261,8 +261,7 @@ func (evm *EVM) Call(caller types.ContractRef, addr common.Address, input []byte
 	)
 
 	// Filter out invalid precompiled address calls, and create a precompiled contract object if it is not exist.
-	// Because IsPrecompiledContractAddress checks for 0..0x400 and ConsoleLog address is outside of the range, we add one more condition if UseConsoleLog.
-	if IsPrecompiledContractAddress(addr, evm.chainRules) || (addr == consoleLogContractAddress && evm.Config.UseConsoleLog) {
+	if common.IsPrecompiledContractAddress(addr, evm.chainRules) {
```
hyeonLewis, line 264:
```suggestion
        // Because IsPrecompiledContractAddress checks for 0..0x400 and ConsoleLog address is outside of the range, we add one more condition if UseConsoleLog.
        if common.IsPrecompiledContractAddress(addr, evm.chainRules) || (addr == consoleLogContractAddress && evm.Config.UseConsoleLog) {
```

tests/block_test_util.go
```diff
@@ -236,13 +356,63 @@ func (t *BlockTest) validatePostState(statedb *state.StateDB) error {
...
+
+	err := t.validateStorage(statedb)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateStorage validates storage while considering the difference between Kana and Ethereum.
```
shiki-tak, line 377: What is `Kana` ? `Kaia` ?

blockchain/vm/evm.go
Mdaiki0730: I accidentally removed it, thank you!

blockchain/genesis.go
```diff
@@ -310,7 +310,17 @@ func (g *Genesis) ToBlock(baseStateRoot common.Hash, db database.DBManager) *typ
 	for addr, account := range g.Alloc {
 		if len(account.Code) != 0 {
 			originalCode := stateDB.GetCode(addr)
-			stateDB.SetCode(addr, account.Code)
+			rules := g.Config.Rules(new(big.Int).SetUint64(g.Number))
+			if rules.IsPrague {
```
ulbqb, line 314: I think this should be applied for all forks.

ulbqb: I have some question:
- Is this ToBlock used for state sync or migration?
- Can the current ToBlock restore codeinfo function even if above is true?

Mdaiki0730: > Is this ToBlock used for state sync or migration?

After some research, I found that when performing state sync or migration, the genesis state is loaded from the database when [loadLastState](https://github.com/kaiachain/kaia/blob/dev/blockchain/blockchain.go#L296) is called in blockchain.NewBlockChain.

**Let's take state_migration_test as an example.**

1. [NewBlockChain (The above process reads the genesis state and the CurrentBlock becomes the genesis)](https://github.com/kaiachain/kaia/blob/27124ed49257486f172c2a6640fd84b85ea3fc77/blockchain/state_migration_test.go#L56)
2. [Read CurrentBlock and call StartStateMigration by using it](https://github.com/kaiachain/kaia/blob/27124ed49257486f172c2a6640fd84b85ea3fc77/blockchain/state_migration_test.go#L72-L73)

In this case, even if we use the same genesis.json, the stateRoot may not match the one before v2.0.0.
Would this be an acceptable hard fork?

> Can the current ToBlock restore codeinfo even if above is true?

No, it can't.

ulbqb: I see. But this is not a problem because the state of Kaia is the fork which is defined by config when loading a genesis block. I mean, at least Kaia mainnet never load a genesis block with Prague fork.

Mdaiki0730: > because the state of Kaia is the genesis fork defined by the config when loading a genesis block

Does the root not change when we load genesis?

Mdaiki0730: In addition, there are anticipated problems with state root tests for past forks that use the traditional calculation for expected, but should we change the expected?

ulbqb: Yes, because genesis is loaded in block number zero.

Mdaiki0730: I don't understand what you mean by "genesis load", could you provide a code reference?

blukat29:
Let's merge it after rc.3.

blockchain/genesis.go
ulbqb: This will be resolved in #246.

Mdaiki0730:
I'm planning to merge dev once after #246 is merged. (To incorporate the eest update and genesis changes)

tests/block_test.go
```diff
@@ -24,28 +24,90 @@ package tests
...
-	t.Parallel()
+// TestExecutionSpecState runs the state_test fixtures from execution-spec-tests.
+type ExecutionSpecBlockTestSuite struct {
+	suite.Suite
+	originalIsPrecompiledContractAddress func(common.Address, interface{}) bool
+}
+
+func (suite *ExecutionSpecBlockTestSuite) SetupSuite() {
+	suite.originalIsPrecompiledContractAddress = common.IsPrecompiledContractAddress
+	common.IsPrecompiledContractAddress = isPrecompiledContractAddressForEthTest
```
yoomee1313, line 40: (1) in the case of Cancun, there is 0xa precompiled contract, but it seems to be ignored. Is this okay?
(2) Is it possible to replace the `PrecompiledContractsByzantium` instead of replacing IsPrecompiledContractAddress?

ulbqb: In `ActivePrecompiles`, 0x0a and 0x0b have been unconditionally added to the active precompiles so I ignored them. But this was wrong. Here needs to get correct active precompiles for tests without using `ActivePrecompiles`.

hyeonLewis:
Please rebase on dev for CI fix https://app.circleci.com/pipelines/github/kaiachain/kaia/1915/workflows/33fd30c6-9855-48bc-902b-a81c999af375/jobs/10398

tests/block_test_util.go
```diff
@@ -236,13 +356,63 @@ func (t *BlockTest) validatePostState(statedb *state.StateDB) error {
...
+			continue
+		}
+
+		storageSize := 0
+		statedb.ForEachStorage(addr, func(key, value common.Hash) bool {
+			storageSize++
+			return true
+		})
+		if storageSize != len(acct.Storage) {
+			return fmt.Errorf("account storage size mismatch for addr: %s, want: %v, have: %v", addr, len(acct.Storage), storageSize)
```
ulbqb, line 403:
```suggestion
			return fmt.Errorf("account storage size mismatch for addr: %s, want: %v, have: %v", addr.String(), len(acct.Storage), storageSize)
```

tests/block_test_util.go
```diff
@@ -236,13 +356,63 @@ func (t *BlockTest) validatePostState(statedb *state.StateDB) error {
...
+
+		// the size of HistoryStorageAddress is the same but the storage data is different
+		if addr == params.HistoryStorageAddress {
+			continue
+		}
+
+		for k, v := range acct.Storage {
+			v2 := statedb.GetState(addr, k)
+			if v2 != v {
+				return fmt.Errorf("account storage mismatch for addr: %s, slot: %x, want: %x, have: %x", addr, k, v, v2)
```
ulbqb, line 414:
```suggestion
				return fmt.Errorf("account storage mismatch for addr: %s, slot: %x, want: %x, have: %x", addr.String(), k, v, v2)
```

tests/block_test.go
Mdaiki0730: @yoomee1313 
> (1) in the case of Cancun, there is 0xa precompiled contract, but it seems to be ignored. Is this okay?

As Shogo said, we shouldn't ignore 0x0a and 0x0b. I adopted Shogo's proposal.

> (2) Is it possible to replace the PrecompiledContractsByzantium instead of replacing IsPrecompiledContractAddress?

The function needs to be replaced because the current `IsPrecompiledContractAddress` sets the entire range of `lastPrecompiledContractAddressHex` to false.
https://github.com/kaiachain/kaia/blob/906d8c2428ffa8e8b5d3191224f6836f61268005/common/types.go#L360

yoomee1313: (2) means  overwriting the existing global variable (PrecompiledAddressByzantium) in the test setup section rather than make `IsPrecompiledContract function` to be a variable.
I understood the explanation and no need to change like I mentioned.

common/types.go
```diff
@@ -353,10 +353,8 @@ func BigToAddress(b *big.Int) Address { return BytesToAddress(b.Bytes()) }
 // If s is larger than len(h), s will be cropped from the left.
 func HexToAddress(s string) Address { return BytesToAddress(FromHex(s)) }
 
-var relaxPrecompileRangeForTest bool
-
 // IsPrecompiledContractAddress returns true if the input address is in the range of precompiled contract addresses.
-func IsPrecompiledContractAddress(addr Address) bool {
+var IsPrecompiledContractAddress func(addr Address, rules interface{}) bool = func(addr Address, rules interface{}) bool {
```
ian0371, line 357: Could you add a comment saying `rules` are only for tests?

Mdaiki0730: I added it in https://github.com/kaiachain/kaia/pull/238/commits/1ab1826241fbc892b08ad80c261bf795ff424644. Thank you.

ulbqb (approved):

Mdaiki0730:
I'll make PR such like https://github.com/kaiachain/kaia/pull/250 after this PR will be merged.
Please check the points excluding it.

ian0371 (approved):

blockchain/chain_makers.go
```diff
@@ -102,6 +102,22 @@ func (b *BlockGen) AddTxWithChain(bc *BlockChain, tx *types.Transaction) {
 	b.receipts = append(b.receipts, receipt)
 }
 
+// AddTxWithChainEvenHasError is an AddTx that inherits the vmConfig of the received chain
+// and does not panic even if an error occurs.
+func (b *BlockGen) AddTxWithChainEvenHasError(bc *BlockChain, tx *types.Transaction) error {
```
yoomee1313, line 107: This function doesn't need to return error.

yoomee1313, line 107: typo?
```suggestion
func (b *BlockGen) AddTxWithChainEventHasError(bc *BlockChain, tx *types.Transaction) error {
```

PR #260, opened 2025-02-19
api: Update pending block treatment

tnasu:
## Proposed changes

Change the behavior of APIs when `"pending"` is given as an `rpc.BlockNumber` parameter.

#### 1. Remove support for pending logs
- Following single-shot APIs do not accept "pending" as a block number parameters `fromBlock` nor `toBlock`.
  - eth_getLogs, kaia_getLogs  ([src](https://github.com/kaiachain/kaia/blob/87ea0583223248874724f1f11a69118ad774300d/node/cn/filters/api.go#L379))
  - eth_newFilter, kaia_newFilter ([src](https://github.com/kaiachain/kaia/blob/87ea0583223248874724f1f11a69118ad774300d/node/cn/filters/api.go#L346))
 - Following subscription APIs do not scan pending blocks when requested with "logs" event type ([src](https://github.com/kaiachain/kaia/blob/87ea0583223248874724f1f11a69118ad774300d/node/cn/filters/api.go#L295)) ([docs](https://docs.alchemy.com/reference/eth-subscribe))
   - eth_subscribe, kaia_subscribe
- Users who have been using "pending" must now use "latest" for single-shot APIs for the same result.
- Note that for ENs and PNs the behavior won't change as they don't mine blocks, thus pending block equals latest block.

#### 2. Add support for pending blocks for feehistory and recover APIs
- Following APIs, in CN, will now utilize currently mining block when given "pending" as a block argument.
  - eth_feeHistory, kaia_feeHistory ([src](https://github.com/kaiachain/kaia/blob/87ea0583223248874724f1f11a69118ad774300d/api/api_public_klay.go#L80))
  - kaia_recoverFromTransaction ([src](https://github.com/kaiachain/kaia/blob/87ea0583223248874724f1f11a69118ad774300d/api/api_public_transaction_pool.go#L565))
- Users can keep using "pending" argument.
- In EN and PN these APIs will behave the same. Pending block equals latest block.

#### 3. Do not return error from simple query APIs
- Following APIs no longer return error upon missing data. Instead they return `null`.
  - eth_getBlockTransactionCountByNumber, kaia_getBlockTransactionCountByNumber
  - eth_getBlockTransactionCountByHash, kaia_getBlockTransactionCountByHash
  - eth_getTransactionByBlockNumberAndIndex, kaia_getTransactionByBlockNumberAndIndex
  - eth_getTransactionByBlockHashAndIndex, kaia_getTransactionByBlockHashAndIndex
  - eth_getRawTransactionByBlockNumberAndIndex, kaia_getRawTransactionByBlockNumberAndIndex
  - eth_getRawTransactionByBlockHashAndIndex, kaia_getRawTransactionByBlockHashAndIndex
- Users might need to handle the `null` result.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/accounts/abi/bind/backends/blockchain.go
+++ b/accounts/abi/bind/backends/blockchain.go
@@ -301,7 +301,7 @@ func (b *BlockchainContractBackend) FilterLogs(ctx context.Context, query kaia.F
 	if !ok {
 		return nil, errors.New("BlockChainForCaller is not blockchain.BlockChain")
 	}
-	filter := filters.NewRangeFilter(&filterBackend{state.Database().TrieDB().DiskDB(), bc}, from, to, query.Addresses, query.Topics)
+	filter := filters.NewRangeFilter(&filterBackend{state.Database().TrieDB().DiskDB(), bc, nil}, from, to, query.Addresses, query.Topics)
 
 	logs, err := filter.Logs(ctx)
 	if err != nil {
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -576,18 +576,12 @@ func (s *PublicTransactionPoolAPI) RecoverFromTransaction(ctx context.Context, e
 		return common.Address{}, err
 	}
 
-	var bn uint64
-	if blockNumber == rpc.LatestBlockNumber || blockNumber == rpc.PendingBlockNumber {
-		bn = s.b.CurrentBlock().NumberU64()
-	} else {
-		bn = blockNumber.Uint64()
-	}
-
-	signer := types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(bn))
-	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
+	state, header, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
 	if err != nil {
 		return common.Address{}, err
 	}
+	bn := header.Number.Uint64()
+	signer := types.MakeSigner(s.b.ChainConfig(), new(big.Int).SetUint64(bn))
 	_, err = tx.ValidateSender(signer, state, bn)
 	if err != nil {
 		return common.Address{}, err
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -71,18 +71,18 @@ func (s *PublicTransactionPoolAPI) GetBlockTransactionCountByHash(ctx context.Co
 }
 
 // GetTransactionByBlockNumberAndIndex returns the transaction for the given block number and index.
-func (s *PublicTransactionPoolAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) *EthRPCTransaction {
+func (s *PublicTransactionPoolAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) map[string]interface{} {
 	block, _ := s.b.BlockByNumber(ctx, blockNr)
 	if block != nil {
-		return newEthRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig())
+		return newRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig())
 	}
 	return nil
 }
 
 // GetTransactionByBlockHashAndIndex returns the transaction for the given block hash and index.
-func (s *PublicTransactionPoolAPI) GetTransactionByBlockHashAndIndex(ctx context.Context, blockHash common.Hash, index hexutil.Uint) *EthRPCTransaction {
+func (s *PublicTransactionPoolAPI) GetTransactionByBlockHashAndIndex(ctx context.Context, blockHash common.Hash, index hexutil.Uint) map[string]interface{} {
 	if block, _ := s.b.BlockByHash(ctx, blockHash); block != nil {
-		return newEthRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig())
+		return newRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig())
 	}
 	return nil
 }
--- a/api/api_public_transaction_pool.go
+++ b/api/api_public_transaction_pool.go
@@ -51,59 +51,56 @@ func NewPublicTransactionPoolAPI(b Backend, nonceLock *AddrLocker) *PublicTransa
 }
 
 // GetBlockTransactionCountByNumber returns the number of transactions in the block with the given block number.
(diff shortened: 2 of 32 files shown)


accounts/abi/bind/backends/blockchain.go
```diff
@@ -301,7 +301,7 @@ func (b *BlockchainContractBackend) FilterLogs(ctx context.Context, query kaia.F
 	if !ok {
 		return nil, errors.New("BlockChainForCaller is not blockchain.BlockChain")
 	}
-	filter := filters.NewRangeFilter(&filterBackend{state.Database().TrieDB().DiskDB(), bc}, from, to, query.Addresses, query.Topics)
+	filter := filters.NewRangeFilter(&filterBackend{state.Database().TrieDB().DiskDB(), bc, nil}, from, to, query.Addresses, query.Topics)
```
blukat29, line 304: Can `nil` backend cause panic in `filterBackend.Pending()`?

blockchain/chain_makers.go
```diff
@@ -245,6 +245,20 @@ func GenerateChain(config *params.ChainConfig, parent *types.Block, engine conse
 	return blocks, receipts
 }
 
+// GenerateChainWithGenesis is a wrapper of GenerateChain which will initialize
```
blukat29, line 248: Let's not make an arbitrary helper function just for one test. It even has a panic? and the first return value (database.Database) isn't even used.

You can simply do this in the test.
```go
db := database.NewMemoryDBManager()
genesisBlock := genesis.MustCommit(db)
blocks, receipts := GenerateChain(genesis.Config, genesisBlock, engine, db, n, gen)
```

api/api_public_transaction_pool.go
```diff
@@ -51,59 +51,56 @@ func NewPublicTransactionPoolAPI(b Backend, nonceLock *AddrLocker) *PublicTransa
...
 
 // GetTransactionByBlockNumberAndIndex returns the transaction for the given block number and index.
-func (s *PublicTransactionPoolAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) (map[string]interface{}, error) {
-	block, err := s.b.BlockByNumber(ctx, blockNr)
-	if block != nil && err == nil {
-		return newRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig()), nil
+func (s *PublicTransactionPoolAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, blockNr rpc.BlockNumber, index hexutil.Uint) *EthRPCTransaction {
+	block, _ := s.b.BlockByNumber(ctx, blockNr)
+	if block != nil {
+		return newEthRPCTransactionFromBlockIndex(block, uint64(index), s.b.ChainConfig())
```
blukat29, line 77: You can't change the return format here. The `kaia_getTransaction*` APIs and `eth_getTransaction*` APIs are different and the difference is inevitable. (1) eth_ APIs must be compatible with Ethereum spec (2) kaia_ APIs are the only way the users can access to kaia-specific fields such as Kaia tx types, account update payload, and multiple signatures.

accounts/abi/bind/backends/blockchain.go
tnasu: This is a one-time instantiation, and we won't encounter a Panic unless we use `filterBackend.Pending()` on `NewRangeFilter `. And I modified APIs not to support `pending logs` in this time.

node/cn/gasprice/feehistory.go
```diff
@@ -172,33 +172,63 @@ func (oracle *Oracle) processBlock(bf *blockFees, percentiles []float64) {
...
-	var headBlock rpc.BlockNumber
-	// query either pending block or head header and set headBlock
-	if lastBlock == rpc.PendingBlockNumber {
-		// pending block not supported by backend, process until latest block
-		lastBlock = rpc.LatestBlockNumber
-		blocks--
-		if blocks == 0 {
-			return 0, 0, nil
-		}
+func (oracle *Oracle) resolveBlockRange(ctx context.Context, reqEnd rpc.BlockNumber, blocks uint64) (*types.Block, []*types.Receipt, uint64, uint64, error) {
```
ian0371, line 175: The comment says this:
```
// Pending block does not exist in Kaia, so there is no logic to look up pending blocks.
// This part has a different implementation with Ethereum.
```

Does this PR remove the difference?

api/api_public_transaction_pool.go
tnasu: Thanks. I'll revert them.

tnasu: BTW, I found Kaia doesn't return `AccessList`, `SetCodeAuthentication`, and etc. on `kaia_getTransaction*`. So I changed the return format. 

Should we update the `newRPCTransaction` function following each transaction type of `MakeRPCOutput`?
* https://github.com/kaiachain/kaia/blob/e9c64cc83230d1d8e830ef29867fa327c3e5a19d/api/api_public_blockchain.go#L642-L660
  * https://github.com/kaiachain/kaia/blob/e9c64cc83230d1d8e830ef29867fa327c3e5a19d/blockchain/types/tx_internal_data_ethereum_access_list.go#L468
  * https://github.com/kaiachain/kaia/blob/e9c64cc83230d1d8e830ef29867fa327c3e5a19d/blockchain/types/tx_internal_data_ethereum_dynamic_fee.go#L468
  * https://github.com/kaiachain/kaia/blob/e9c64cc83230d1d8e830ef29867fa327c3e5a19d/blockchain/types/tx_internal_data_ethereum_set_code.go#L505
  * etc.

node/cn/filters/filter_system.go
```diff
@@ -120,11 +112,10 @@ type EventSystem struct {
 //
 // The returned manager has a loop that needs to be stopped with the Stop function
 // or by stopping the given mux.
-func NewEventSystem(mux *event.TypeMux, backend Backend, lightMode bool) *EventSystem {
+func NewEventSystem(mux *event.TypeMux, backend Backend) *EventSystem {
 	m := &EventSystem{
 		mux:       mux,
 		backend:   backend,
-		lightMode: lightMode,
```
ian0371, line 127: Is lightMode related to "Remove to filter pending logs"? I'd like to know the reason for this removal

node/cn/gasprice/feehistory.go
tnasu: Yes, it does. I'll remove the comment.

node/cn/filters/filter_system.go
tnasu: Thanks for your attention. I guess there is no point in searching and filtering the latest logs that `lightMode` is doing with `NewHead` after removing `pending logs`. I mean, if setting `lightMode=true`, it is the same as `lightMode=false`.
(Geth [PR](https://github.com/ethereum/go-ethereum/pull/29574) doesn't mention it.)

blukat29 (approved):

ian0371 (approved):

PR #263, opened 2025-02-21
kaiax/gasless: implement logic

ulbqb:
## Proposed changes

- Describe your changes to communicate to the maintainers why we should accept this pull request.
- If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


kaiax/gasless/impl/getter.go
```diff
@@ -185,6 +186,11 @@ func (g *GaslessModule) IsExecutable(approveTxOrNil, swapTx *types.Transaction)
 		if !ok || !g.isApproveTx(approveArgs) {
 			return false
 		}
+
+		if swapTx.Nonce() != approveTxOrNil.Nonce()+1 {
+			return false
+		}
```
blukat29, line 192: Let's move it below SP2, and call it SP3. If the KIP changes https://github.com/kaiachain/kips/pull/47/files#r1967130975.

tests/kaia_test_blockchain_test.go
```diff
@@ -176,10 +179,12 @@ func NewBCDataWithForkConfig(maxAccounts, numValidators int, chainCfg *params.Ch
 		return nil, err
 	}
 
+	mGasless := gasless_impl.NewGaslessModule()
+
```
blukat29, line 183: Nit: move it to L147.

blockchain/tx_pool.go
```diff
@@ -828,8 +836,8 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 			logger.Trace("[tx_pool] insufficient funds for cost(gas * price + value)", "from", from, "balance", senderBalance, "cost", tx.Cost())
 			return ErrInsufficientFundsFrom
 		}
-	} else {
-		// balance check for non-fee-delegated tx
+	} else if !(pool.gaslessModule.IsApproveTx(tx) || pool.gaslessModule.IsSwapTx(tx)) {
+		// balance check for non-fee-delegated tx and non-gasless tx
```
blukat29, line 840: Nit: for clarity,
```suggestion
	} else if pool.gaslessModule.IsApproveTx(tx) || pool.gaslessModule.IsSwapTx(tx) {
		logger.Trace("[tx_pool] accept gasless transaction", "from", from, "txhash", tx.Hash().Hex())
	} else {
		// balance check for non-fee-delegated tx and non-gasless tx
```

blockchain/tx_pool.go
```diff
@@ -1439,7 +1447,9 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) {
 			pool.priced.Removed()
 		}
 		// Drop all transactions that are too costly (low balance)
-		drops, _ := list.Filter(addr, pool)
+		drops, _ := list.Filter(addr, pool, func(tx *types.Transaction) (bool, bool) {
```
ian0371, line 1450: Let's introduce IsModuleTx(tx) and ValidateTx(tx) method in TxPoolModule.

```
drops, _ := list.Filter(addr, pool, modules)

func (l *txList) Filter(sender common.Address, pool *TxPool, modules []TxPoolModule) {
	...
	for each module of modules {
		if module.IsModuleTx(tx) {
			return module.ValidateTx(tx)
		}
	}
}
```

blockchain/tx_list.go
```diff
@@ -216,6 +216,11 @@ func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Tr
 		if m.items[next].GasPrice().Cmp(baseFee) < 0 {
 			break
 		}
+		if mReady != nil {
+			if isMoudleTx, ok := mReady(m.items, next); isMoudleTx && !ok {
```
ian0371, line 220: Let's introduce `IsReady(txs)` method in TxPoolModule.

```
func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int, modules []TxPoolModule) types.Transactions {
	for each module of modules {
		if module.IsModuleTx(txs) {
			check module.IsReady(txs)
		}
	}
}
```

ian0371: Or we can change like this from the call site:
```
		var readyTxs types.Transactions
		if pool.rules.IsMagma {
			readyTxs = list.ReadyWithGasPrice(pool.getPendingNonce(addr), pool.gasPrice)
		} else {
			readyTxs = list.Ready(pool.getPendingNonce(addr))
		}

		for each module of modules {
			readyTxs = append(readyTxs, module.ReadyTxs(txs))
		}
```

blockchain/tx_pool.go
```diff
@@ -828,8 +836,8 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 			logger.Trace("[tx_pool] insufficient funds for cost(gas * price + value)", "from", from, "balance", senderBalance, "cost", tx.Cost())
 			return ErrInsufficientFundsFrom
 		}
-	} else {
-		// balance check for non-fee-delegated tx
+	} else if !(pool.gaslessModule.IsApproveTx(tx) || pool.gaslessModule.IsSwapTx(tx)) {
```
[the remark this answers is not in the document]
ulbqb: This means that module.ValidateTx will contain a lot of duplicate code with TxPool.validateTx, which seems like it could be a breeding ground for bugs. What do you think?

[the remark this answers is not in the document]
ian0371: You're right, nvm.

kaiax/gasless/impl/tx_pool.go
```diff
@@ -0,0 +1,68 @@
...
+	return nil
+}
+
+func (g *GaslessModule) IsModuleTx(_ kaiax.TxPool, tx *types.Transaction) bool {
+	return g.IsApproveTx(tx) || g.IsSwapTx(tx)
+}
+
+func (g *GaslessModule) GetCheckBalance() func(pool kaiax.TxPool, tx *types.Transaction) error {
+	return func(kaiax.TxPool, *types.Transaction) error { return nil }
+}
```
ian0371, line 40: why return a callback instead of checking balance here?
If needed, we can add statedb to GaslessModule, which is also required for nonce check (SP3).

kaiax/gasless/impl/getter.go
```diff
@@ -197,9 +197,13 @@ func (g *GaslessModule) IsExecutable(approveTxOrNil, swapTx *types.Transaction)
 		if approveArgs.Amount.Cmp(swapArgs.AmountIn) < 0 {
 			return false
 		}
+		// SP3.
+		if approveTxOrNil.Nonce()+1 != swapTx.Nonce() {
+			return false
+		}
```
ian0371, line 203: We should also check from statedb. Please add it to GaslessModule if needed.

blockchain/tx_pool.go
```diff
@@ -828,6 +849,8 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 			logger.Trace("[tx_pool] insufficient funds for cost(gas * price + value)", "from", from, "balance", senderBalance, "cost", tx.Cost())
 			return ErrInsufficientFundsFrom
 		}
+	} else if checkedBalance {
+		// skip balance check
 	} else {
```
ian0371, line 854:
```suggestion
	} else if !moduleCheckedBalance {
```

blockchain/tx_pool.go
```diff
@@ -779,9 +783,26 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
...
+	// balance check for module transaction if module have check balance function
+	checkedBalance := false
+	for _, module := range pool.modules {
+		if module.IsModuleTx(pool, tx) {
+			if checkBalance := module.GetCheckBalance(); checkBalance != nil {
+				checkedBalance = true
+				err := checkBalance(pool, tx)
+				if err != nil {
+					logger.Trace("[tx_pool] invalid funds of module transaction sender", "from", from, "txhash", tx.Hash().Hex())
+					return err
```
ian0371, line 795: If a module returns checkBalance function, it will override the default balance check (by setting `checkedBalance=true`).
What if a module doesn't want to ignore the default balance check?

blockchain/tx_list.go
```diff
@@ -380,6 +392,16 @@ func (l *txList) Filter(sender common.Address, pool *TxPool) (types.Transactions
 				return senderBalance.Cmp(tx.Value()) < 0 || feePayerBalance.Cmp(tx.Fee()) < 0
 			}
 		}
+
+		// balance check for module transaction
+		for _, module := range pool.modules {
+			if module.IsModuleTx(pool, tx) {
+				if checkBalance := module.GetCheckBalance(); checkBalance != nil {
+					return checkBalance(pool, tx) != nil
```
ian0371, line 400: Only the first module's GetCheckBalance is executed, unlike in `validateTx` where the function is executed for all modules.

kaiax/gasless/impl/tx_pool.go
ulbqb: If the balance check is overwritten simply, all modules need to implement this even if they want to use the existing logic. So I provided the option to overwrite logic.

blockchain/tx_list.go
ulbqb: https://github.com/ulbqb/kaia/blob/949f09bb4032b2bbf179a65aa2dba02cb3f26ede/blockchain/tx_pool.go#L797

`validateTx` is the same as well.

ian0371: I missed it, thanks

blockchain/tx_pool.go
ulbqb: You mean the situation where a module want additional check in addition to the default logic?

blockchain/tx_list.go
```diff
@@ -216,6 +216,18 @@ func (m *txSortedMap) ReadyWithGasPrice(start uint64, baseFee *big.Int) types.Tr
...
+
+		rejectByModule := false
+		for _, module := range pool.modules {
+			if module.IsModuleTx(pool, m.items[next]) && !module.IsReady(pool, m.items, next, ready) {
+				rejectByModule = true
+				break
+			}
+		}
+		if rejectByModule {
+			break
```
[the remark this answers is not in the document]
ian0371: No, I assumed you used break here because there's break above this line.
Any reason you used break instead of continue? IMO we should use continue.

blockchain/tx_pool.go
ian0371: Right. For now, we can leave it as-is, but let's consider it later

blockchain/tx_list.go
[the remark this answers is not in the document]
ulbqb: https://github.com/kaiachain/kaia/blob/ab4a78d98caa22d2600fb16dbd05bcb94739e4a7/blockchain/tx_list.go#L200
This function has to return sequential txs in nonce. Later txs don't need to be checked because of invalid nonce when a transaction is rejected.

blockchain/tx_pool.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/32ba61a19c1e5ba6af2b9f48553cd4e8d426f212

kaiax/gasless/impl/getter.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/b352df52dc3b604305f212343f1f79b47773aea7

blockchain/tx_list.go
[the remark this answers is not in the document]
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/a28a08c8c14edc92a51c66c400d0f306d432396c

tests/kaia_test_blockchain_test.go
```diff
@@ -71,6 +73,7 @@ type BCData struct {
 	engine             consensus.Istanbul
 	genesis            *blockchain.Genesis
 	govModule          gov.GovModule
+	modules            []kaiax.TxPoolModule
```
ian0371, line 76:
```suggestion
	txPoolModules            []kaiax.TxPoolModule
```

kaiax/interface.go
```diff
@@ -142,6 +142,17 @@ type TxPoolModule interface {
 	// Additional actions to be taken when a new tx arrives at txpool
 	PreAddLocal(*types.Transaction) error
 	PreAddRemote(*types.Transaction) error
+
+	// Additional checks to check if a given transaction should be handled by module.
+	IsModuleTx(tx *types.Transaction) bool
+
+	// Optional actions to check if sender balance is valid for module transaction.
+	// This is mainly used on checking if module transaction be appended to queue.
+	// If nil is returned, existing check perform, otherwise returned fuction perform instead of it.
```
ian0371, line 151:
```suggestion
	// This is mainly used on checking if module transaction should be appended to queue.
	// If nil is returned, default check (balance > txFee) is performed. Otherwise, the returned function overrides default check.
```

kaiax/gasless/impl/tx_pool.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	isApproveTx := g.IsApproveTx(tx)
+	isSwapTx := g.IsSwapTx(tx)
+
+	if isApproveTx && i == nonce && i+1 < uint64(math.MaxUint64) {
+		if next := txs[i+1]; next != nil && g.IsSwapTx(next) {
+			return g.IsExecutable(tx, next)
+		}
+	}
```
ian0371, line 60: How about introducing a new function for this branch?
```
// IsApproveTxReady assumes that the caller checked `g.IsApproveTx(approveTx)`
func (g *GaslessModule) IsApproveTxReady(approveTx, swapTx *types.Transaction) bool {
	nonce := approveTx.nonce
	if nonce != statedb.nonce {
		return false
	}
	if nonce == uint64(math.MaxUint64) {
		return false
	}
	if swapTx == nil || g.IsSwapTx(swapTx) {
		return false
	}
	
	return g.IsExecutable(approveTx, swapTx)
}
```

ian0371 (commented):
GaslessModule interface should have TxPoolModule:

```
type GaslessModule interface {
	kaiax.BaseModule
	kaiax.TxPoolModule
	builder.TxBundlingModule
}
```

Also, can we remove `IsExecutable` from GaslessModule interface? It's because module user will not call `IsExecutable`.

accounts/abi/bind/backends/blockchain_test.go
```diff
@@ -281,7 +282,7 @@ func TestBlockChainSendTransaction(t *testing.T) {
 	state, err := bc.State()
 	txPoolConfig := blockchain.DefaultTxPoolConfig
 	txPoolConfig.Journal = "/dev/null" // disable journaling to file
-	txPool := blockchain.NewTxPool(txPoolConfig, bc.Config(), bc, &dummyGovModule{chainConfig: bc.Config()})
+	txPool := blockchain.NewTxPool(txPoolConfig, bc.Config(), bc, &dummyGovModule{chainConfig: bc.Config()}, []kaiax.TxPoolModule{})
```
blukat29, line 285: nit: `nil` is more concise.
```suggestion
	txPool := blockchain.NewTxPool(txPoolConfig, bc.Config(), bc, &dummyGovModule{chainConfig: bc.Config()}, nil)
```

blockchain/tx_pool.go
```diff
@@ -779,6 +783,22 @@ func (pool *TxPool) validateTx(tx *types.Transaction) error {
 		return ErrNonceTooLow
 	}
 
+	// balance check for module transaction if module have check balance function
+	moduleCheckedBalance := false
```
blukat29, line 787:
```suggestion
	// If module recognizes the tx, run an alternative balance check and then skip the default balance check later.
	shouldSkipBalanceCheck := false
```

blockchain/tx_pool.go
blukat29: Yes. Let's consider things like this _later_.
```
module.ShouldOverrideBalanceCheck(tx) -> bool
module.BalanceCheck(tx)
```

node/cn/backend.go
```diff
@@ -344,7 +345,19 @@ func New(ctx *node.ServiceContext, config *Config) (*CN, error) {
 	}
 	// TODO-Kaia-ServiceChain: add account creation prevention in the txPool if TxTypeAccountCreation is supported.
 	config.TxPool.NoAccountCreation = config.NoAccountCreation
-	cn.txPool = blockchain.NewTxPool(config.TxPool, cn.chainConfig, bc, mGov)
+
+	mGasless := gasless_impl.NewGaslessModule()
+	statedb, err := bc.State()
```
blukat29, line 350: You can't pass the statedb to gasless module and use it forever. `bc.State()` returns the latest state "at the point of node startup". And StateDB represents a frozen state at a certain block then never changes. In the module you must either
1. receive statedb argument to getter functions
2. call `g.Chain.State`every time to retrieve the latest state
3. Don't use statedb. Give up the checks that depend on nonce. Let TxPool deal with the nonce. (perhaps promoteTx is already checking tx.nonce == getNonce(tx.from)?)

ian0371: fyi, nonce check was to prevent multiple gasless txs from same sender in a block, thus allowing only one gasless tx pair at the getNonce(user).

tests/kaia_test_blockchain_test.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/7a9c83ca426b9647112113a517fb7c50effc9005

kaiax/interface.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/c135c8f4a3596e4dd4e79739be49f31afd1c8e44

kaiax/gasless/impl/tx_pool.go
ulbqb: Do you assume this is used in tx pool?

accounts/abi/bind/backends/blockchain_test.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/0f6f9d4519d3e5bc0b5f0906f3eae01324bda366

blockchain/tx_pool.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/17fbbe4e7f28e89df17de1ab5c19856fa90a7991

ulbqb:
I added rpc api `kaia_sendRawTransactions` at https://github.com/kaiachain/kaia/pull/263/commits/e426131b60ddafb0ef5b371cba84b0eb9680eebd.

kaiax/gasless/impl/tx_pool.go
ian0371: No, only from `IsReady`

ulbqb:
I added rpc api kaia_sendRawtransactions in builder module.

node/cn/backend.go
blukat29: Oh, you're right @ian. Number 3 is not an option, because we specifically want to allow only one gasless tx per sender. promoteTx would allow multiple consecutive txs.

Between 1 and 2, I'd prefer 1 because you can simply pass `pool.currentState` as the argument.
- `chain.State()` is quite costly as it allocates multiple memory objects in `state.New`. Considering that the promotion check is performed for every transaction, the cost could be non-negligible.
- This makes the gasless module simpler because you can forget about states when testing/using any other getter functions.
- The same trick has been used by `engine.Finalize()`, `StateProcessor.Process()`.

kaiax/gasless/impl/tx_pool.go
ulbqb: fix at https://github.com/kaiachain/kaia/pull/263/commits/a15e7868058165b585799dc051d5086cdf6a40b5

node/cn/backend.go
ulbqb: Or can we use reset hook as TxPoolModule on executing TxPool.reset? Statedb of gasless module can be updated every `Reset` execution. This function is also used for gov module.

https://github.com/kaiachain/kaia/blob/ab4a78d98caa22d2600fb16dbd05bcb94739e4a7/blockchain/tx_pool.go#L504-L507

blukat29: While you can call `TxPoolModule.Reset()` in `txpool.reset()`, you can also do the same in `ExecutionModule.PostInsertBlock()`. In that case, you may maintain the gasless module's internal variable `g.currentState`.

Note that reset and PostInsertBlock are called in very similar timings.
- insertChain -> PostInsertBlock
- insertChain -> PostChainEvents -> pool.chainHeadCh -> pool.reset

ulbqb: I see. In this time, I select Reset of txpool because the main logic of gasless module is in the tx pool.

ian0371 (approved):

kaiax/builder/impl/api.go
```diff
@@ -39,4 +47,37 @@ func NewBuilderAPI(b *BuilderModule) *BuilderAPI {
 	return &BuilderAPI{b}
 }
 
-// TODO: implement SendRawTransactions
+func (s *BuilderAPI) SendRawTransactions(ctx context.Context, inputs []hexutil.Bytes) ([]common.Hash, error) {
```
hyeonLewis, line 50: It's a question. As far as I understand, if a user sends the [approvalTx, swapTx], then the first attempt to promote `approvalTx` will fail (`IsReady == false`) since the tx pool doesn't know about the swapTx but will be retained in the queue and promoted with `swapTx` insertion. Is it right?

ulbqb: Yes, right.

hyeonLewis (approved):

PR #272, opened 2025-03-05
kaiax/gasless: Implement worker for tx bundle

Mdaiki0730:
## Proposed changes

Merging this PR will enable worker bundling in [kip-245](https://github.com/kaiachain/kips/blob/main/KIPs/kip-245.md#bundle-generation-logic).
**New features about kip-245**
- Bundle Generation Logic
- Bundle Execution (commitBundleTransaction)
- Bundle Rollback
- Delaying Timeout

Also, some changes were made to implement above.
- The transactions that the worker iterates when ApplyTransactions has been changed from txs(TransactionsByPriceAndNonce) to incorporatedTxs([]interface{}).
  - For this we have added new methods to task, which mirror Pop and Shift from TransactionsByPriceAndNonce.
- Define the methods in the builderModule as public functions.
  - These do not use any struct fields so functions are sufficient.
- Now that we have the `RegisterTxBundlingModule`, a node can use it to register the `TxBundlingModule`.

**About Pop and Shift**
We create a dependency graph during pop for the following reasons.

Shift just advances to the next tx, while pop deletes all tx that have the same sender as the failed tx.
This deletion may affect other bundles in the following cases:
```
txs =     [ G1 A.T1 B.T1 A.T2 B.T2 C.T1 ]
bundle =  [ G1 A.T1 ]
bundle2 = [ B.T1 A.T2 ]
bundle3 = [ B.T2 C.T1 ]
on A.T1 revert, pop(txs, 2) => txs = [ ]

[ G1 A.T1 B.T1 A.T2 B.T2 C.T1 ]
pop 2 elements
[ B.T1 A.T2 B.T2 C.T1 ]
A.T2 is broken (nonce) so remove it
[ B.T1 B.T2 C.T1 ]
bundle2 is broken so remove B.T1
[ B.T2 C.T1 ]
B.T2 is broken (nonce) so remove it
[ C.T1 ]
bundle3 is broken so remove C.T1
[ ]
```

**Create TestScenario**
1. Transfer (reservoir1 -> anon) using a legacy transaction four times. 
Create a mock TxBundlingModule and create a bundle for each two tx. 
At that point, tx4 fails with nonceTooHigh and tx3 is reverted.
2. Transfer (reservoir2 -> anon) using a legacy transaction two times.

There were also some test util changes.
- Add the ability to assign `TxBundlingModules` and `BuilderModules` to `MineABlock`.
- Add txHashesExpectedFail to accountMap Update
  - To support tests that assume tx failure

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


work/worker.go
```diff
@@ -750,25 +779,43 @@ CommitTransactionLoop:
...
+		targetBundle := &builder.Bundle{}
+		numShift := 1
+		if bundleIdx := builder_impl.FindBundleIdx(bundles, tx); bundleIdx != -1 {
+			targetBundle = bundles[bundleIdx]
+			numShift = len(targetBundle.BundleTxs)
+		}
+
+		if len(targetBundle.BundleTxs) != 0 {
+			atomic.StoreInt32(&isExecutingBundleTxs, 1)
+			err, tx, logs = env.commitBundleTransaction(targetBundle, bc, rewardbase, vmConfig)
```
ian0371, line 792: We must ensure that `commitBundleTransaction` always returns non-nil `tx` because it can be accessed.
I think always returning last executed tx (regardless of error or not) can be ok.

tests/kaia_scenario_test.go
```diff
@@ -1891,6 +1895,239 @@ func TestValidateSender(t *testing.T) {
 	}
 }
 
+// TestTxBundleAndRevert tests a following scenario:
+//  1. Transfer (reservoir1 -> anon) using a legacy transaction four times.
+//     Create a mock TxBundlingModule and create a bundle for each two tx.
+//     At that point, tx4 fails with nonceTooHigh and tx3 is reverted.
+//  2. Transfer (reservoir2 -> anon) using a legacy transaction two times.
+//
+// Therefore, the expected txs for a block are [tx0, tx1, tx4, tx5].
```
ian0371, line 1904:
```suggestion
// In summary, input txs are [ [tx0, tx1], [tx2, tx3], [tx4, tx5] ]. tx3 reverts.
// Therefore, the expected txs for a block are [tx0, tx1, tx4, tx5].
```

work/worker.go
Mdaiki0730: Yes, I agree with you.
However, the first tx is a generator, and if gen fails there is no tx to return.
https://github.com/kaiachain/kaia/pull/272/files#diff-aa53dfe64c81ad7f4673af6857ca8ae7873bb18224202adb09f765ce4f458411R912

What about returning a Transaction pointer instead of nil? I mean `&types.Transaction{}`.

ian0371: Sounds great

Mdaiki0730: I've fixed it at f7474a6. Thank you!

kaiax/builder/impl/getter.go
```diff
@@ -17,14 +17,108 @@
...
+	"reflect"
+
 	"github.com/kaiachain/kaia/blockchain/types"
 	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/kaiax/builder"
 )
 
+type Pair struct {
+	u, v int
+}
```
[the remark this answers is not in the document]
Mdaiki0730: I've fixed it at f7474a6. Thank you!

tests/kaia_scenario_test.go
```diff
@@ -1891,6 +1895,240 @@ func TestValidateSender(t *testing.T) {
...
+
+			err := tx.SignWithKeys(signer, reservoir1.Keys)
+			assert.Equal(t, nil, err)
+
+			txs = append(txs, tx)
+		}
+
+		// 2. Transfer (reservoir2 -> anon) using a legacy transaction two times.
+		for i := 0; i < 2; i++ {
+			fmt.Println(reservoir2.Nonce)
```
ulbqb, line 2010:
```suggestion
```

kaiax/builder/bundle.go
```diff
@@ -30,22 +30,21 @@ type Bundle struct {
...
+	for i, txOrGen := range b.BundleTxs {
 		switch v := txOrGen.(type) {
 		case *types.Transaction:
 			if v.Hash() == hash {
-				return true
-			}
-		case TxGenerator:
-			if tx, err := v(nonce); err != nil && tx.Hash() == hash {
-				return true
+				return i
```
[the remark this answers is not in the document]
Mdaiki0730: That's good. I've fiexed it at cfa418d.

Mdaiki0730:
I rebased feat/gasless to include #263. I force pushed and there are no conflicts.
According to 768fcea, we have changed it so that TxBundlingModule is registered to Miner during cn init.

ian0371 (approved):

kaiax/builder/impl/getter.go
```diff
@@ -17,14 +17,104 @@
...
 	"github.com/kaiachain/kaia/blockchain/types"
 	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/kaiax/builder"
 )
 
+// BuildDependencyGraph builds a dependency graph of txs.
+// Two txs with the same sender has an edge.
+// Two txs in the same bundle has an edge.
+// TxGenerator is not connected to any other txs.
+func BuildDependencyGraph(txs []interface{}, bundles []*builder.Bundle, signer types.Signer) ([][]int, error) {
```
hyeonLewis, line 31: fyi. I benchmarked the disjoin set (union-find) if it shows better performance but it seems BFS is much better.

<details><summary>Disjoin impl</summary>

```
type DisjointSet struct {
    parent []int
    rank   []int
}

func NewDisjointSet(size int) *DisjointSet {
    parent := make([]int, size)
    rank := make([]int, size)
    for i := range parent {
        parent[i] = i
    }
    return &DisjointSet{parent: parent, rank: rank}
}

func (ds *DisjointSet) Find(x int) int {
    if ds.parent[x] != x {
        ds.parent[x] = ds.Find(ds.parent[x])
    }
    return ds.parent[x]
}

func (ds *DisjointSet) Union(x, y int) {
    px, py := ds.Find(x), ds.Find(y)
    if px == py {
        return
    }
    if ds.rank[px] < ds.rank[py] {
        ds.parent[px] = py
    } else if ds.rank[px] > ds.rank[py] {
        ds.parent[py] = px
    } else {
        ds.parent[py] = px
        ds.rank[px]++
    }
}

func BuildDependencyGraph2(txs []interface{}, bundles []*builder.Bundle, signer types.Signer) (*DisjointSet, error) {
    ds := NewDisjointSet(len(txs))
    
    // 1. Group txs by sender
    senderToIndices := make(map[common.Address][]int)
    for i, txOrGen := range txs {
        if tx, ok := txOrGen.(*types.Transaction); ok {
            from, err := types.Sender(signer, tx)
            if err != nil {
                return nil, err
            }
            senderToIndices[from] = append(senderToIndices[from], i)
        }
    }

    // 2. Union same sender transactions
    for _, indices := range senderToIndices {
        for i := 1; i < len(indices); i++ {
            ds.Union(indices[i-1], indices[i])
        }
    }

    // 3. Union bundle transactions
    for _, bundle := range bundles {
        indices := make([]int, 0)
        for i, txOrGen := range txs {
            if tx, ok := txOrGen.(*types.Transaction); ok {
                if bundle.Has(tx.Hash()) {
                    indices = append(indices, i)
                }
            }
        }
        for i := 1; i < len(indices); i++ {
            ds.Union(indices[i-1], indices[i])
        }
    }

    return ds, nil
}

func FindDependentNodes2(ds *DisjointSet, starts []int) map[int]bool {
    result := make(map[int]bool)
    
    for _, start := range starts {
        startRoot := ds.Find(start)
        for i := 0; i < len(ds.parent); i++ {
            if ds.Find(i) == startRoot {
                result[i] = true
            }
        }
    }

    return result
}
```

```
// benchmark 
BenchmarkPopTxs/PopTxs-11                1000000                54.84 ns/op
BenchmarkPopTxs/PopTxs2-11               1000000                67.72 ns/op
```
</details>

ian0371:
In terms of timeout test, temporarily setting timeout to 150us depends on PC spec, so timeout can occur (1) before executing any txs, (2) during tx execution, (3) after all tx execution. In case of (1), the test can fail. In case of (3), the test will be successful but timeout wouldn't have occurred. Later let's improve accuracy of timeout test so timeout almost always occurs at (2).
```
     (1)                   (2)                    (3)
|------------|-------------------------------|-------------|
             V                               V
    bundle execution start         bundle execution end
```

work/worker.go
```diff
@@ -722,20 +733,38 @@ func (env *Task) ApplyTransactions(txs *types.TransactionsByPriceAndNonce, bc Bl
...
+		switch txOrGen.(type) {
+		case *types.Transaction:
+			tx = txOrGen.(*types.Transaction)
+		case builder.TxGenerator:
+			txGen := txOrGen.(builder.TxGenerator)
+			nodeNonce := env.state.GetNonce(rewardbase)
+			tx, err = txGen(nodeNonce)
+			if tx == nil {
+				logger.Warn("TxGenerator returned a nil tx", "error", err)
+			}
```
hyeonLewis, line 760:
```suggestion
			tx, err = txGen(nodeNonce)
			if tx == nil {
				logger.Warn("TxGenerator returned a nil tx", "error", err)
				continue
			}
```

work/worker.go
```diff
@@ -750,25 +779,43 @@ CommitTransactionLoop:
 		// Start executing the transaction
 		env.state.SetTxContext(tx.Hash(), common.Hash{}, env.tcount)
 
-		err, logs := env.commitTransaction(tx, bc, rewardbase, vmConfig)
+		// Verify that tx is included in the bundle
+		targetBundle := &builder.Bundle{}
+		numShift := 1
+		if bundleIdx := builder_impl.FindBundleIdx(bundles, tx); bundleIdx != -1 {
```
hyeonLewis, line 785: I think it can't detect the first lend tx since lend tx is just created by `txGen` so isn't in the `bundle.BundleTxs`.

work/worker.go
```diff
@@ -829,6 +887,60 @@ func (env *Task) commitTransaction(tx *types.Transaction, bc BlockChain, rewardb
...
+			if err != vm.ErrInsufficientBalance && err != vm.ErrTotalTimeLimitReached {
+				for _, txInBundle := range bundle.BundleTxs {
+					switch v := txInBundle.(type) {
+					case *types.Transaction:
+						v.MarkUnexecutable(true)
+					}
+				}
+			}
+			*env.state = *lastSnapshot
+			env.header.GasUsed = gasUsedSnapshot
```
hyeonLewis, line 929: AFAIK we need to roll back the state if the tx is *reverted*.

The error returned by `ApplyTransaction` isn't a VM error, but a validation or time limit error. We also need to check the `receipt.status` if it has any vmerr.

work/worker.go
```diff
@@ -829,6 +887,60 @@ func (env *Task) commitTransaction(tx *types.Transaction, bc BlockChain, rewardb
...
+	gasUsedSnapshot := env.header.GasUsed
+	txs := []*types.Transaction{}
+	receipts := []*types.Receipt{}
+	logs := []*types.Log{}
+
+	for _, txOrGen := range bundle.BundleTxs {
+		var tx *types.Transaction
+		if gen, ok := txOrGen.(builder.TxGenerator); ok {
+			var err error
+			tx, err = gen(env.state.GetNonce(rewardbase) + 1)
```
ulbqb, line 901: Why is `env.state.GetNonce(rewardbase) + 1` used while `env.state.GetNonce(rewardbase)` is used in L756?

kaiax/builder/impl/getter.go
```diff
@@ -100,3 +190,121 @@ func (b *BuilderModule) IsConflict(prevBundles []*builder.Bundle, newBundles []*
...
+			var itx interface{} = tx
+			flattenedTxs = append(flattenedTxs, itx)
+		}
+		return flattenedTxs, nil
+	}
+
+	for _, txBundlingModule := range txBundlingModules {
+		newBundles := txBundlingModule.ExtractTxBundles(arrayTxs, bundles)
+		if IsConflict(bundles, newBundles) {
+			logger.Warn("Gas limit exceeded for current block", "", txBundlingModule)
```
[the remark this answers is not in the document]
Mdaiki0730: Yes, I'll delete this.

kaiax/builder/impl/getter.go
Mdaiki0730: Thanks for your confirmation, I'm glad to hear it.

kaiax/builder/impl/getter.go
```diff
@@ -100,3 +190,121 @@ func (b *BuilderModule) IsConflict(prevBundles []*builder.Bundle, newBundles []*
...
+		for _, tx := range arrayTxs {
+			var itx interface{} = tx
+			flattenedTxs = append(flattenedTxs, itx)
+		}
+		return flattenedTxs, nil
+	}
+
+	for _, txBundlingModule := range txBundlingModules {
+		newBundles := txBundlingModule.ExtractTxBundles(arrayTxs, bundles)
+		if IsConflict(bundles, newBundles) {
```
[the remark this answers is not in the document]
Mdaiki0730: This is indeed the case currently, and the expectation is that only conflicting bundles will be excluded.
It seems better to use FindConflict instead of IsConflict and exclude it from bundles.

work/worker.go
```diff
@@ -722,20 +733,39 @@ func (env *Task) ApplyTransactions(txs *types.TransactionsByPriceAndNonce, bc Bl
...
+		switch txOrGen.(type) {
+		case *types.Transaction:
+			tx = txOrGen.(*types.Transaction)
+		case builder.TxGenerator:
+			txGen := txOrGen.(builder.TxGenerator)
+			nodeNonce := env.state.GetNonce(rewardbase)
+			tx, err = txGen(nodeNonce)
+			if tx == nil {
+				logger.Warn("TxGenerator returned a nil tx", "error", err)
+				continue
```
ian0371, line 760: We should pop `txGen`, otherwise it can fall into an infinite loop

work/worker.go
Mdaiki0730: Oh, that's true.
Has, which is used when creating BFS, is not expected to include TxGenerator, so it seems necessary to separate this.

work/worker.go
Mdaiki0730: I see. However, since the target bundle cannot be identified, is it difficult to pop it here?
We need a method to compare the interface with the bundle.

work/worker.go
Mdaiki0730: It seems that L756 is wrong. `env.state.GetNonce(rewardbase) + 1` should be used. Thanks.

work/worker.go
Mdaiki0730: If it is vmerr does that mean there is no need to roll back?

work/worker.go
Mdaiki0730: It was fixed at 600f7e8.

work/worker.go
hyeonLewis: No, I think vmerr should lead to a rollback. 

In [LendTx, ApprovalTx, SwapTx], if `SwapTx` is reverted but not rollback whole bundle, the proposer can't receive KAIA spent in `LendTx`. I thought it's the main purpose of bundling. cc. @ian0371 @blukat29

work/worker.go
Mdaiki0730: This allows us to put this processing before tx creation.
https://github.com/kaiachain/kaia/pull/272/files#diff-aa53dfe64c81ad7f4673af6857ca8ae7873bb18224202adb09f765ce4f458411R782-R788

Mdaiki0730: It was fixed at c866a31. Thank you.

work/worker.go
Mdaiki0730: It was fixed at https://github.com/kaiachain/kaia/commit/c866a31888141d2d51e538da6e2285b29d8f8150. Thank you.

work/worker.go
ian0371: This is an extension of `commitTransaction`, in which state is rolled back (via `RevertToSnapshot`) when EVM reverts.
Likewise, I think this can handle EVM error.

ian0371: https://github.com/kaiachain/kaia/blob/cf0ebf012cf7d60bc67072836dd2d9c0cd05f59a/blockchain/state_transition.go#L461
I think Lewis also has a valid point.
Isn't the state rollback performed in `commitTransaction`'s `RevertToSnapshot` upon EVM revert?

Mdaiki0730: I also agree with the point Lewis mentioned.

> Isn't the state rollback performed in commitTransaction's RevertToSnapshot upon EVM revert?

Yes, in the case of vmerr, commitTransaction's revert is not executed.

ian0371: State rollback: `lastSnapshot` is a copy of statedb before bundle execution, so any state changes are discarded.
Txs discard: txs in bundle is added to `env.txs` in the outer loop, which means only when all txs succeeded.

Mdaiki0730: @ian0371 
As you said, there is a difference in the scope of rollback when vmerr fails as a single tx and when vmerr fails as a bundle.
Can we tolerate this difference?

work/worker.go
ian0371: I don't think DeepEqual is ok. https://go.dev/play/p/B9E5K9I6p-u

work/worker.go
Mdaiki0730: I'm in favor of tolerating this and rolling back anyway in the case of vmerr.

work/worker.go
ian0371: Because we refresh `bundles` every loop, I think it's safe to assume:
- TxGenerator must be in `bundles[0]`
- If tx is in bundle, it must be in `bundles[0]`. So it's sufficient to check `bundles[0].Has(tx)`

work/worker.go
Mdaiki0730: In the case of commitBundleTransaction, we have changed it so that a revert occurs if receipt.Status is not Successful. PTAL 582fd7c

work/worker.go
ian0371: On, nvm. For a sec I thought `bundles` was sorted.
Now that we update bundles' `TargetTxHash`, maybe we can find and use the bundle whose `TargetTxHash` is zero.

Mdaiki0730: That's good idea! I'll try it.

kaiax/builder/impl/getter.go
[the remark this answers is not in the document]
Mdaiki0730: It was fixed at 1f2f17e. PTAL

work/worker.go
Mdaiki0730: I added a hash to TxGenerator to make it comparable. PTAL 6618e63.
There is room for refactoring, such as making TxOrGen(interface{}) a type, but this will be done in a separate PR.

kaiax/builder/impl/getter.go
```diff
@@ -100,3 +200,162 @@ func (b *BuilderModule) IsConflict(prevBundles []*builder.Bundle, newBundles []*
...
+func SetCorrectTargetTxHash(bundles []*builder.Bundle, txs []interface{}) []*builder.Bundle {
+	ret := make([]*builder.Bundle, 0)
+	for _, bundle := range bundles {
+		bundle.TargetTxHash = FindTargetTxHash(bundle, txs)
+		ret = append(ret, bundle)
+	}
+	return ret
+}
+
+func FindTargetTxHash(bundle *builder.Bundle, txs []interface{}) common.Hash {
```
ian0371, line 267: To look similar as `Find*` functions above:

```suggestion 
func FindTxIdx(txs []interface{}, tx *types.Transaction) int {
	...
}
```

Usage:
```
for _, bundle := range bundles {
	newTargetTx = txs[FindTxIdx(txs, bundle.BundleTxs[0]) - 1]
	bundle.TargetTxHash = newTargetTx.Hash()
	ret = append(ret, bundle)
}
```

kaiax/builder/impl/getter.go
```diff
@@ -41,7 +132,7 @@ func (b *BuilderModule) IncorporateBundleTx(txs []*types.Transaction, bundles []
 }
 
 // incorporate assumes that `txs` does not contain any bundle transactions.
-func incorporate(txs []interface{}, bundle *builder.Bundle) ([]interface{}, error) {
+func incorporate(txs []interface{}, bundle *builder.Bundle, bundleIdx int) ([]interface{}, error) {
```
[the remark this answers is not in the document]
Mdaiki0730: It was removed in f0bbffc. Thank you!

kaiax/builder/impl/getter.go
Mdaiki0730: This usage may be possible in the future, but newTargetTx is an interface{} and does not have Hash().

kaiax/builder/impl/getter.go
```diff
@@ -17,14 +17,105 @@
...
+	"github.com/kaiachain/kaia/crypto"
 	"github.com/kaiachain/kaia/kaiax/builder"
 )
 
+// BuildDependencyGraph builds a dependency graph of txs.
+// Two txs with the same sender has an edge.
+// Two txs in the same bundle has an edge.
+// TxGenerator is not connected to any other txs.
+func BuildDependencyGraph(txs []interface{}, bundles []*builder.Bundle, signer types.Signer) ([][]int, error) {
+	edges := make([][]int, len(txs))
```
hyeonLewis, line 33: I'm thinking of congested network, it will create n^2 edges but most of them will be sparse. Could we use edges as something like `map[int][]int` to only mark the tx we need?

hyeonLewis: I tried like below (need more tests I think), but it shows ~97% improvement.

<details><summary>Code</summary>

```
func buildDependencyIndices(txs []interface{}, bundles []*builder.Bundle, signer types.Signer) (map[common.Address][]int, map[int][]int, error) {
    senderToIndices := make(map[common.Address][]int)
    bundleToIndices := make(map[int][]int)

    for i, txOrGen := range txs {
        if tx, ok := txOrGen.(*types.Transaction); ok {
            from, err := types.Sender(signer, tx)
            if err != nil {
                return nil, nil, err
            }
            senderToIndices[from] = append(senderToIndices[from], i)

            if bundleIdx := FindBundleIdx(bundles, tx); bundleIdx != -1 {
                bundleToIndices[bundleIdx] = append(bundleToIndices[bundleIdx], i)
            }
        }
    }

    return senderToIndices, bundleToIndices, nil
}

func PopTxs2(txs *[]interface{}, num int, bundles *[]*builder.Bundle, signer types.Signer) {
    if len(*txs) == 0 || num == 0 {
        return
    }

    senderToIndices, bundleToIndices, err := buildDependencyIndices(*txs, *bundles, signer)
    if err != nil {
        logger.Error("Failed to build dependency indices", "err", err)
        ShiftTxs(txs, num)
        return
    }

    toRemove := make(map[int]bool)
    queue := make([]int, 0, num)

    for i := 0; i < min(num, len(*txs)); i++ {
        toRemove[i] = true
        queue = append(queue, i)
    }

    for len(queue) > 0 {
        curIdx := queue[0]
        queue = queue[1:]

        if tx, ok := (*txs)[curIdx].(*types.Transaction); ok {
            from, _ := types.Sender(signer, tx)
            for _, idx := range senderToIndices[from] {
                if idx > curIdx && !toRemove[idx] {
                    toRemove[idx] = true
                    queue = append(queue, idx)
                }
            }

            if bundleIdx := FindBundleIdx(*bundles, tx); bundleIdx != -1 {
                for _, idx := range bundleToIndices[bundleIdx] {
                    if !toRemove[idx] {
                        toRemove[idx] = true
                        queue = append(queue, idx)
                    }
                }
            }
        }
    }

    newTxs := Filter(txs, toRemove)
    
    bundleIdxToRemove := map[int]bool{}
    for bundleIdx, txIndices := range bundleToIndices {
		for _, txIdx := range txIndices {
			if toRemove[txIdx] {
				bundleIdxToRemove[bundleIdx] = true
				break
			}
		}
	}

    newBundles := SetCorrectTargetTxHash(Filter(bundles, bundleIdxToRemove), newTxs)

    *txs = newTxs
    *bundles = newBundles
}
```
</details>

```
BenchmarkPopTxs/PopTxs-11                       1000000                52.26 ns/op
BenchmarkPopTxs/PopTxs2-11                      1000000                1.656 ns/op
```

kaiax/builder/impl/getter.go
```diff
@@ -100,3 +200,162 @@ func (b *BuilderModule) IsConflict(prevBundles []*builder.Bundle, newBundles []*
...
+
+func ShiftTxs(txs *[]interface{}, num int) {
+	if len(*txs) <= num {
+		*txs = (*txs)[:0]
+		return
+	}
+	*txs = (*txs)[num:]
+}
+
+func PopTxs(txs *[]interface{}, num int, bundles *[]*builder.Bundle, signer types.Signer) {
```
hyeonLewis, line 293: It seems the current PopTxs can't handle the case that `num == 1` and `txs[0] == genTx`. The 

```
// Create test bundles
	bundles := []*builder.Bundle{
		{
			BundleTxs: []interface{}{g, txs[1], txs[2]},
		},
	}

	testCases := []struct {
		name            string
		incorporatedTxs []interface{}
		numToPop        int
		bundles         []*builder.Bundle
		expectedTxs     []interface{}
	}{
		{
			name:            "One bundle - first tx is generator",
			incorporatedTxs: []interface{}{g, txs[1], txs[2], txs[3], txs[4], txs[5], txs[6]},
			numToPop:        1,
			bundles:         []*builder.Bundle{bundles[0]},
			expectedTxs:     []interface{}{txs[4], txs[5],txs[6]},
		},
	}
```

Mdaiki0730: Ah, bundle[0] is [tx1, tx2], so when numPop=1, only the tx generator is deleted.
So it seems that the expected setting is incorrect.
```
                {
			name:            "One bundle - first tx is generator",
			incorporatedTxs: []interface{}{g, txs[1], txs[2], txs[3], txs[4], txs[5], txs[6]},
			numToPop:        1,
			bundles:         []*builder.Bundle{bundles[0]},
			expectedTxs:     []interface{}{txs[1], txs[2], txs[3], txs[4], txs[5],txs[6]},
		},
```

Mdaiki0730: I added a test case for "One bundle - first tx is generator". Thank you.

kaiax/builder/impl/getter.go
Mdaiki0730: Amazing! I've incorporated this change and the logic doesn't seem to be broken. Thank you.

work/worker.go
ulbqb: https://github.com/kaiachain/kaia/blob/47d8d543e61c7e8bbe23b731d0e5e32f72cc81d9/blockchain/tx_pool.go#L794-L796
Isn't `env.state.GetNonce(rewardbase)` correct? `getNonce(from) == tx.Nonce()` is allowed.

kaiax/builder/impl/getter.go
ian0371: It seems we need to draw edges between generator and txs if they're in the same bundle. So we need to add Lewis' testcase

kaiax/builder/impl/getter.go
ian0371: `newTargetTx` need to be type-asserted to `*types.Transaction`, because we know for sure that TargetTxHash can only be the type.

kaiax/builder/impl/getter.go
hyeonLewis: fyi. The PopTxs2 contains a fix of https://github.com/kaiachain/kaia/pull/272#discussion_r1986323832.

kaiax/builder/impl/getter.go
Mdaiki0730: Is the tx at index-1 in the beginning of the bundle guaranteed to be *types.Transaction?

ian0371: yes, because bundle can only be placed after tx.

kaiax/builder/impl/getter.go
Mdaiki0730: > Ah, bundle[0] is [tx1, tx2], so when numPop=1, only the tx generator is deleted.
So it seems that the expected setting is incorrect.

I overlooked the premise of the bundle.
I got fixed here: 7fa8532

kaiax/builder/impl/getter.go
Mdaiki0730: The generator may become a target due to the effects of pop etc.
e.g) 
[[A.tx1, A.tx2] [B.tx1, B.tx2, g] A.tx3 [C.tx1, C.tx2]]
if A.tx1 failed
[[B.tx1, B.tx2, g] [C.tx1, C.tx2]]

kaiax/builder/impl/getter.go
```diff
@@ -17,14 +17,41 @@
...
 	"github.com/kaiachain/kaia/blockchain/types"
 	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/crypto"
 	"github.com/kaiachain/kaia/kaiax/builder"
 )
 
+// buildDependencyIndices builds a dependency indices of txs.
+// Two txs with the same sender has an edge.
+// Two txs in the same bundle has an edge.
+// TxGenerator is not connected to any other txs.
```
ian0371, line 31: > // TxGenerator is not connected to any other txs.

Please update this comment

Mdaiki0730: It was removed in [d03e039](https://github.com/kaiachain/kaia/pull/272/commits/d03e0396a6f92b22715eac6dfa9b36f6fee5910e).

ulbqb (approved):

hyeonLewis (approved):

PR #275, opened 2025-03-11
kaiax/gasless: unify multiple gasless module instances

ulbqb:
## Proposed changes

There are two instances of gasless module as TxPoolMoudule and TxBundlingModule. However, TxBundlingModule's one doesn't call tx pool's reset so state db has never been set. This PR fixes this.

- Make two gasless modules instances one
- Register transaction pool modules by `RegisterTxPoolModule` after initializing transaction pool

Moreover, updating statedb is a little complicated so this changes the way to pass statedb to setting pool field in moduels. This also change the way to get sender address from a transaction.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/kaiax/gasless/impl/builder.go
+++ b/kaiax/gasless/impl/builder.go
@@ -30,7 +30,10 @@ func (g *GaslessModule) ExtractTxBundles(txs []*types.Transaction, prevBundles [
 	approveTxs := map[common.Address]*types.Transaction{}
 	targetTxHash := common.Hash{}
 	for _, tx := range txs {
-		addr := tx.ValidatedSender()
+		addr, err := types.Sender(g.signer, tx)
+		if err != nil {
+			panic(err)
+		}
 		if g.IsApproveTx(tx) {
 			approveTxs[addr] = tx
 		} else if g.IsSwapTx(tx) && g.IsExecutable(approveTxs[addr], tx) {
--- a/kaiax/gasless/impl/builder.go
+++ b/kaiax/gasless/impl/builder.go
@@ -32,7 +32,7 @@ func (g *GaslessModule) ExtractTxBundles(txs []*types.Transaction, prevBundles [
 	for _, tx := range txs {
 		addr, err := types.Sender(g.signer, tx)
 		if err != nil {
-			return nil
+			continue
 		}
 		if g.IsApproveTx(tx) {
 			approveTxs[addr] = tx
--- a/kaiax/gasless/impl/builder.go
+++ b/kaiax/gasless/impl/builder.go
@@ -32,7 +32,7 @@ func (g *GaslessModule) ExtractTxBundles(txs []*types.Transaction, prevBundles [
 	for _, tx := range txs {
 		addr, err := types.Sender(g.signer, tx)
 		if err != nil {
-			panic(err)
+			return nil
 		}
 		if g.IsApproveTx(tx) {
 			approveTxs[addr] = tx
--- a/kaiax/gasless/impl/init.go
+++ b/kaiax/gasless/impl/init.go
@@ -19,8 +19,8 @@ package impl
 import (
 	"crypto/ecdsa"
 
-	"github.com/kaiachain/kaia/blockchain/state"
 	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/kaiax"
 	"github.com/kaiachain/kaia/log"
 	"github.com/kaiachain/kaia/params"
 )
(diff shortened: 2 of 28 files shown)


Mdaiki0730:
Should we panic when types.Sender fails?
Since it's a module, I think it's safer to leave the processing to the host as much as possible.

ulbqb:
I was thinking of ignoring the error, but that may not be able to realize the wrong implementation on testing.

kaiax/gasless/impl/builder.go
```diff
@@ -30,7 +30,10 @@ func (g *GaslessModule) ExtractTxBundles(txs []*types.Transaction, prevBundles [
 	approveTxs := map[common.Address]*types.Transaction{}
 	targetTxHash := common.Hash{}
 	for _, tx := range txs {
-		addr := tx.ValidatedSender()
+		addr, err := types.Sender(g.signer, tx)
+		if err != nil {
+			panic(err)
```
ian0371, line 35:
```suggestion
			return nil
```

kaiax/gasless/impl/init.go
```diff
@@ -30,13 +31,14 @@ var logger = log.NewModuleLogger(log.KaiaxGasless)
 type InitOpts struct {
 	ChainConfig *params.ChainConfig
 	NodeKey     *ecdsa.PrivateKey
+	TxPool      kaiax.TxPoolForCaller
```
ian0371, line 34: It seems unused

kaiax/gasless/impl/tx_pool.go
```diff
@@ -80,8 +79,11 @@ func (g *GaslessModule) isApproveTxReady(approveTx, nextTx *types.Transaction) b
 
 // isSwapTxReady assumes that the caller checked `g.IsSwapTx(swapTx)`
 func (g *GaslessModule) isSwapTxReady(swapTx, prevTx *types.Transaction) bool {
-	addr := swapTx.ValidatedSender()
-	nonce := g.currentState.GetNonce(addr)
+	addr, err := types.Sender(g.signer, swapTx)
+	if err != nil {
+		panic(err)
```
ian0371, line 84:
```suggestion
		return false
```

kaiax/gasless/impl/init.go
ulbqb: It's being used for getting the current state db.

kaiax/gasless/impl/init.go
```diff
@@ -55,6 +57,7 @@ func (g *GaslessModule) Init(opts *InitOpts) error {
 	g.allowedTokens = map[common.Address]bool{
 		common.HexToAddress("0xabcd"): true,
 	}
+	g.signer = types.LatestSignerForChainID(g.ChainConfig.ChainID)
```
ian0371, line 60: Can you add `opts.TxPool` nil check above?

ulbqb:
@Mdaiki0730 
Honestly, either is fine. I removed panic.

kaiax/gasless/impl/builder.go
Mdaiki0730: What about continue?
Even if one abnormal tx fails, other bundles can still be generated.

kaiax/gasless/impl/init.go
```diff
@@ -55,6 +57,7 @@ func (g *GaslessModule) Init(opts *InitOpts) error {
 	g.allowedTokens = map[common.Address]bool{
 		common.HexToAddress("0xabcd"): true,
```
shiki-tak, line 58: Excuse me, this may be unrelated to this PR, but how do I set the actual swapRouters and allowedTokens values?

ulbqb: You can set them by only hard-code now.

ian0371: @shiki-tak We should deploy the contracts before release. I think we should use deterministic deploy so the address is the same across chains.

kaiax/gasless/impl/builder.go
ulbqb: fixed at https://github.com/kaiachain/kaia/pull/275/commits/18b2de490c0b5c019133d00da4656783223a8128

kaiax/gasless/impl/init.go
ulbqb: It's also good to set addresses in chain config.

ian0371: Actually, even if we used deterministic deploy, WKAIA address is different between mainnet and testnet, so Router will have different addresses.
ChainConfig sounds good.

ian0371 (approved):

PR #282, opened 2025-03-13
kaiax/gasless: fix some bugs in Task.ApplyTransactions

Mdaiki0730:
## Proposed changes

This PR fix following bugs in Task.ApplyTransactions.
- `env.tcount` used in `SetTxContext` within `commitBundleTransaction` is not being incremented.
- `numTxsChecked` does not increment correctly for bundles.
- When the return value of `bc.ApplyTransaction` is `err == nil && receipt.Status != ReceiptStatusSuccessful`, `ShiftTxs` is executed instead of `PopTxs`.

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/work/worker.go
+++ b/work/worker.go
@@ -40,6 +40,7 @@ import (
 	"github.com/kaiachain/kaia/kaiax/builder"
 	builder_impl "github.com/kaiachain/kaia/kaiax/builder/impl"
 	"github.com/kaiachain/kaia/kaiax/gov"
+	"github.com/kaiachain/kaia/kerrors"
 	kaiametrics "github.com/kaiachain/kaia/metrics"
 	"github.com/kaiachain/kaia/params"
 	"github.com/kaiachain/kaia/storage/database"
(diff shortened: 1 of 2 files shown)


hyeonLewis (commented):
I think https://github.com/kaiachain/kaia/pull/282/files#diff-aa53dfe64c81ad7f4673af6857ca8ae7873bb18224202adb09f765ce4f458411L773 should be updated also.

ulbqb (approved):

ulbqb:
> EEST is the successor to [ethereum/tests](https://github.com/ethereum/tests) (aka "original tests")

https://github.com/ethereum/execution-spec-tests/pull/1165/files#diff-710a99e33d5a7edcaa8f1464a11316b463987558f100a960a496318389ab0b20

hyeonLewis:
I found one more corner case.

In the `commitBundleTransaction`, it catches `err != nil || receipt.Status != types.ReceiptStatusSuccessful`. It means the `err` can be nil and switch statement recognizes this case as `case nil` so call `ShiftTxs` not `PopTxs`. I think we need to add custom error message for bundle call failure and make a separate case to pop the txs.

Mdaiki0730:
@hyeonLewis 
You're right, I'll change the PR name to broaden the scope of the fix.

ian0371 (approved):

work/worker.go
```diff
@@ -932,9 +935,14 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain,
 			}
 			*env.state = *lastSnapshot
 			env.header.GasUsed = gasUsedSnapshot
+			env.tcount = tcountSnapshot
+			if err == nil {
```
hyeonLewis, line 939: How about ignoring the original `err` and return `kerrors.ErrRevertedBundleByVmErr` always?

Mdaiki0730: I think we don't have to override this for proper logging and error counting.
Also, in the case of NonceTooLow, I think it is enough to use shift instead of pop, even if it is a tx in a bundle.
https://github.com/kaiachain/kaia/pull/282/files#diff-aa53dfe64c81ad7f4673af6857ca8ae7873bb18224202adb09f765ce4f458411R812-R816

hyeonLewis: Ok, thanks.

PR #289, opened 2025-03-18
kaiax/builder: Introduce TxOrGen

ian0371:
## Proposed changes

- Now gasless module initializes TxGenerator hash with `ApproveTx.Hash || SwapTx.Hash`. Previously, hash was initialized at `incorporate` lazily, which resulted in TxGenerators having incorrect hashes. TxGenerator's type is reverted back to function.
- Introduce `TxOrGen`; it replaces `[]interfaces{}` array from builder. `txOrGen` can be evaluated to tx by `GetTx(nonce)`.
- Change `Has(hash common.Hash)` -> `Has(txOrGen *TxOrGen)`. To find by hash, `FindIdx(id common.Hash)` can be used.
- Lambda functions are introduced in worker, such as `markAllTxUnexecutable`.
- Fix `tests/kaia_test_account_map_test.go` so txOrGen is evaluated to tx in the correct order.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/kaiax/builder/tx_or_gen.go
+++ b/kaiax/builder/tx_or_gen.go
@@ -8,37 +8,37 @@ import (
 )
 
 type TxOrGen struct {
-	ConcreteTx  *types.Transaction
-	TxGenerator TxGenerator
+	concreteTx  *types.Transaction
+	txGenerator TxGenerator
 	Id          common.Hash
 }
 
 func (t *TxOrGen) IsConcreteTx() bool {
-	return t.ConcreteTx != nil
+	return t.concreteTx != nil
 }
 
 func (t *TxOrGen) IsTxGenerator() bool {
-	return t.TxGenerator != nil
+	return t.txGenerator != nil
 }
 
 func NewTxOrGenFromTx(tx *types.Transaction) *TxOrGen {
+	if tx == nil {
+		return nil
+	}
+
 	return &TxOrGen{
-		ConcreteTx: tx,
+		concreteTx: tx,
 		Id:         tx.Hash(),
 	}
 }
 
-func NewTxOrGenListFromTxs(txs ...*types.Transaction) []*TxOrGen {
-	txOrGens := make([]*TxOrGen, len(txs))
-	for i, tx := range txs {
-		txOrGens[i] = NewTxOrGenFromTx(tx)
-	}
-	return txOrGens
-}
-
 func NewTxOrGenFromGen(generator TxGenerator, id common.Hash) *TxOrGen {
+	if generator == nil {
+		return nil
+	}
+
 	return &TxOrGen{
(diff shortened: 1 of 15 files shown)


kaiax/builder/tx_or_gen.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
+)
+
+type TxOrGen struct {
+	concreteTx  *types.Transaction
+	txGenerator TxGenerator
+	Id          common.Hash
```
ulbqb, line 13: Please explain what Id should be like.

ian0371: fixed 6c9182006

tests/kaia_test_account_map_test.go
```diff
@@ -112,17 +112,10 @@ func (a *AccountMap) Initialize(bcdata *BCData) error {
...
 	for _, txOrGen := range incorporatedTxs {
-		if gen, ok := txOrGen.(builder.TxGenerator); ok {
-			// To simulate tx, the nonce given to generate is set to zero.
-			// This does not affect subsequent operations on the AccountMap state.
-			tx, err := gen.Generate(0)
-			if err != nil {
-				continue
-			}
-			txs = append(txs, tx)
+		tx, err := txOrGen.GetTx(0)
```
Mdaiki0730, line 115: Would you like to leave these comments?
```
// To simulate tx, the nonce given to generate is set to zero.
// This does not affect subsequent operations on the AccountMap state.
```

tests/kaia_test_account_map_test.go
```diff
@@ -112,17 +112,10 @@ func (a *AccountMap) Initialize(bcdata *BCData) error {
...
+		if err != nil {
+			return err
```
Mdaiki0730, line 117: This seems like a good enough continue, but what do you think?
```suggestion
			continue
```

tests/kaia_test_account_map_test.go
ian0371: fixed cd06022

tests/kaia_test_account_map_test.go
ian0371: Wouldn't it be better to explicitly return an error so caller can fix on error, rather than implicitly skipping? Also, this function already propagates subroutine errors.

Mdaiki0730: I agree with the opinion. Thank you.

Mdaiki0730 (approved):

ulbqb (approved):

PR #290, opened 2025-03-20
api: Added pending tx subscription

hyunsooda:
## Proposed changes

This PR introduces two new types of subscription APIs: pending transactions with hash-only and with bodies. Additionally, it refactors the existing code to remove unnecessary components and resolve a potential import cycle.

This is a running sample snippet:
```go
package main

import (
    "context"
    "fmt"

    "github.com/kaiachain/kaia/blockchain/types"
    "github.com/kaiachain/kaia/client"
    "github.com/kaiachain/kaia/common"
)

func main() {
    var (
        headCh   = make(chan *types.Header)
        txCh     = make(chan *types.Transaction)
        txHashCh = make(chan common.Hash)
    )
    c, _ := client.Dial("ws://localhost:8146")
    c.SubscribeNewHead(context.Background(), headCh)
    c.SubscribeFullPendingTransactions(context.Background(), txCh)
    c.SubscribePendingTransactions(context.Background(), txHashCh)
    for {
        select {
        case header := <-headCh:
            fmt.Println("header arrived", header.Number)
        case txHash := <-txHashCh:
            fmt.Println("tx hash arrived", txHash.String())
        case tx := <-txCh:
            fmt.Println("full tx arrived", tx)
        }
    }
}
```

@yoomee1313 @tnasu @2dvorak PTAL.
@hyeonLewis I believe this subscription does not need to be in the auction module, so I placed it in a more general location.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments


node/cn/filters/api.go
```diff
@@ -475,10 +450,29 @@ func (api *PublicFilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
...
 			hashes := f.hashes
 			f.hashes = nil
 			return returnHashes(hashes), nil
+		case PendingTransactionsSubscription:
+			if f.fullTx {
+				latest, err := api.backend.HeaderByNumber(context.Background(), rpc.LatestBlockNumber)
+				if err == nil && latest != nil {
+					txs := make([]map[string]interface{}, 0, len(f.txs))
+					for idx, tx := range f.txs {
+						txs = append(txs, kaiaApi.NewRPCTransaction(latest, tx, latest.Hash(), latest.Number.Uint64(), uint64(idx), chainConfig))
```
hyeonLewis, line 463: I think `NewRPCTransaction` is misused and it's enough with `tx.MakeRPCOutput()`. The `NewRPCTransaction` assumes the tx is already mined so it needs block metadata (e.g., `index`)

hyeonLewis (commented):
I thought we decided to separate the `newPendingTransactionBodies` to separate namespace (`auction_`) since it's heavy.

node/cn/filters/api.go
```diff
@@ -169,22 +171,29 @@ func (api *PublicFilterAPI) NewPendingTransactions(ctx context.Context) (*rpc.Su
...
 
 		for {
 			select {
-			case hashes := <-txHashes:
+			case txs := <-txs:
 				// To keep the original behaviour, send a single tx hash in one notification.
 				// TODO(rjl493456442) Send a batch of tx hashes in one notification
-				for _, h := range hashes {
-					notifier.Notify(rpcSub.ID, h)
+				latest, err := api.backend.HeaderByNumber(ctx, rpc.LatestBlockNumber)
```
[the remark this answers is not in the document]
hyunsooda: Yes, not necessary. Removed. Thank you

node/cn/filters/api.go
hyunsooda: I've updated it. Thank you

ian0371 (commented):
Please don't merge into dev but use a feature branch instead.

hyunsooda:
@hyeonLewis There seems two approaches to add `time` field.
1. Manually adding `time` field into all tx types 
2. Get rpc output from each tx type first and add `time` field into returned map. (seems better)

I'm considering second method and please give your input on this.
Additionally, I'm considering to repoen this PR, targeting `feat/gasless-auction`.

hyunsooda:
@hyeonLewis  In the latest commit, I added a new API that accepts a map (raw data) type to prevent tx.time from being set to the current time during decoding.

This PR introduces three new APIs in total. The first two are already available in Geth, so we may also export them under the kaia namespace.

blockchain/types/transaction.go
```diff
@@ -591,6 +590,10 @@ func (tx *Transaction) Size() common.StorageSize {
 	return size
 }
 
+func (tx *Transaction) SetTime(t time.Time) {
```
hyeonLewis, line 593: It seems this is for old design.

hyeonLewis:
@hyunsooda Oh, I wasn't aware of it. Then we could do the same way as geth did for pending transactions.

blockchain/types/transaction.go
hyunsooda: Apologies for the missing context.

This may be needed on the client side to set the time field in tx when processing the raw data (map type). However, it is not directly related to the outdated design. If setting the time field in tx is unnecessary—because the client already has the correct time value—then this step can be skipped. (fyi, this public function is also available in Geth.)

hyunsooda:
Can then we add both namespaces `kaia` and `auction`? Or, just only for auction?

hyunsooda:
Will be rebased and reimplemented in another PR

PR #305, opened 2025-04-02
kaiax/gasless: Fix restore env upon revert

ian0371:
## Proposed changes

- Fix recovering `env.state` with snapshot (`work/worker.go`). It fixes bad block issue when a tx in bundle reverts (`Error: invalid gas used (remote: 461415 local: 463655)`), which is caused by shallow-copying of `statedb.stateObjects`.
  - Instead of using a new `statedb` object from `commitBundleTransaction()` (i.e., `env.state = lastSnapshot`), old `statedb` must be reused because `bc.State()` callers will use it. Thus, `func (s *StateDB) Set(src *StateDB)` is introduced.
  - Now, all fields are copied by `copyStateDB()`.
- Update `GenABlockWithTransactionsWithBundle` to filter failing transactions
- Remove `txHashesExpectedFail` argument from `Update` and `GenABlockWithTransactions`, so now it's the same as  [v1.0.3](https://github.com/kaiachain/kaia/blob/f329a688e17b7741d35f2b4736c05d3d7a77d430/tests/kaia_test_blockchain_test.go#L355-L356)
- Fix log15 error (`kaiax/gasless/impl/getter.go`)

## Types of changes

- [x] Bugfix
- [ ] Others

## Checklist

### DIFF
--- a/work/worker.go
+++ b/work/worker.go
@@ -907,7 +907,7 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain, 
 	}
 
 	restoreEnv := func() {
-		*env.state = *lastSnapshot
+		env.state = lastSnapshot
 		env.header.GasUsed = gasUsedSnapshot
 		env.tcount = tcountSnapshot
 	}
@@ -929,6 +929,7 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain, 
 			if err != vm.ErrInsufficientBalance && err != vm.ErrTotalTimeLimitReached {
 				markAllTxUnexecutable()
 			}
+			logger.Error("ApplyTransaction error, restoring env", "error", err)
 			restoreEnv()
 			if err == nil {
 				err = kerrors.ErrRevertedBundleByVmErr
--- a/work/worker.go
+++ b/work/worker.go
@@ -907,7 +907,7 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain, 
 	}
 
 	restoreEnv := func() {
-		env.state = lastSnapshot
+		env.state.Set(lastSnapshot)
 		env.header.GasUsed = gasUsedSnapshot
 		env.tcount = tcountSnapshot
 	}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,14 @@ executors:
         auth:
           username: $DOCKER_LOGIN
           password: $DOCKER_PASSWORD
+  test-tests-executor: # this executor is for test-tests job
+    working_directory: ~/go/src/github.com/kaiachain/kaia
+    resource_class: medium+
+    docker:
+      - image: kaiachain/build_base:go1.23-solc0.8.13-ubuntu-22.04
+        auth:
+          username: $DOCKER_LOGIN
+          password: $DOCKER_PASSWORD
   test-others-executor: # this executor is for test-others job
     working_directory: /go/src/github.com/kaiachain/kaia
     resource_class: xlarge
@@ -396,7 +404,7 @@ jobs:
           no_output_timeout: 30m
           command: make test-node
   test-tests:
-    executor: test-executor
+    executor: test-tests-executor
     steps:
       - checkout
       - run:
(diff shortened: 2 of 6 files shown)


Mdaiki0730:
> Update GenABlockWithTransactionsWithBundle to filter failing transactions

This fix does not test revert because it removes transactions from the list before creating a block.
Would you like to undo the changes in test?

ian0371:
@Mdaiki0730 It was because the test fails due to invalid merkle root:
```
$ go test ./tests -run=TestTxBundle
--- FAIL: TestTxBundle (3.06s)
    --- FAIL: TestTxBundle/TestTxBundleRevert (0.35s)
        kaia_scenario_test.go:2073: err = invalid merkle root (remote: 50f0a818a4d39c2c41a73eb2e5c76eced007b1b768eb695304959fdd356ee2cf local: ad1a7cfac0d2affa89b58a8f653826a5ea8d1cf512b46963d54331239d996294), n = 0

    --- FAIL: TestTxBundle/TestTxBundleRevertByEvmError (1.99s)
        kaia_scenario_test.go:2166: err = invalid merkle root (remote: 0fb833ac89e8b9c10d4de405013554710b2eb17451e8e3326e82dba0e1ca3470 local: 1bb3d9e364647e280a9e352e345eac4fcb84f4e680bf1c733b4ff61bb6409201), n = 0

FAIL
FAIL    github.com/kaiachain/kaia/tests 3.885s
FAIL
```

I assumed `MineABlock()` should exclude the reverting txs, just like the previous `Update` logic. So I simply merged the logic and put it in `GenABlockWithTransactionsWithBundle`.

But I understand your point. What do you think should be the new logic of `MineABlock()`? Do you think we need to implement restore-env-on-revert logic inside `MineABlock()`? If so, would it make a big difference from the logic in this PR?

Mdaiki0730:
Since `MineABlock()` executes task.ApplyTransaction, it is expected that revert will occur even without any changes to test.
https://github.com/kaiachain/kaia/blob/dev/tests/kaia_test_blockchain_test.go#L270

Do you know what causes `invalid merkle root`?

work/worker.go
```diff
@@ -907,7 +905,7 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain,
 	}
 
 	restoreEnv := func() {
-		*env.state = *lastSnapshot
+		env.state = lastSnapshot
```
hyeonLewis, line 908: Just a question: How did it lead to different gas used?

ian0371:
@Mdaiki0730 Oh I see. I reverted for now. 5ed118c

work/worker.go
ian0371: I'm not sure exactly, but I guess some member variables of `env.state` are not completely restored by the assignment.

hyeonLewis: It might be related to cold storage access since the gas difference is close to 2100.

Maybe when tx has been reverted, the `env.state` will remember the context of reverted tx, so the next tx considers it warm storage, while validators consider it cold storage. I think it's the most important part, so it's worth investigating to find the root cause. Please share your thoughts. Thanks!

ian0371: I confirmed that bad block occurs when shallow-copying state objects as follows:

```
func copyStateDB(dst, src *StateDB) {
	...
	dst.stateObjects = src.stateObjects
	dst.stateObjectsDirty = src.stateObjectsDirty
	dst.stateObjectsDirtyStorage = src.stateObjectsDirtyStorage
}
```

ian0371: fyi, `maps.Clone` is a shallow copy.

blukat29 (approved):

work/worker.go
hyeonLewis: Does it also fix the copy-of-copy issue?

ulbqb (approved):

work/worker.go
ian0371: What is the issue? Is it related to this test? 
https://github.com/kaiachain/kaia/blob/beb73629c0bf41af1f07100f416448daf5d7dc5c/blockchain/state/statedb_test.go#L139

hyeonLewis: Isn't `Two deep-copy of statedb upon bundle fail (c.f. one deep-copy upon bundle success).` the issue? I saw this in the scrum today.

ian0371: Oh, I meant upon bundle failure, two deep-copies (taking a snapshot, and restoring the old object) take place. Sorry for the confusion

hyeonLewis: Oh, I misunderstood it. Thanks!

hyeonLewis (approved):

PR #136, opened 2024-11-21
Feat/upgrade build base

Sotatek-TinnNguyen:
## Proposed changes

Upgrade build_base image:
- Upgrade ubuntu from 20.04 to 22.04
- Upgrade golang version to 1.23
- Upgrade goangci-lint to 1.60.1
Upgrade rpm_package:
- Replace centos7 to centos stream 9 due to deprecated of centos7 

## Types of changes

Please put an x in the boxes related to your change.

- [ ] Bugfix
- [ ] New feature or enhancement
- [ ] Others

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING GUIDELINES](https://github.com/kaiachain/kaia/blob/main/CONTRIBUTING.md) doc
- [ ] I have read the [CLA](https://gist.github.com/kaiachain-dev/bbf65cc330275c057463c4c94ce787a6) and signed by comment `I have read the CLA Document and I hereby sign the CLA` in first time contribute
- [ ] Lint and unit tests pass locally with my changes (`$ make test`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Related issues

- Please leave the issue numbers or links related to this PR here.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #586, opened 2025-10-15
Upgrade Go to 1.25

EthGiveawa375:
Upgrade the project to Go 1.25 to resolve toolchain incompatibilities and align builds.


blukat29:
Hello @EthGiveawa375. Thanks for your interest. However, upgrading the build environment needs more work than modifying this Dockerfile. I need to push a new docker image to our DockerHub, which only our team has the permission. We'll handle it like we did before in https://github.com/kaiachain/kaia/pull/136.

PR #639, opened 2025-11-17
work/worker: Refine tx execution time related logs

hyunsooda:
## Proposed changes

This PR fixes and refines mining logs.

1. auction module:
- Added the error log why the bundle was skipped. Note that the level is bumped to warning level
2. time-constraint log
-   The error log "A single transaction exceeds total time limit" seems wrong, placed from many years ago. According to the condition https://github.com/kaiachain/kaia/blob/dev/work/worker.go#L762, it is an unreachable path. I also reviewed the logs from DD and Grafana, but couldn't find their recorded history. This PR handles it together
3. In both Kairos and Mainnet, "Transaction aborted due to time limit"  logs can be found a lot.  The PR's version shows a detailed execution status, including the number of normal tx and bundle tx.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist

## Related issues

## Further comments

### DIFF
--- a/work/worker.go
+++ b/work/worker.go
@@ -765,6 +765,13 @@ func (env *Task) ApplyTransactions(txs *types.TransactionsByPriceAndNonce, bc Bl
 				if env.tcount > 0 && atomic.LoadInt32(&isExecutingBundleTxs) == 0 {
 					// The total time limit reached, thus we stop the currently running EVM.
 					evm.Cancel(vm.CancelByTotalTimeLimit)
+				} else if env.tcount == 0 && len(arrayTxs) > 0 {
+					// Case 'Single long tx':
+					//   T0 (executing tx) ------- T_limit --> abort=1 (but let T0 finish)
+					//   Result: tcount=0 → Log "A single transaction exceeds limit" and "unexecuted transactions due to time limit"
+					//
+					logger.Warn("A single transaction exceeds total time limit", "hash", arrayTxs[0].Hash().String())
+					tooLongTxCounter.Inc(1)
 				}
 				evm = nil
 			}
@@ -779,7 +786,6 @@ func (env *Task) ApplyTransactions(txs *types.TransactionsByPriceAndNonce, bc Bl
 	var numTxsNonceTooLow int64 = 0
 	var numTxsNonceTooHigh int64 = 0
 	var numTxsGasLimitReached int64 = 0
-	txAbortedByTimeLimit := false // is true when a transaction was aborted mid-execution due to time limit. Note that the first transaction or first bundle of a block does not abort.
 
 CommitTransactionLoop:
 	for atomic.LoadInt32(&abort) == 0 {
@@ -879,7 +885,6 @@ CommitTransactionLoop:
 
 		case vm.ErrTotalTimeLimitReached:
 			logger.Warn("Transaction aborted due to time limit", "hash", tx.Hash().String())
-			txAbortedByTimeLimit = true
 			// NOTE-Kaia Exit for loop immediately without checking abort variable again.
 			break CommitTransactionLoop
 
@@ -917,13 +922,9 @@ CommitTransactionLoop:
 		}
 	}
 
-	// Case 1: Single long tx
-	//   T0 (executing tx) ------- T_limit --> abort=1 (but let T0 finish)
-	//   Result: txAbortedByTimeLimit=false, tcount=1 → Log "unexecuted transactions due to time limit" and "A single transaction exceeds limit"
-	//
-	// Case 2: Multiple txs, limit hit after first tx
+	// Case 'Multiple txs, limit hit after first tx':
 	//   T0 -- ... -- TN (executing tx) -- T_limit --> abort=1 (cancel TN immediately)
-	//   Result: txAbortedByTimeLimit=true, tcount=N → Log "unexecuted transactions due to time limit"
+	//   Result: tcount=N → Log "unexecuted transactions due to time limit"
 	if atomic.LoadInt32(&abort) == 1 {
 		timeLimitReachedCounter.Inc(1)
 		var txHash common.Hash
@@ -942,10 +943,6 @@ CommitTransactionLoop:
 				"firstUnexecutedTxHash", txHash.String(),
 				"firstUnexecutedTxInBundle", builder.FindBundleIdx(bundles, txOrGen) != -1,
 			)
-		}
-		if !txAbortedByTimeLimit && env.tcount == 1 && len(arrayTxs) > 0 {
-			logger.Warn("A single transaction exceeds total time limit", "hash", arrayTxs[0].Hash().String())
-			tooLongTxCounter.Inc(1)
 		}
 	}
 


work/worker.go
```diff
@@ -1015,17 +1037,22 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain,
...
-		return bundle.TargetTxHash != common.Hash{}
+		return bundle.TargetTxHash != common.Hash{}, errors.New("absent of target tx hash")
 	} else {
-		return !(bundle.TargetTxHash == env.txs[env.tcount-1].Hash() &&
-			env.receipts[env.tcount-1].Status == types.ReceiptStatusSuccessful)
+		if bundle.TargetTxHash != env.txs[env.tcount-1].Hash() {
+			return true, errors.New("target tx hash not found:" + bundle.TargetTxHash.String())
+		}
+		if env.receipts[env.tcount-1].Status != types.ReceiptStatusSuccessful {
+			return true, errors.New(fmt.Sprintf("target tx failed (status = %d)", env.receipts[env.tcount-1].Status))
```
blukat29:
```suggestion
			return true, fmt.Errorf("target tx %s failed with status %d", bundle.TargetTxHash.Hex(), env.receipts[env.tcount-1].Status))
```

work/worker.go
```diff
@@ -1015,17 +1037,22 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain,
 	return nil, nil, logs
 }
 
-func (env *Task) shouldDiscardBundle(bundle *builder.Bundle) bool {
+func (env *Task) shouldDiscardBundle(bundle *builder.Bundle) (bool, error) {
 	if !bundle.TargetRequired {
-		return false
+		return false, nil
 	}
```
blukat29:
```suggestion
	if !bundle.TargetRequired {
		return false, nil
	}
	if bundle.TargetTxHash == common.Hash{} {
		return false, nil
	}
```

work/worker.go
```diff
@@ -1015,17 +1037,22 @@ func (env *Task) commitBundleTransaction(bundle *builder.Bundle, bc BlockChain,
...
 
 	if env.tcount == 0 {
-		return bundle.TargetTxHash != common.Hash{}
+		return bundle.TargetTxHash != common.Hash{}, errors.New("absent of target tx hash")
```
blukat29:
```suggestion
	if env.tcount == 0 {
		return true, fmt.Errorf("target tx %s does not precede the bundle", bundle.TargetTxHash.Hex())
	}
```

work/worker.go
blukat29: Let's use early-return pattern.
- !TargetRequired, TargetTxHash == zero, tcount == 0.

work/worker.go
hyunsooda: This suggestion is understood based on https://github.com/kaiachain/kaia/pull/639#discussion_r2536116077. A snippet below seems more readable. PTAL.
```
if env.tcount == 0 && bundle.TargetTxHash != (common.Hash{}) {
```

hyunsooda: On second thought, the suggestion looks better. @hyeonLewis Do you have some inputs here?

hyeonLewis: Both ways look good, but slightly prefer the suggestion.

work/worker.go
```diff
@@ -914,6 +910,31 @@ CommitTransactionLoop:
...
+	if atomic.LoadInt32(&abort) == 1 {
+		timeLimitReachedCounter.Inc(1)
+		var txHash common.Hash
+		if len(incorporatedTxs) > 0 {
+			txOrGen := incorporatedTxs[0]
+			tx, err := txOrGen.GetTx(env.state.GetNonce(nodeAddr))
+			if err == nil {
+				txHash = tx.Hash()
+			}
+			totalTxs := len(arrayTxs) + len(bundles)
```
hyeonLewis: This isn't total txs. We need to check `BundleTxs` in each bundle.

work/worker.go
```diff
@@ -914,6 +910,31 @@ CommitTransactionLoop:
...
+				"txs", len(arrayTxs),
+				"bundlesTxs", len(bundles),
+				"totalTxs", totalTxs,
+				"executed", env.tcount,
+				"unexecuted", totalTxs-env.tcount,
+				"lastUnexecuted", txHash.String(),
+			)
+		}
+		if env.tcount == 1 && len(arrayTxs) > 0 {
+			logger.Error("A single transaction exceeds total time limit", "hash", arrayTxs[0].Hash().String())
```
hyeonLewis: This might not be true when abortion happens during the execution of the second transaction. For example, `consumedByFirstTx = 200ms, consumedBySecondTx = 100ms`.

work/worker.go
```diff
@@ -914,6 +910,31 @@ CommitTransactionLoop:
...
+		timeLimitReachedCounter.Inc(1)
+		var txHash common.Hash
+		if len(incorporatedTxs) > 0 {
+			txOrGen := incorporatedTxs[0]
+			tx, err := txOrGen.GetTx(env.state.GetNonce(nodeAddr))
+			if err == nil {
+				txHash = tx.Hash()
+			}
+			totalTxs := len(arrayTxs) + len(bundles)
+			logger.Warn("Unexecuted transactions due to time limit",
```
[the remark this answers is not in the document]
hyunsooda: The purpose of the PR was to determine how many subsequent txs, including bundles, were being excluded or cut off. On the other hand, the scope of the PR's changes might be considered excessive.

This is the main motivation of that PR. So, let's establish a consensus here first, whether the motivation hits. @blukat29 , Please add your input on this.

work/worker.go
hyunsooda: Fixed at https://github.com/kaiachain/kaia/pull/639/commits/68afe27530cbd43ce3a5336e43612424a1627032

work/worker.go
```diff
@@ -914,6 +910,31 @@ CommitTransactionLoop:
...
+				"txs", len(arrayTxs),
+				"bundlesTxs", len(bundles),
+				"totalTxs", totalTxs,
+				"executed", env.tcount,
```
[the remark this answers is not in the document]
hyunsooda: That log appears only when EVM is aborted due to a time constraint, under the condition specified in https://github.com/kaiachain/kaia/blob/dev/work/worker.go#L762. So, the first heavy transaction is not the scope of this log. 

I chose not to remove this log because its format has been in service for many years and should be preserved. Simultaneously, the additions in this PR provide a detailed breakdown of the cut-off transactions.

work/worker.go
hyunsooda: The change is identical with the previous one. I guess that you're mentioning to add `env.tcount == 0` to this condition https://github.com/kaiachain/kaia/commit/68afe27530cbd43ce3a5336e43612424a1627032#diff-aa53dfe64c81ad7f4673af6857ca8ae7873bb18224202adb09f765ce4f458411R1045 right? Since the target tx is not specified, `env.tcount` does not appear to need to be looked up.

work/worker.go
hyeonLewis: > Because bundle is all or nothing, I intended to indicate a single tx, rather than each tx in the bundle.

Then how do we use `env.tcount` for the `executed` field? Expanding the concept of bundle here breaks semantics for me. I think leaving `txs`, `bundleTxs` (`bundles` looks better to me) and deleting `totalTxs` would be better.

And I think we should also consider GA bundle here. The `approval`, `swap` tx will be included in `arrayTxs` since they're coming from the tx pool, not a separate channel like the auction bundle.

work/worker.go
hyeonLewis: I meant the log message `A single transaction exceeds total time limit`. When the second tx got caught by the time constraint, then `env.tcount == 1 && len(arrayTxs) > 0` but it's not by a **single** transaction.

work/worker.go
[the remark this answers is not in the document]
hyeonLewis: I think we can continue discussing it in https://github.com/kaiachain/kaia/pull/639#discussion_r2536289369.

work/worker.go
hyeonLewis: > Since the target tx is not specified

It means we must check `env.tcount == 0` since if not, we should discard bundle.

Sorry for the unclear point; what I preferred was to return `true` or `false` explicitly, following other branches.

hyunsooda: At the code level, both look the same. I couldn't get your point. Could you suggest a concrete code snippet?

work/worker.go
hyunsooda: Ahh, okay, you're arguing the semantics of the message. First of all, I preserved the message as it is. Your scenario would work, but a heavy single tx scenario would also hit. 

I'll make a state variable that indicates `vm.ErrTotalTimeLimitReached` was met. If that variable is false, then the current message would work. Please add your input on the expected change.

work/worker.go
```diff
@@ -872,11 +876,6 @@ CommitTransactionLoop:
 
 		case vm.ErrTotalTimeLimitReached:
 			logger.Warn("Transaction aborted due to time limit", "hash", tx.Hash().String())
-			timeLimitReachedCounter.Inc(1)
```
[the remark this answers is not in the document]
hyeonLewis: The `evm.Cancel(vm.CancelByTotalTimeLimit)` only happens when `env.tcount == 0`, which means the first transaction won't be affected by `CancelByTotalTimeLimit` but will keep being executed. What Lake mentioned is only below branch.

https://github.com/kaiachain/kaia/blob/dev/work/worker.go#L876-L879

work/worker.go
```diff
@@ -914,6 +913,30 @@ CommitTransactionLoop:
...
+			if err == nil {
+				txHash = tx.Hash()
+			}
+			logger.Warn("Unexecuted transactions due to time limit",
+				"txs", len(arrayTxs),
+				"bundlesTxs", len(bundles),
+				"totalTxs", totalTxs,
+				"executed", env.tcount,
+				"unexecuted", totalTxs-env.tcount,
+				"lastUnexecuted", txHash.String(),
```
ian0371:
```suggestion
				"firstDroppedTx", txHash.String(),
```

work/worker.go
hyunsooda: Fixed at  https://github.com/kaiachain/kaia/pull/639/commits/9174d7ac001b91ae724141086de63e8a1776b353

work/worker.go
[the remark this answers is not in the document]
hyunsooda: I found this comment after pushing this [comment](https://github.com/kaiachain/kaia/pull/639/commits/9bd206ab8c4b1e19c266944948c6e66823e784e3)
I combined your suggestion and confirmed its working. Let's continue on this thread https://github.com/kaiachain/kaia/pull/639#discussion_r2536768378

work/worker.go
```diff
@@ -914,6 +916,31 @@ CommitTransactionLoop:
...
+				txHash = tx.Hash()
+			}
+			logger.Warn("Unexecuted transactions due to time limit",
+				"totalTxs", totalTxs,
+				"totalBundles", totalBundles,
+				"executedTxs", env.tcount,
+				"unexecutedTxs", len(incorporatedTxs),
+				"unexecutedBundles", len(bundles),
+				"firstDroppedTxHash", txHash.String(),
+				"isFirstDroppedTxBundle", builder.FindBundleIdx(bundles, txOrGen) != -1,
```
hyunsooda: Request change: `firstUnexecutedTxInBundle`

hyunsooda:
@blukat29 Description and field name changes are at https://github.com/kaiachain/kaia/pull/639/commits/f0751d3ef012543e7f15c212ea431cc05a6f3edd

work/worker.go
```diff
@@ -774,6 +779,7 @@ func (env *Task) ApplyTransactions(txs *types.TransactionsByPriceAndNonce, bc Bl
 	var numTxsNonceTooLow int64 = 0
 	var numTxsNonceTooHigh int64 = 0
 	var numTxsGasLimitReached int64 = 0
+	txAbortedByTimeLimit := false // indicate that first tx is elapsed over execution time budget
```
blukat29: Isn't it the opposite?

txAbortedByTimeLimit // is true when a transaction was aborted mid-execution due to time limit. Note that the first transaction or first bundle of a block does not abort.

hyunsooda: Ahh, right. I was too biased and focused on a `false` value. Your description is correct. Fixed.

work/worker.go
```diff
@@ -914,6 +917,31 @@ CommitTransactionLoop:
 		}
 	}
 
+	if atomic.LoadInt32(&abort) == 1 {
```
hyeonLewis: How do you think about adding detailed comments? If it feels so verbose, please ignore this.

```suggestion
	// Case 1: Single long tx
	//   T0 (executing tx) ------- T_limit --> abort=1 (but let T0 finish)
	//   Result: txAbortedByTimeLimit=false, tcount=1 → Log "single transaction exceeds limit"
	//
	// Case 2: Multiple txs, limit hit after first tx
	//   T0 -- ... -- TN (executing tx) -- T_limit --> abort=1 (cancel TN immediately)
	//   Result: txAbortedByTimeLimit=true, tcount=N → Log "unexecuted transactions due to time limit"
	if atomic.LoadInt32(&abort) == 1 {
```

PR #645, opened 2025-11-25
kaiax/gov: Remove params.{stake,proposer}UpdateInterval

ian0371:
## Proposed changes

- Remove global variables `params.stakeUpdateInterval`, `params.proposerUpdateInterval` and related functions.
- One must use the values from ChainConfig instead, because they are immutable. ~~It falls back to the default value if `ChainConfig.Governance.Reward` is nil, which I assume is a rare case even for private net.~~ fyi,  `ChainConfig.Governance` can never be nil due to [this reason](https://github.com/kaiachain/kaia/pull/645#discussion_r2559942779).
- New codes are unlikely to use those value because they are practically deprecated.

## Types of changes

- [ ] Bugfix
- [x] Others

## Checklist

### DIFF
--- a/params/governance_params.go
+++ b/params/governance_params.go
@@ -22,11 +22,6 @@ import (
 	"math/big"
 )
 
-var (
-	stakingUpdateInterval  uint64 = DefaultStakeUpdateInterval
-	proposerUpdateInterval uint64 = DefaultProposerRefreshInterval
-)
-
 const (
 	// Block reward will be separated by three pieces and distributed
 	RewardSliceCount      = 3


params/governance_params.go
```diff
@@ -112,61 +106,6 @@ var (
 	DefaultDeriveShaImpl             = uint64(0)     // Orig
 )
 
-func IsStakingUpdateInterval(blockNum uint64) bool {
```
hyunsooda: For tidyng up instead of inlining, is it not better refactor and revive this function by adding additional param for interval value?

hyunsooda: For tidying up instead of inlining, is it not better to refactor and reuse this function by adding an additional parameter for the interval value?

PR #660, opened 2025-12-03
blockchain/types: alleviate sig cache check

yoomee1313:
## Proposed changes

sigCache stores a recovered public key so it can be reused instead of performing the expensive recovery again. In the past, multiple Signer types existed, so the cache was frequently invalidated. 

With modernSigner, the rules(signer struct) are unified. modernSigner works under **the assumption that a given txType always uses the same signature algorithm and the same signing rules**. Because of this assumption, sigCache only needs to check `IsSupported(txType)` to safely reuse the cached result, while the stricter checks—such as whether the full set of supported txTypes matches—are handled separately by Equal.

## Types of changes

- [ ] Bugfix
- [ ] Others

## Checklist


shiki-tak (approved):
I also confirmed that TPS improved after applying the PR fix.

PR #729, opened 2026-01-28
consensus/istanbul/core: Remove validator.go dependency

tnasu:
## Proposed changes

There is duplicate `kaiax/valset` logic in consensus. For deduplicating, the dependency should be removed. This PR refactors the Istanbul consensus core to decouple it from `RoundCommitteeState`, `BlockValSet`, and `validator.go`. The core now uses `valsetModule` and `govModule` directly for committee state, and keeps round committee data as local variables and roundState fields without depending on types from the istanbul validator package for this logic.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

After this PR, `consensus/istanbul/backend` package should be refactored to be the same as this PR

### DIFF
--- a/consensus/istanbul/core/core.go
+++ b/consensus/istanbul/core/core.go
@@ -223,7 +223,11 @@ func (c *core) currentView() *istanbul.View {
 }
 
 func (c *core) isProposer() bool {
-	return c.current.proposer == c.backend.Address()
+	v := c.current
+	if v == nil {
+		return false
+	}
+	return v.proposer == c.backend.Address()
 }
 
 func (c *core) commit() {


consensus/istanbul/core/roundstate.go
```diff
@@ -61,6 +62,14 @@ type roundState struct {
 
 	mu             *sync.RWMutex
 	hasBadProposal func(hash common.Hash) bool
+
+	// Ignore RLP ----------------------------------------------------------------------------
```
yoomee1313: Is this roundState rlp-encoded elsewhere?

tnasu:
```
// The DecodeRLP method should read one value from the given
// Stream. It is not forbidden to read less or more, but it might
// be confusing.
func (s *roundState) DecodeRLP(stream *rlp.Stream) 
```
```
// EncodeRLP should write the RLP encoding of its receiver to w.
// If the implementation is a pointer method, it may also be
// called for nil pointers.
//
// Implementations should generate valid RLP. The data written is
// not verified at the moment, but a future version might. It is
// recommended to write only a single value but writing multiple
// values or no value at all is also permitted.
func (s *roundState) EncodeRLP(w io.Writer) error {
```

ulbqb (approved):

PR #743, opened 2026-02-09
permless: Add VRank module

ian0371:
## Proposed changes

This PR adds initial VRank module according to [KIP-227](https://github.com/kaiachain/kips/pull/27) which consists of message handlers.

The ideal process of VRank message processing:

```
proposer@N -> candidates@N: broadcast VRankPreprepare (`HandleIstanbulPreprepare()`)
candidates@N -> validators@N+1: broadcast VRankCandidate (`HandleVRankPreprepare()`)
validators@N+1: collect VRankCandidate messages (`HandleVRankCandidate()`)
proposer@N+1: fill `header.VRank` based on messages (`GetCfReport()`)
```

However, the implementation uses `validators@N` because `validators@N+1` is not finalized during `consensus@N`.

Note that this PR implements only module which does not interact with other objects; it does not yet handle actual Istanbul messages, send VRank messages, or fill `header.VRank` field.

Nil checks must be done by caller.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/kaiax/vrank/impl/handler.go
+++ b/kaiax/vrank/impl/handler.go
@@ -34,15 +34,21 @@ func (v *VRankModule) HandleIstanbulPreprepare(block *types.Block, view *istanbu
 		return
 	}
 
+	prepreparedAt := time.Now()
 	blockNum := block.NumberU64()
 	// if I'm a validator for the next block, then I need to collect VRankCandidate
 	if v.isValidator(blockNum + 1) {
-		v.startNewCollection(block, view)
+		v.prepreparedView = *view
+		v.collector.AddPrepreparedTime(vrank.ViewKey{N: blockNum, R: uint8(view.Round.Uint64())}, prepreparedAt)
 	}
 	// if I'm the proposer that broadcasted IstsanbulPreprepare to other validators,
 	// then I need to broadcast VRankPreprepare as well
 	if v.isProposer(blockNum, view.Round.Uint64()) {
 		v.BroadcastVRankPreprepare(&vrank.VRankPreprepare{Block: block, View: view})
+	}
+
+	if blockNum > maxCollectorWindow {
+		v.collector.RemoveOldViews(vrank.ViewKey{N: blockNum - maxCollectorWindow, R: maxRound})
 	}
 }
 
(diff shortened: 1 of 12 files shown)


kaiax/vrank/impl/handler.go
```diff
@@ -0,0 +1,231 @@
...
+			Round:       uint8(view.Round.Uint64()),
+			BlockHash:   block.Hash(),
+			Sig:         sig,
+		})
+	}
+	return nil
+}
+
+// HandleVRankCandidate collects VRankCandidate from candidates
+func (v *VRankModule) HandleVRankCandidate(msg *vrank.VRankCandidate) error {
```
hyeonLewis: Is it correct that we currently do not consider the case where the `VrankCandidate` message arrives before the `startNewCollection`?

kaiax/vrank/impl/handler.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	if msg == nil {
+		logger.Error("Unexpected nil")
+		return vrank.ErrVRankCandidateNil
+	}
+
+	elapsed := time.Since(v.prepreparedTime)
+	// if I'm a validator for the next block, then I need to collect VRankCandidate
+	if v.prepreparedView.Sequence != nil && v.isValidator(v.prepreparedView.Sequence.Uint64()+1) {
```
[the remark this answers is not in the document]
hyeonLewis: So do you mean that if `Proposer@N+1` comes from `ValPaused@N`, then make `Header@N+1.VRank = nil`?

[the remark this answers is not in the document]
hyeonLewis: Understood, thanks.

kaiax/vrank/impl/handler.go
hyeonLewis: What happens if `v.prepreparedView` hasn't been updated since a validator didn't receive a `ppp` but received a `VRankCandidate` message from the candidate? I think it will be discarded anyway. Is it intended behavior?

ian0371: That's regarded as future messages and collected into collector (e.g., "future block number" from `handler_test.go`). Why do you think it'll be discarded?

hyeonLewis: First, I think the `wantErr` isn't matched to what we want to test. (e.g., future block returns `ErrMsgFromNonCandidate`, please correct me if I'm wrong)

Let's assume a validator becomes `VA` at block `N`, but `v.prepreparedView` is still `N-1` because it hasn't yet received a new `ppp`. In this case, `v.isValidator(N-1)` will be `false`, so CandMsg won't be collected.

ian0371: In that case, I think `header@N.VRank` should be nil. Only validators in `VA` at `N-1` can start the timer because validators in other states will not receive IstanbulPreprepare message.

If we want `header@N.VRank` to be non-nil, then validators should broadcast IstanbulPreprepare messages to all `VA`, `CT`, `VP`, `VI`, which would require changes to Istanbul core.

ian0371: fyi, this relates to "Problem 2" mentioned in other thread

hyeonLewis: I got it, thanks. I think we only need to update the test then.

ian0371: Please check out new tests c11d9c7

hyeonLewis: Thanks. Actually what I mentioned above is [this](https://github.com/kaiachain/kaia/pull/743/changes#diff-b50f5bfcd2db3e1adb58a3ebd54bea0072dd342d1c53d131d6ccef3dc68bac1cR426-R435). Is it intended behavior to return `ErrMsgFromNonCandidate` for those cases?

ian0371: The test cases are `(block2 hash, block1 sig)` and `(block1 hash, block2 sig)`, so they result in incorrect ecrecover address. So we should expect the `ErrMsgFromNonCandidate` error.

hyeonLewis (approved):

kaiax/vrank/impl/handler.go
```diff
@@ -0,0 +1,252 @@
...
+		return common.Address{}, vrank.ErrMsgFromNonCandidate
+	}
+	return sender, nil
+}
+
+// BroadcastVRankPreprepare is called by the proposer
+func (v *VRankModule) BroadcastVRankPreprepare(vrankPreprepare *vrank.VRankPreprepare) {
+	block := vrankPreprepare.Block
+	candidates, err := v.Valset.GetCandidates(block.NumberU64())
+	if err != nil || candidates == nil {
```
hyunsooda, line 133: When no candidate exists, a `nil` return is expected, but the log interprets this as an error. All uses of `GetCandidates` follow the same pattern. Behaviorally, the logic seems faultless, but the error log might look confusing.

ian0371: @hyunsooda  If none exists, I think `GetCandidates` should return empty array instead of nil.
Or are you suggesting that users should always check separately?

```
candidates, err := v.Valset.GetCandidates(block.NumberU64())
if err != nil {
		logger.Error("Sign VRankPreprepare failed", "blockNum", block.NumberU64())
		return
}
if candidates == nil {
		return
}
```

hyunsooda: Yes, I think so. Is there any issue if `GetCandidates()` returns `nil` when no candidates are found? My understanding was that returning `nil` is a common and safe practice with primitive behaviors like `for range` and `slices.XX()`. Could you explain why an empty array is preferred over `nil` in your scenario?

ian0371: I checked other getters in this module — `GetCouncil` (getter.go:30) and `GetDemotedValidators` (getter.go:52) both return via `List()` which always produces a non-nil slice (via `make([]common.Address, len(as.list))` + copy in address_set.go:110-111), even when empty.
There are a few other cases such as `GetAllowed` / `GetThrottled` (blockchain/spam_throttler.go:272,283), `getGaslessInfo` (kaiax/gasless/impl/getter.go:377) which prefers empty slice over nil.
So I think it's a convention.

As you said, `candidates == nil` check might be unnecessary, but personally I would like to avoid dealing with nil as much as possible because of fear of nil dereference.

hyunsooda: Ah okay. `GetCandidates` returns an empty slice if there's no error.

PR #748, opened 2026-02-12
permless: Implemented KIP286 and KIP287

hyunsooda:
## Proposed changes

Implemented KIP-286 (Permissionless Validator) and KIP-287 (State Transition).

> **⚠️ This PR is a working version for reference only. It will be split into 4 smaller PRs for review.**
> See [`permissionless/PR_SPLIT_PLAN.md`](https://github.com/hyunsooda/kaia/blob/state-transition/permissionless/PR_SPLIT_PLAN.md) for the split plan.

### PR Split Plan

| # | Scope | Key Files | Review Focus |
|---|-------|-----------|--------------|
| PR 1 | Solidity contracts (OZ v5 migration + new contracts) | `contracts/` | OZ v4→v5, ABv2 contract logic, new interfaces |
| PR 2 | ABv2 Go layer (read/write/install) | `blockchain/system/` | ABI encoding, contract read helpers, genesis alloc |
| PR 3 | Permissionless state transition (core logic) | `kaiax/valset/`, `kaiax/staking/` | State machine, epoch/timeout/violation transitions |
| PR 4 | Integration (config, API, misc) | `params/`, `node/cn/`, `cmd/` | Config wiring, Staking, API, CLI  |

Dependency: PR 1 → PR 2 → PR 3 → PR 4

### Contract (OZ v5 Migration)
- Migrate OpenZeppelin from v4 to v5 (upgradeable contracts)
- Update ABv2 contract: ABv2DataContract accepts impl address, new NodeVerifier
- Add CnStakingV4, CnStakingV4Factory, PublicDelegation, Proxy contracts
- Remove legacy proxy.sol/proxy.go and ICnStaking.sol
- Update all Go bindings

### Core Implementation
- **ABv2**: Read/write functions (ReadGetAllValidators, EncodeWriteNodes, InstallAddressBookV2, ReadAddressBookV2BlsAll)
- **Genesis**: AllocPermissionless for HF=0 permissionless genesis
- **State Transition**: Epoch/timeout/violation transitions with cache-based node state management
- **Staking**: Permissionless staking info parsing via multicall (KEF/KIF from ABv2)
- **Council**: Permissionless council filtering (ValActive/ValPaused/ValReady)

### Unit Tests
- State transition: epoch, timeout, violation, applyAllTransitions (table-driven)
- Genesis: AllocPermissionless (normal, single validator, error cases)
- ABv2 activation: InstallAndInitializeABv2, ReadGetAllValidators, EncodeWriteNodes
- ABv2 reading: ReadAddressBookV2BlsAll, GetNodeByState
- Council: GetCouncilPermissionless, GetPermlessCouncil
- Voting: PermissionlessIgnoresVote
- Multicall: MultiCallStakingInfoPermissionless

### Bug Fixes
- NodeStateMap.Addresses() now returns deterministically sorted addresses

## Types of changes

- [x] 🧪 Test improvements

## Checklist

- [x] 🟢 Lint and unit tests pass locally with my changes

## Related issues

- KIP-286: Permissionless Validator
- KIP-287: State Transition
- [system-contracts#139](https://github.com/kaiachain/system-contracts/issues/139): EPOCH_BLOCK_INTERVAL is a constant

## Further comments

See [`permissionless/TEST_PLAN.md`](https://github.com/hyunsooda/kaia/blob/state-transition/permissionless/TEST_PLAN.md) for full test coverage details, [`permissionless/TODO.md`](https://github.com/hyunsooda/kaia/blob/state-transition/permissionless/TODO.md) for remaining work items, and [`permissionless/PR_SPLIT_PLAN.md`](https://github.com/hyunsooda/kaia/blob/state-transition/permissionless/PR_SPLIT_PLAN.md) for the detailed split plan.


kaiax/valset/interface.go
```diff
@@ -32,4 +35,10 @@ type ValsetModule interface {
 	GetCommittee(num uint64, round uint64) ([]common.Address, error)
 	GetDemotedValidators(num uint64) ([]common.Address, error)
 	GetProposer(num uint64, round uint64) (common.Address, error)
+
```
ian0371, line 48: Could you implement `GetCandidates(num uint64) ([]common.Address, error)` which is required for vrank module ([link](https://github.com/kaiachain/kaia/pull/743/changes#diff-21293fdf1268c75a1c49f08454d3126e629a0e35501ace4d08a4188ad7f05c2dR34))?

kaiax/valset/impl/getter.go
```diff
@@ -86,3 +87,23 @@ func (v *ValsetModule) GetProposer(num, round uint64) (common.Address, error) {
 	}
 	return v.getProposer(c, round)
 }
+
+func (v *ValsetModule) EpochTransition(validators valset.ValidatorChartMap, num uint64, state *state.StateDB) (valset.ValidatorChartMap, error) {
```
ian0371: nit: It seems this function updates internal variables, so I don't think it belongs to `getter.go` which is supposed to contain read-only getters.

kaiax/valset/impl/getter_demote.go
```diff
@@ -49,38 +50,47 @@ func (v *ValsetModule) getDemotedValidators(council *valset.AddressSet, num uint
...
+		if demoted.Len() == len(council.List()) {
+			demoted = valset.NewAddressSet(nil)
+		}
 
-	// Under single governance mode, governing node cannot be demoted.
-	if singleMode && demoted.Contains(governingNode) {
-		demoted.Remove(governingNode)
+		// Under single governance mode, governing node cannot be demoted.
+		if singleMode && demoted.Contains(governingNode) {
+			demoted.Remove(governingNode)
```
ian0371, line 70: I think this policy should remain after permissionless. HeaderGov will still be used for non-validator votings.
Or are you thinking of a different way to forcefully inject validator set (just like we discussed the forceful removal by super user from contract-level)?

kaiax/valset/impl/getter.go
ian0371: Maybe we'll need to add ExecutionModule.PreInsertBlock()

kaiax/valset/impl/getter_council.go
```diff
@@ -53,28 +54,59 @@ func (v *ValsetModule) getCouncilGenesis() (*valset.AddressSet, error) {
...
 
-func (v *ValsetModule) getCouncilDB(num uint64) (*valset.AddressSet, bool, error) {
-	pMinVoteNum := v.readLowestScannedVoteNumCached()
-	if pMinVoteNum == nil {
-		return nil, false, errNoLowestScannedNum
-	}
-	nums := v.readValidatorVoteBlockNumsCached()
-	if nums == nil {
-		return nil, false, errNoVoteBlockNums
+func (v *ValsetModule) getCouncilDB(num uint64) (*ValidatorList, bool, error) {
```
ian0371: We can use the same pattern as in `state_transition.go`:

```
func (v *ValsetModule) getCouncilDB(num uint64) (*ValidatorList, bool, error) {
  if prepermless {
    return getCouncilDBPrePermless(num)
  }
}

func (v *ValsetModule) getCouncilDBPrePermless(num uint64) (*valset.AddressSet, bool, error) {}
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+			}
+		case valset.ValReady, valset.ValActive, valset.ValPaused:
+			if val.StakingAmount >= minStake {
+				potentialActiveValSet = append(potentialActiveValSet, val)
+			} else {
+				val.State = valset.ValInactive
+			}
+		}
+	}
+	sort.Slice(potentialActiveValSet, func(i, j int) bool {
```
ian0371: nit: https://pkg.go.dev/sort#Slice mentions [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) is often faster

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+		return potentialActiveValSet[i].StakingAmount > potentialActiveValSet[j].StakingAmount
```
ian0371: Tie breaker to enforce strict total ordering (as in KIP-286):
```
func(i, j int) bool {
	if potentialActiveValSet[i].StakingAmount == potentialActiveValSet[j].StakingAmount {
	    return address comparison
	}
	return potentialActiveValSet[i].StakingAmount > potentialActiveValSet[j].StakingAmount
}
```

kaiax/valset/types.go
```diff
@@ -0,0 +1,164 @@
...
+	return nil
+}
+
+func (s State) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%q", s.String())), nil
+}
+
+type ValidatorChart struct {
+	State         State     `json:"state"`
+	StakingAmount float64   `json:"stakingAmount"` // in KAIA unit
```
ian0371: Why float64, instead of uint64?

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+func (v *ValsetModule) timeoutTransition(validators valset.ValidatorChartMap) valset.ValidatorChartMap {
+	newValidators := validators.Copy()
+	for _, val := range newValidators {
+		switch val.State {
+		case valset.ValReady, valset.ValInactive:
+			if time.Now().After(val.IdleTimeout) {
+				val.State = valset.CandInactive
+			}
+		case valset.ValPaused:
+			if time.Now().After(val.PausedTimeout) {
```
ian0371: In this case, is the default value uint256max (in contract)?
```
// new validator
val.PausedTimeout = uint256max

// on pause()
val.PausedTimeout = block.timestamp + ValPausedTimeout

// on unpause()
val.PausedTimeout = uint256max
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+// - [O] ValActive -> ValActive
+func (v *ValsetModule) epochTransition(si *staking.StakingInfo, num uint64, validators valset.ValidatorChartMap) valset.ValidatorChartMap {
+	defer func() {
+		for addr, val := range validators {
+			logger.Info("TODO-Permissionless: Remove this log", "num", num, "addr", addr.String(), "state", val.State.String(), "stakingamount", val.StakingAmount, "idleTimeout", val.IdleTimeout.String(), "pausedtimeout", val.PausedTimeout.String())
+		}
+	}()
+	if !isVrankEpoch(num) {
+		// return early if the `num` is not vrank epoch number
+		return validators
```
ian0371:
```suggestion
		return validators.Copy()
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+				val.State = valset.CandInactive
+			}
+		case valset.ValReady, valset.ValActive, valset.ValPaused:
+			if val.StakingAmount >= minStake {
+				potentialActiveValSet = append(potentialActiveValSet, val)
+			} else {
+				val.State = valset.ValInactive
+			}
+		}
+	}
```
ian0371: nit: Let's mark each transition from KIP

```suggestion
	for _, val := range newValidators {
		switch val.State {
		case valset.ValExiting:
			val.State = valset.ValInactive // T1
		case valset.CandReady:
			if val.StakingAmount >= minStake {
				val.State = valset.CandTesting // T4a
			} else {
				val.State = valset.CandInactive // T4b
			}
		case valset.CandTesting:
			if v.isPassVrankTest() {
				if val.StakingAmount >= minStake {
					potentialActiveValSet = append(potentialActiveValSet, val) // T3a
				} else {
					val.State = valset.ValInactive // T3b
				}
			} else {
				val.State = valset.CandInactive // T2
			}
		case valset.ValReady, valset.ValActive, valset.ValPaused:
			if val.StakingAmount >= minStake {
				potentialActiveValSet = append(potentialActiveValSet, val) // T3a
			} else {
				val.State = valset.ValInactive // T3b
			}
		}
	}
```

kaiax/valset/impl/getter_council.go
```diff
@@ -53,28 +54,59 @@ func (v *ValsetModule) getCouncilGenesis() (*valset.AddressSet, error) {
...
+		readNum = voteNum
+	)
+	if isPermissionless {
+		curNum := v.Chain.CurrentBlock().Number().Uint64()
+		if stateChangeNum > curNum {
+			stateChangeNum = curNum
+		}
+		readNum = stateChangeNum
 	}
+	council = ReadCouncil(v.ChainKv, readNum)
```
ian0371: I think candidates are included in the council. Shouldn't they be removed?

kaiax/valset/impl/getter_council.go
```diff
@@ -53,28 +54,59 @@ func (v *ValsetModule) getCouncilGenesis() (*valset.AddressSet, error) {
...
-		council := valset.NewAddressSet(ReadCouncil(v.ChainKv, voteNum))
-		return council, true, nil
+	var (
+		council *ValidatorList
+		readNum = voteNum
+	)
+	if isPermissionless {
+		curNum := v.Chain.CurrentBlock().Number().Uint64()
+		if stateChangeNum > curNum {
+			stateChangeNum = curNum
```
ian0371: When does this happen `stateChangeNum > curNum`?
Is it ok to read DB from `curNum` when it is not a checkpoint?

kaiax/valset/impl/init.go
```diff
@@ -101,16 +104,24 @@ func (v *ValsetModule) Init(opts *InitOpts) error {
 
 func (v *ValsetModule) initSchema() error {
 	// Ensure mandatory schema at block 0
-	if voteBlockNums := ReadValidatorVoteBlockNums(v.ChainKv); voteBlockNums == nil {
-		writeValidatorVoteBlockNums(v.ChainKv, []uint64{0})
-		v.validatorVoteBlockNumsCache = nil
+	curBlockNum := v.Chain.CurrentBlock().Number()
+	if v.Chain.Config().Rules(curBlockNum).IsPermissionless {
+		if stateChangeBlockNums := ReadValidatorStateChangeBlockNums(v.ChainKv); stateChangeBlockNums == nil {
+			writeValidatorStateChangeBlockNums(v.ChainKv, []uint64{0})
```
ian0371: This will not be executed when the node is launched before permissionless hardfork block.
I think hardfork check here needs to be removed.

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,468 @@
...
+	ValPausedTimeout     = time.Hour * 8
+	ActiveValidatorCount = 50
+)
+
+type ValidatorList struct {
+	permlessMu       sync.RWMutex
+	isLegacy         bool
+	permissionedVals *valset.AddressSet
+	permlessVals     valset.ValidatorChartMap
+}
```
ian0371: nit: IMO instead of having a wrapper for the legacy `permissionedVals`, I prefer having a new type and functions for everything, and leave the legacy functions mostly as-is.

For example,
```
type ValidatorList struct {
	permlessMu       sync.RWMutex
	permlessVals     valset.ValidatorChartMap
}

type AddrSet interface {
    Add()
    Remove()
    List()
    ...
}

func (v *ValsetModule) getCouncilPermissioned(num uint64) (AddrSet, error) {
  return &valset.AddressSet{...}
}
func (v *ValsetModule) getCouncil(num uint64) (AddrSet, error) {
  return &ValidatorList{...}
}
```
and enforce this pattern everywhere:

```
func (v *ValsetModule) GetCouncil(num uint64) ([]common.Address, error) {
  if num is permissionless { return getCouncil(num).List() }
  else { return getCouncilPermissioned(num).List() }
}
```

wdyt?

kaiax/staking/impl/getter.go
```diff
@@ -96,10 +130,26 @@ func (s *StakingModule) getFromStateByNumber(num uint64) (*staking.StakingInfo,
 	return s.getFromState(header, statedb)
 }
 
+func (s *StakingModule) getFromStateByNumberFromState(num uint64, statedb *state.StateDB) (*staking.StakingInfo, error) {
```
hyeonLewis: I think there are too many `getFromState*` functions internally, which is confusing. How about merging those functions? Can we only have getFromState?

contracts/contracts/system_contracts/multicall/MultiCallContract.sol
```diff
@@ -90,18 +105,36 @@ contract MultiCallContract {
         stakingAmounts = new uint256[](lenCnAddress / 3);
 
         for (uint256 i = 0; i < lenCnAddress; i += 3) {
-            stakingAmounts[i / 3] = _getCnStakingAmounts(addressList[i + 1]);
+            stakingAmounts[i / 3] = _getCnStakingAmounts(kip290Enabled, ignoreAbookVersion, addressList[i + 1]);
```
hyeonLewis: This will require extra check `if (kip290Enabled && (ignoreAbookVersion || IAddressBook(ADDRESS_BOOK_ADDRESS).VERSION() >= 2))` for every loop. I think we can divide the functions into `_multiCallStakingInfoLegacy` and `_multiCallStakingInfoKIP290`.

contracts/contracts/system_contracts/vrank/ValidatorState.sol
```diff
@@ -0,0 +1,167 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the kaia library. If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.25;
+
+import {EnumerableSet} from "openzeppelin-contracts-5.0/utils/structs/EnumerableSet.sol";
+import {IValidatorState, IAddressBook, ICnStaking} from "./IValidatorState.sol";
+
+contract ValidatorState is IValidatorState {
```
hyeonLewis: AFAIK, we decided to use MiscDB for it, but why do we use contract storage?

contracts/contracts/system_contracts/multicall/MultiCallContract.sol
```diff
@@ -65,14 +67,27 @@ contract MultiCallContract {
     /* ========== STAKING INFORMATION ========== */
 
     // multiCallStakingInfo returns the staking information of all CNs.
-    function multiCallStakingInfo()
+    function multiCallStakingInfo(bool kip290Enabled, bool ignoreAbookVersion)
```
hyeonLewis: Is `ignoreAbookVersion` for testing?

kaiax/valset/interface.go
hyunsooda: Implemented at https://github.com/kaiachain/kaia/pull/748/commits/475d1eed7c427c009d87fa7001cf5ae9dd9c7411

kaiax/valset/impl/getter.go
hyunsooda: Could you pinpoint which side effect does exist in that function? If you mean the `StakingAmount`, then it's updated in the previous step now at https://github.com/kaiachain/kaia/pull/748/commits/475d1eed7c427c009d87fa7001cf5ae9dd9c7411.

kaiax/valset/impl/getter_demote.go
hyunsooda: Ah, yes, it should still apply even in a permissionless operation.

Regarding `governance.addvalidator` and `governance.removevalidator`, my understanding is that the former (`addvalidator`) would be entirely removed, effectively making `ValidatorList.Add()` a no-op. For the latter (`removevalidator`), I believe the validator would be forcefully transitioned to the `CandInactive` state. If not, we can modify it later, but I installed its transition handler.

I've added a voting transition handler before the VRank violation handler. We now have a total of four transition handlers: anytime, vote, VRank violation, and epoch.

kaiax/valset/types.go
hyunsooda: Uhm, I can't remember why I chose that type, but `uint64` looks fit. Thank you (https://github.com/kaiachain/kaia/pull/748/changes/475d1eed7c427c009d87fa7001cf5ae9dd9c7411)

kaiax/valset/impl/state_transition.go
hyunsooda: I intended like below.

// new validator
val.PausedTimeout = 0

// on pause()
val.PausedTimeout = block.timestamp + ValPausedTimeout

// on unpause
no-op (timeout value is not explicitly zeroized)

If the validator is on pause, new timer sets. If it do unpause(), it goes to `ValActive` state without explicit zeroization for this mock contract. Do you think we need to make it zero explicitly?

kaiax/valset/impl/state_transition.go
hyunsooda: Thanks: https://github.com/kaiachain/kaia/pull/748/changes/475d1eed7c427c009d87fa7001cf5ae9dd9c7411

kaiax/valset/impl/getter_council.go
hyunsooda: That's a really good point to discuss. I also wanted to discuss how we structure each entity. Initially, I thought the council would include all entities, and then each entity would be filtered by functions such as `getDemoteValidators` or `getCommittee()`. If we want to return a candidate set, we could also create a filter function for it, which might be required for VRank and RPC/Console API.

kaiax/valset/impl/getter_council.go
hyunsooda: The condition hits when `engine.Prepare()` is called. Inside of this, it calls `sb.GetValidatorSet()`. The number is the new header's one, but the global current block number is -1.

This condition is unnecessary now https://github.com/kaiachain/kaia/pull/748/changes/475d1eed7c427c009d87fa7001cf5ae9dd9c7411. The `readNum` is fetched from MiscDB and points the latest recording number.

kaiax/valset/impl/init.go
hyunsooda: Could you elaborate on this more? If the permissionless HF is not yet available, it falls back to the legacy code in the `else` branch.

kaiax/valset/impl/state_transition.go
hyunsooda: Applied. But the T2b doesn't seem to align exactly with the actual behavior. How do you think?

kaiax/valset/impl/init.go
ian0371: CNs must be upgraded before hardfork block.

CN upgrade to v3 -> node start -> Init() called -> initSchema() called -> curBlockNum is not permissionless -> new schema is not initialized
Later, Permissionless HF block passes. At this moment, new schema is never initialized, unless we restart the node again.

So the logic should be
```
writeValidatorStateChangeBlockNums (if not exists) // always call it
if cur is not permissionless:
    writeValidatorVoteBlockNums (if not exists)
```

kaiax/valset/impl/state_transition.go
ian0371: Oh, I missed `case valset.ValPaused:` condition. nvm.

kaiax/valset/impl/state_transition.go
ian0371: Do you mean T3b? I think it's ok because it's below top50 by less than minstake.

kaiax/valset/impl/init.go
hyunsooda: Ahh okay your're right. I saw it's incorrectness and verified the effectiveness of the change here  https://github.com/kaiachain/kaia/pull/748/changes/afa1109a13f71fd7e8befa90193a25c84d60568b

kaiax/valset/impl/state_transition.go
hyunsooda: Ah sorry, I pointed T2, not T2b.

> T2 (CandTesting evaluation): Evaluate VRank before calculating top 50 since passed candidates become eligible for the active set.

So, it seems like align with the `v.isPassVrankTest()` and actual transition is decided on the further condition like `staking amount`. The definition says only it's behavior and, not input/output. So, a little bit confused to adaopt that comment

kaiax/valset/impl/state_transition.go
hyunsooda: Thanks. https://github.com/kaiachain/kaia/pull/748/changes/4a9b00d26d7ba3c6c4503a89748929ff6fac1214

kaiax/valset/impl/state_transition.go
hyunsooda: Thanks. https://github.com/kaiachain/kaia/pull/748/changes/4a9b00d26d7ba3c6c4503a89748929ff6fac1214

kaiax/valset/impl/state_transition.go
hyunsooda: Your suggestion looks better than the current one. I've refactored them out. Could you please take a look? https://github.com/kaiachain/kaia/pull/748/changes/0a1476317d1c2d676b2eae056019f7496f43e90b

kaiax/valset/impl/state_transition.go
ian0371: Oh, I was looking at the graph when adding the comments. The comments make it easier to check all dotted lines in the graph.

image

kaiax/valset/impl/getter.go
ian0371: Yes I meant `updateStakingAmount`

kaiax/valset/impl/getter_council.go
hyunsooda: Formal initialization process should be established to transit smoothly from pemissioned to permissionless network.

kaiax/staking/impl/getter.go
hyunsooda: There are three `getFromState*` functions
- `getFromStateByNumber()`: the previously existing function, just renamed.
- `getFromStateByNumberFromState()`: newly added function. It uses the state from the param. The difference between `getFromStateByNumber()` and `getFromStateByNumberFromState()` is the source of statedb.
- `getFromState()`: call multicall contract

So, the new added function is a distinguished function to avoid using the state from `Chain.StateAt()` as   `getFromStateByNumber()` does.

Q: Why we need this new function
A: Sometimes the node generates a historical state, and it calls `engine. Initialize ()` at every block, but at this moment, the call of `Chain.StateAt()` returns empty. The progress state is directly passed into the `engine. Initialize ()` function. It makes adding a new getter with desired state.

hyunsooda:
This PR is a reference for the split PRs (PR1~PR4) and will not be merged directly. The conflict with the permissionless base branch is expected and can be ignored.

hyunsooda:
Close via
- https://github.com/kaiachain/kaia/pull/792
- https://github.com/kaiachain/kaia/pull/796
- https://github.com/kaiachain/kaia/pull/797
- https://github.com/kaiachain/kaia/pull/802
- https://github.com/kaiachain/kaia/pull/803
- https://github.com/kaiachain/kaia/pull/810
- https://github.com/kaiachain/kaia/pull/811
- https://github.com/kaiachain/kaia/pull/812

PR #765, opened 2026-02-24
permless/bind: Add backend with owned state

hyunsooda:
## Proposed changes

- Added `StateBlockchainContractCaller` caller, which implements `ContractCaller` with owned-state.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Related issues

https://github.com/kaiachain/kaia/pull/748


accounts/abi/bind/backends/blockchain_state.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/kaiachain/kaia/node/cn/filters"
+)
+
+// `StateBlockchainContractBackend` has the same internal behavior with BlockchainContractBackend
+// execpt for the `ContractCaller` interface functions i.e., `CodeAt()` and `CallContract()`
+type StateBlockchainContractBackend struct {
+	bc              BlockChainForCaller
+	state           *state.StateDB
+	bcCommonBackend *BlockchainContractCommonBackend
+}
```
ian0371, line 40: Can we simplify this with struct embedding (inheritance-like composition in Go)?

All three structures newly introduced appear to share most behavior, with only a few state-related differences. We could centralize shared logic and override only the methods that need different state resolution.

```go
// Base-like type (shared behavior)
type BlockchainContractBackendWithState struct {
	bc     BlockChainForCaller
	txPool TxPoolForCaller
	events *filters.EventSystem
	state  *state.StateDB
}

// State-independent behavior stays shared
func (b *BlockchainContractBackendWithState) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
	...
}

// State-dependent behavior uses b.state
func (b *BlockchainContractBackendWithState) CallContract(
	ctx context.Context,
	block *types.Block,
	call kaia.CallMsg,
	blockNumber *big.Int,
) ([]byte, error) {
	res, err := b.callContract(call, block, b.state)
	...
}
```

Then the chain-backed variant can embed this and override only state-related entry points:

```go
type BlockchainContractBackend struct {
	BlockchainContractBackendWithState
}

// no need to re-implement SuggestGasPrice, etc.

func (b *BlockchainContractBackend) CallContract(ctx context.Context, call kaia.CallMsg, blockNumber *big.Int) ([]byte, error) {
	block, state, err := b.getBlockAndState(blockNumber)
	if err != nil {
		return nil, err
	}
	b.state = state
	return b.BlockchainContractBackendWithState.CallContract(ctx, block, call, blockNumber)
}
```

hyunsooda: Could you take a look https://github.com/kaiachain/kaia/pull/765/changes/e6a7c813f331e855baa3dbbd05b88d494e30700a?

accounts/abi/bind/backends/blockchain_common.go
```diff
@@ -0,0 +1,297 @@
+// Modifications Copyright 2026Kaia Authors
```
hyeonLewis:
```suggestion
// Modifications Copyright 2026 The Kaia Authors
```

accounts/abi/bind/backends/blockchain_state.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/kaiachain/kaia"
+	"github.com/kaiachain/kaia/accounts/abi/bind"
+	"github.com/kaiachain/kaia/blockchain/state"
+	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/node/cn/filters"
+)
+
+// `StateBlockchainContractBackend` has the same internal behavior with BlockchainContractBackend
+// execpt for the `ContractCaller` interface functions i.e., `CodeAt()` and `CallContract()`
```
hyeonLewis:
```suggestion
// except for the `ContractCaller` interface functions i.e., `CodeAt()` and `CallContract()`
```

accounts/abi/bind/backends/blockchain_state.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/kaiachain/kaia"
+	"github.com/kaiachain/kaia/accounts/abi/bind"
+	"github.com/kaiachain/kaia/blockchain/state"
+	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/node/cn/filters"
+)
+
+// `StateBlockchainContractBackend` has the same internal behavior with BlockchainContractBackend
```
hyeonLewis: The `BalanceAt` has a different function signature and was dropped (no wrapper) in [e6a7c81](https://github.com/kaiachain/kaia/pull/765/commits/e6a7c813f331e855baa3dbbd05b88d494e30700a). Is it intentional since we don't uuse `StateBlockchainContractBackend` in `sc.Backend`? If so I think we should update this comment.

hyunsooda: Yes, the newly added `StateBlockchainContractBackend` is not used in ServiceChain.
That special comment is kept placed here: https://github.com/hyunsooda/kaia/blob/e6a7c813f331e855baa3dbbd05b88d494e30700a/accounts/abi/bind/backends/blockchain.go#L123

hyunsooda (commented):
Ah, I'm sorry, my comment was pending state.

hyeonLewis (approved):

accounts/abi/bind/backends/blockchain_common.go
```diff
@@ -0,0 +1,297 @@
...
+
+// Executes a read-only function call with respect to the specified block's state, or latest state if not specified.
+//
+// Returns call result in []byte.
+// Returns error when:
+// - cannot find the corresponding block or stateDB
+// - VM revert error
+// - VM other errors (e.g. NotProgramAccount, OutOfGas)
+// - Error outside VM
+func (b *BlockchainContractCommonBackend) CallContract(ctx context.Context, state *state.StateDB, block *types.Block, call kaia.CallMsg, blockNumber *big.Int) ([]byte, error) {
```
ian0371: `BlockchainContractCommonBackend` does not implement `ContractCaller`, right?
Then can we remove ctx and blockNumber?

accounts/abi/bind/backends/blockchain_common.go
```diff
@@ -0,0 +1,297 @@
...
+	bc     BlockChainForCaller
+	txPool TxPoolForCaller
+	events *filters.EventSystem
+}
+
+func newBlockchainContractCommonBackend(bc BlockChainForCaller, tp TxPoolForCaller, es *filters.EventSystem) *BlockchainContractCommonBackend {
+	return &BlockchainContractCommonBackend{bc, tp, es}
+}
+
+func (b *BlockchainContractCommonBackend) CodeAt(ctx context.Context, state *state.StateDB, account common.Address, blockNumber *big.Int) ([]byte, error) {
```
ian0371: `BlockchainContractCommonBackend` does not implement `ContractCaller`, right?
Then can we remove ctx and blockNumber?

accounts/abi/bind/backends/blockchain_common.go
```diff
@@ -0,0 +1,297 @@
...
+	// But our sender (usually 0x0) won't have enough balance. Instead we override gasPrice = 0 here
+	txContext.GasPrice = big.NewInt(0)
+	evm := vm.NewEVM(blockContext, txContext, state, b.bc.Config(), &vm.Config{})
+
+	return blockchain.ApplyMessage(evm, msg)
+}
+
+// bind.ContractTransactor defined methods
+
+func (b *BlockchainContractCommonBackend) PendingCodeAt(ctx context.Context, account common.Address) ([]byte, error) {
```
ian0371: Why is `statedb` missing here?
Is it because of `ContractTransactor`? If so, does `BlockchainContractCommonBackend` need to implement `ContractTransactor`?

accounts/abi/bind/backends/blockchain_state.go
```diff
@@ -0,0 +1,82 @@
...
+	if block == nil {
+		return nil, errBlockDoesNotExist
+	}
+	return block, nil
+}
+
+// bind.ContractCaller defined methods
+
+func (b *StateBlockchainContractBackend) CodeAt(ctx context.Context, account common.Address, blockNumber *big.Int) ([]byte, error) {
+	return b.BlockchainContractCommonBackend.CodeAt(ctx, b.state, account, blockNumber)
```
ian0371: What do you think about returning error when `blockNumber` is set?
blockNumber is a dummy argument here.

```
if blockNumber != nil {
    return nil, err
}
```

accounts/abi/bind/backends/blockchain_state.go
ian0371: Do you think CommonBackend is still needed despite the aforementioned design which requires only two structs?

accounts/abi/bind/backends/blockchain_common.go
hyunsooda: There was no `statedb` in the parameter in this function. Could you elaborate where `statedb` should be placed or fetched from? I couldn't get your context. Could you please elaborate on this more?

accounts/abi/bind/backends/blockchain_common.go
hyunsooda: Let me split this question into two parts.
- unused `ctx` param: The `ctx` param had been unused in the previous codebase. I think it would be better to handle the refactoring in a subsequent PR. Since it involves changes to several files, it might be cleaner to keep it separate from this scope. What do you think?
- `BlockchainContractCommonBackend` provides a default `ContractCaller` implementation and `*BlockchainContractBackend` calls it via embedding.

accounts/abi/bind/backends/blockchain_state.go
hyunsooda: Do you meant that can this be inlined directly into the `BlockchainContractBackend` methods? If so,  we can delete `CommonBackend`, but `StateBlockchainContractBackend` also should have that inlined functions. I understood that the https://github.com/kaiachain/kaia/commit/e6a7c813f331e855baa3dbbd05b88d494e30700a commit handled your suggestion, but I feel now the further comment seems to revert to the previous commit if I correctly understood.

accounts/abi/bind/backends/blockchain_state.go
hyunsooda: The `blockNumber`  is required becuase`BlockchainContractBackend` fetches the state from block number, while the `StateBlockchainContractBackend` does not use `blockNumber` at all (the unused params can be refactored in the following PR; currently just aligned with previous signatures).
Because the owned-state is directly used, the block number must not be related with that usage. I believe that refactoring would simply remove your concern.

hyunsooda:
@hyeonLewis Refactored. The newly added `StateBlockchainContractCaller` implements only `ContractCaller` interface and nothign changed for exisiting codebase.

ian0371 (approved):

hyeonLewis (approved):

Issue #751, opened 2026-02-16
p2p Dial and Discovery upgrade


blukat29:
# Purpose

To realize the permissionless network topology, the p2p stack has to accommodate an advanced authorization features. In the process, the implementation is going to be streamlined.

# Components affected

- `p2p.Server,BaseServer,MultiChannelServer`: Will be simplified by removing the run() loop.
- `p2p.dialstate`: Will be replaced by the new `p2p.dialsched`, and simplified.
- `discover.Discovery,Table`: Will be replaced by the new `discover.table`, and simplified
- `discover.UDP`: Slightly refactored

# Change tasks

All changes are first committed to the [openbn](https://github.com/kaiachain/kaia/tree/openbn) branch.

- Refactor/modernize UDP
  #744
- Replace dialstate with dialsched
  - Extract Server methods from run() loop's event handlers
    #772
    #773
  - Add dialsched.go for DialSched.
    Server launches DialSched instead of dialstate
    Delete Server.run() and dialstate
    Reduce Server interface
    #774
- Replace Table with table (new implementation)
  - Add table_*.go, discovery.go for `Table2` and `Discovery2`
    #775 
  - Use Discovery2 interface from dialsched and bootnode APIs
    #787 
  - Replace Table with Table2, Discovery with Discovery2
    Reduce Bootnode APIs
    #791 
    #793 

# Desired changes

### Reduce Server interface

```diff
  // Manage subprotocols
  GetProtocols() []Protocol
  AddProtocols(p []Protocol)

  // Peer connection
  SetupConn(fd net.Conn, flags connFlag, dialDest *discover.Node) error
  AddPeer(node *discover.Node)
  RemovePeer(node *discover.Node)
  SubscribeEvents(ch chan *PeerEvent) event.Subscription
  Disconnect(destID discover.NodeID)
- NodeDialer // dial scheduler should have it internally

  // Dial scheduling assistance -- dial scheduler should handle it internally.
- AddLastLookup() time.Time
- SetLastLookupToNow()

  // Wrapper to discover.Table -- dial scheduler should directly call table, not through server.
- CheckNilNetworkTable() bool
- GetNodes(nType discover.NodeType, max int) []*discover.Node
- Lookup(target discover.NodeID, nType discover.NodeType) []*discover.Node
- Resolve(target discover.NodeID, nType discover.NodeType) *discover.Node

  // Lifecycle
  Start() (err error)
  Stop()

  // Getters
  PeersInfo() []*PeerInfo
[...]

PR #774, opened 2026-03-01
p2p: Replace dialstate with DialSched

blukat29:
## Proposed changes

- Replace [`Server`.run() task loop + `dialstate` task generator] with [`DialSched`.dialLoop() task loop]
  - Delete `dialstate` and its dependencies.
  - Use `DialSched` instead.
  - The run() loop's `tab.Close()` and `disconnect all peer` logic is absorbed into `srv.Stop()`.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Related issues

- #751


blukat29:
Rebased on top of latest #773

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+}
+
+func (ds *DialSched) removeStatic(id discover.NodeID) {
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	ds.static.remove(id)
+
+	// Additionally cleanup the dial speed bumps.
+	// This allows admin_removePeer -> admin_addPeer combo to force an immediate reconnect.
```
2dvorak, line 442: It looks like your intent is:
- remove by `admin_removePeer`: clear `dialBackoff` and `connFails`
- removed due to repeated connection fail: keep `dialBackoff` and `connFails`

but users might expect `admin_addPeer` to try connection immediately. Is there a reason not to allow `admin_addPeer` to reconnect immediately for the latter?

networks/p2p/dialsched_test.go
```diff
@@ -0,0 +1,697 @@
...
+	assert.True(t, ds.static.contains(staticNode.ID), "static node must not be removed by disconnection events only")
+
+	// But Failure does remove the static node.
+	for i := 0; i < dialMaxRetries+2; i++ {
+		ds.markConnFailure(staticNode.ID)
+	}
+	assert.False(t, ds.static.contains(staticNode.ID), "static node must be removed by failure events")
+}
+
+// Special feature: Removing and re-adding a static node re-dials immediately1.
```
2dvorak:
```suggestion
// Special feature: Removing and re-adding a static node re-dials immediately.
```

Minor typo..

networks/p2p/dialsched_util.go
```diff
@@ -0,0 +1,112 @@
...
+			discover.NodeTypeEN: cfg.maxDynDials,
+		}
+	default:
+		return map[discover.NodeType]int{}
+	}
+}
+
+// typedNodeSet is a set of discover.Nodes organised by NType.
+// It is not safe for concurrent use; callers must hold DialSched.mu.
+type typedNodeSet struct {
```
ian0371, line 48: Why organize with nodeType when `map[discover.NodeID]*discover.Node` seems sufficient?
For faster `count()` method?

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+
+	logger.Debug("Dialing candidates", "needs", needs, "candidates", len(candidates))
+	return candidates, needRefresh
+}
+
+func (ds *DialSched) getDynamicCandidates(targetType discover.NodeType, need int) (nodes []*discover.Node, needRefresh bool) {
+	if ds.tab == nil || need <= 0 {
+		return nil, false
+	}
+	oversample := need * 4 // Oversample in case some nodes are not reachable.
```
ian0371, line 311: Is this a new logic? Is the intention getting as many candidates as possible, and skip most of them during dial?
Which part of the existing code does this correspond to?

In addition, can this delay the initial dial attempt to ENs? (because `candidates = [ static | CNs | PNs | ENs ]` but with more CNs and PNs)

networks/p2p/server_multi.go
```diff
@@ -388,98 +404,15 @@ func (srv *MultiChannelServer) handleAddPeerConn(c *conn) error {
 	srv.incrementConnectionCounts(p)
 	srv.reportPeerMetric()
 	srv.logger.Debug("Adding p2p peer", "name", truncateName(c.name), "addr", c.fd.RemoteAddr(), "peers", len(srv.peers))
+	if srv.dialSched != nil {
+		srv.dialSched.OnPeerConnected(c.id, ConvertNodeType(c.conntype), c.Inbound())
```
ian0371, line 408: I'm not sure if it might block so maybe we need goroutine?
```suggestion
		go srv.dialSched.OnPeerConnected(c.id, ConvertNodeType(c.conntype), c.Inbound())
```

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+	ds.connectedOutbound.remove(id)
+	if !inbound {
+		ds.connectedOutbound.add(n)
+	}
+
+	// Reset consecutive failure counter.
+	delete(ds.connFails, id)
+}
+
+func (ds *DialSched) markConnFailure(id discover.NodeID) {
```
ian0371: Maybe distinguish dial events vs. (peer) connection events?

```suggestion
func (ds *DialSched) markDialFailure(id discover.NodeID) {
```

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+
+func (ds *DialSched) markDialingEnd(id discover.NodeID) {
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	ds.dialing.remove(id)
+	ds.dialBackoff[id] = time.Now().Add(dialBackoff)
+}
+
+func (ds *DialSched) markConnSuccess(id discover.NodeID, nType discover.NodeType, inbound bool) {
```
ian0371:
```suggestion
func (ds *DialSched) markPeerConnected(id discover.NodeID, nType discover.NodeType, inbound bool) {
```

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+	if n := ds.static.get(id); n != nil {
+		ds.connFails[id]++
+		if ds.connFails[id] > dialMaxRetries {
+			logger.Warn("Removing static node after too many connection failures", "node", n.String())
+			ds.static.remove(id)
+		}
+	}
+}
+
+func (ds *DialSched) markConnDisconnect(id discover.NodeID) {
```
ian0371:
```suggestion
func (ds *DialSched) markConnDisconnected(id discover.NodeID) {
```

networks/p2p/dialsched.go
```diff
@@ -0,0 +1,534 @@
...
+func (ds *DialSched) dialOnce(n *discover.Node) {
+	defer ds.markDialingEnd(n.ID)
+
+	if ds.dialer == nil || ds.backend == nil {
+		ds.markConnFailure(n.ID)
+		return
+	}
+
+	flags := dynDialedConn
+	if ds.static.contains(n.ID) {
```
ian0371: RLock should be acquired for `ds.static.contains`

networks/p2p/dialsched_test.go
blukat29: Fixed c97583bd1

networks/p2p/dialsched.go
blukat29: Renamed at c97583bd1.

networks/p2p/dialsched.go
blukat29: Renamed at c97583bd1.

networks/p2p/dialsched.go
blukat29: Fixed in 5985d5447. Actually I didn't want `go { dialOnce() }` to hold the lock, so I extracted the branch to `launchDialTasks()`, which already acquires the lock via `shouldDial` and `markDialingStart`.

networks/p2p/dialsched.go
blukat29: Renamed at c97583bd1.

networks/p2p/dialsched.go
blukat29: You are right. Also cleared the backoff at addStatic: ed86c1757.

networks/p2p/dialsched_util.go
blukat29: Yes, for faster count() method.
Previously there was a for-loop to count per type https://github.com/kaiachain/kaia/blob/v2.2.0/networks/p2p/dial.go#L308-L311
typedNodeSet encapsulates and optimizes that.

networks/p2p/dialsched.go
blukat29: Also at 5924ba89c.

networks/p2p/server_multi.go
blukat29: No need. I assess that SetupConn-OnPeerConnected won't block. See the comments added at f45242262.

networks/p2p/dialsched_util.go
ian0371: If the only motivation is faster `count()` method, wouldn't this be more ideal (also faster for all operations)?
```
nodes map[discover.NodeID]*discover.Node
cnt map[discover.NodeType]int
```

networks/p2p/dialsched.go
blukat29: > Is this a new logic?

Kind of.
- Previous dialstate oversampled ENs:
  - When you want dynamic candidates (not enough `dynamicEN in lookupBuf`), you launch a discoverTask. (https://github.com/kaiachain/kaia/blob/v2.2.1/networks/p2p/dial.go#L416-L420).
  - The discoverTask performs tab.Lookup (https://github.com/kaiachain/kaia/blob/v2.2.1/networks/p2p/dial.go#L616)
  - If the target type is EN, Lookup returns at most bucketSize(16) items (https://github.com/kaiachain/kaia/blob/v2.2.1/networks/p2p/discover/discover_storage_kademlia.go#L92)
- Previous dialstate do not oversample CNs,PNs:
  - When you want dynamic candidates (not enough `CN/PN/staticEN in static[]`), you launch a discoverTypedStaticTask, only up to (want - have) items (https://github.com/kaiachain/kaia/blob/v2.2.1/networks/p2p/dial.go#L379-L380)
  - The discoverTypedStaticTask performs tab.GetNodes with only the requested count. (https://github.com/kaiachain/kaia/blob/v2.2.1/networks/p2p/dial.go#L172)

> Is the intention getting as many candidates as possible, and skip most of them during dial?

Correct. I assume that the Discovery table could be stale, or most of them are already connected/dialing.

> can this delay the initial dial attempt to ENs?

Under specific conditions, yes. In practice, unlikely. See the comments added in 8cc022df6.

networks/p2p/dialsched_util.go
blukat29: Simplified as per your suggestion 6bf436fdc.

ian0371 (approved):

PR #784, opened 2026-03-10
reward: move reward logic from istanbul backend to kaiax/reward

yoomee1313:
This PR moves reward logic from istanbul backend to reward module.

### DIFF
--- a/kaiax/reward/impl/consensus.go
+++ b/kaiax/reward/impl/consensus.go
@@ -29,7 +29,6 @@ func (r *RewardModule) VerifyHeader(header *types.Header) error {
 }
 
 func (r *RewardModule) PrepareHeader(header *types.Header) error {
-	// unused fields, force to set to empty
 	header.Rewardbase = r.Rewardbase
 	return nil
 }
--- a/kaiax/reward/impl/consensus.go
+++ b/kaiax/reward/impl/consensus.go
@@ -19,6 +19,9 @@ package impl
 import (
 	"github.com/kaiachain/kaia/blockchain/state"
 	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
+	"github.com/kaiachain/kaia/consensus/istanbul"
+	"github.com/kaiachain/kaia/kaiax/valset"
 )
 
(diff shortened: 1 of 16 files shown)


kaiax/reward/impl/init.go
```diff
@@ -56,7 +60,7 @@ func NewRewardModule() *RewardModule {
 }
 
 func (r *RewardModule) Init(opts *InitOpts) error {
-	if opts == nil || opts.ChainConfig == nil || opts.Chain == nil || opts.GovModule == nil || opts.StakingModule == nil {
+	if opts == nil || opts.ChainConfig == nil || opts.Chain == nil || opts.GovModule == nil || opts.StakingModule == nil || opts.ValsetModule == nil {
```
hyeonLewis, line 63: Is it safe to accept empty `NodeAddress` and `Rewardbase`?

kaiax/reward/impl/consensus.go
```diff
@@ -19,18 +19,38 @@ package impl
...
+	"github.com/kaiachain/kaia/consensus/istanbul"
+	"github.com/kaiachain/kaia/kaiax/valset"
 )
 
 func (r *RewardModule) VerifyHeader(header *types.Header) error {
 	return nil
 }
 
 func (r *RewardModule) PrepareHeader(header *types.Header) error {
+	// unused fields, force to set to empty
```
hyeonLewis: Can we remove this comment?

kaiax/reward/impl/init.go
yoomee1313: Yes, it is safe (and necessary) to accept empty `NodeAddress` / `Rewardbase` in `RewardModule.Init`.
PN/EN also initialize and use `RewardModule`, so hard validation breaks non-proposer nodes.
`NodeAddress` and local `Rewardbase` are proposer-local inputs. They are only needed on CN block-production paths.

kaiax/reward/impl/consensus.go
yoomee1313: Thanks, applied at https://github.com/kaiachain/kaia/pull/784/commits/30d308251cdb6c13fa7630fd1c12e4e4462458fc.

hyeonLewis (approved):

PR #786, opened 2026-03-11
consensus: remove VerifySeal and Initialize from Engine interface

yoomee1313:
## Proposed changes

This PR removes VerifySeal and Initialize from consensus.Engine. Pre-transaction initialization now runs via BlockChain.Initialize(...). In addition, calcBlockScore-related blockscore handling was cleaned up by replacing legacy DefaultBlockScore references with params.DefaultBlockScore.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist


consensus/mocks/engine_mock.go
2dvorak, line 1: We don't need `Initialize` too. Did you run `go generate ./consensus/...`? When I run it locally it gives me a different result.

yoomee1313: Oh Nope.. I didn't. Thanks for finding it! Fixed at https://github.com/kaiachain/kaia/pull/786/commits/73217de60a9a9e35211bb6b87df5445916242a66

blukat29:
@hyunsooda @ian0371 @hyeonLewis Do you think the Initialize() logic belongs to consensus layer or execution layer?
- EIP-2935 is definitely an execution layer construct. Its implementation is also in blockchain/state_processor.go
- What about VRank state transition? The decision is about consensus nodes but writing the results is about EVM execution.

2dvorak (approved):

hyunsooda:
@blukat29 The execution layer seems to be the right place for this logic. Therefore, `blockchain/state_processor.go` would be a good location to relocate it. Similarly, what about `Finalize()` belongs? 
How do you see the following structure?

```go
type ExecutionModule interface { kaiax: Define common module interface 
    // exisiting functions
    PostInsertBlock(block *types.Block) error
    // new functions
    Initialize(...) // e.g., called by valset / vrank modules
    Finalize(...) // reloacte `ConsensusModule's FianlizeHeader`
}

type ConsensusModule interface {
    VerifyHeader(header *types.Header) error
    PrepareHeader(header *types.Header) error
    // Fianlize(....) relocated to ExecutionModule
}
```

blukat29:
@hyeonLewis That seems right in the long term.

hyeonLewis:
@blukat29, it seems Lake and I have confusingly similar profiles. Anyway, I agree with Lake's idea. One caveat is that currently we assume `ExecutionModule` won't make any state changes, while the newly introduced `Initialize/Finalize` might change the state. It means we need to relocate where `ExecutionModule` is executed or make a separate module for them.

ian0371 (commented):
I think `Initialize()` belongs to execution layer.
Also I think execution layer and execution layer can be dependent on each other.
So while VRank state transition is consensus layer decision, it can rely on execution layer for the input and output.

PR #787, opened 2026-03-11
p2p: Replace Discovery with Discovery2

blukat29:
## Proposed changes

- Replace `Discovery` with `Discovery2`. Use it in p2p.Server and p2p.DialSched.
- Update DialSched's dynamic candidate selection logic.
  - Previously, the oversampling could probabilistically return the same ineligible node over and over.
  - Now the samples are first dumped onto a queue then consumed sequentially.
```
# First attempt. Empty queue. Async refill.
getDynamicCandidates():   queue empty -> return (nil, needRefresh=true)
dialLoop:                 needRefresh -> refreshOnce(refreshResCh)
dialLoop:                 case <-refreshResCh: refillCandidates()
refillCandidates():       tab.RandomNodes() → fill candidateQueue

# Second attempt. Consume queue
getDynamicCandidates():   pop from candidateQueue
```

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Related issues

- #751

### DIFF
--- a/networks/p2p/dialsched.go
+++ b/networks/p2p/dialsched.go
@@ -271,8 +271,7 @@ func (ds *DialSched) dialLoop() {
 		select { // Loop upon any of the following signals.
 		case <-ds.wakeDial: // situation changed
 		case <-dialResCh: // one dial task completed (either success or failure)
-		case <-refreshResCh: // discovery refresh completed; refill the candidate queue
-			ds.refillCandidates()
+		case <-refreshResCh: // discovery refresh completed; next getCandidates will
 		case <-idle: // done idling
 		case <-ds.closeReq:
 			return
@@ -337,16 +336,17 @@ func (ds *DialSched) getDynamicCandidates(targetType discover.NodeType, need int
 		return nil, false
 	}
 
-	// Empty queue: signal the dial loop to run a discovery refresh.
-	// We'll come back after the refresh completes and refill the queue.
-	if len(ds.candidateQueue[targetType]) == 0 {
-		return nil, true
+	// Try to refill from the table if the queue can't satisfy the demand.
+	// The table may already have nodes (e.g. from local nodeDB) without needing a refresh
+	if len(ds.candidateQueue[targetType]) < need {
+		ds.refillCandidates()
 	}
 
 	// Pop up to `need` candidates from the queue.
 	q := ds.candidateQueue[targetType]
 	end := min(need, len(q))
 	nodes, ds.candidateQueue[targetType] = q[:end], q[end:]
+	// If the queue is short even after a refill, signal the dial loop to run a discovery refresh.
 	return nodes, len(q) < need
 }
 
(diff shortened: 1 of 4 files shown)


networks/p2p/dialsched.go
```diff
@@ -294,36 +314,52 @@ func (ds *DialSched) getCandidates() (candidates []*discover.Node, needRefresh b
...
+		return nil, true
 	}
+
+	// Scan the queue, advancing past undialable entries, until we have collected `need` candidates.
+	// shouldDial() called here as an inaccurate screening to reject hopeless candidates.
+	// shouldDial() called later as an accurate enforcement before launching the dialOnce goroutine.
+	q := ds.candidateQueue[targetType]
+	i := 0
+	for i < len(q) && len(nodes) < need {
+		if ds.shouldDial(q[i]) {
```
ian0371: I'm not sure if `shouldDial()` fulfills the goal "reject hopeless candidates". It seems too aggressive.
For example, `ds.dialing.len() >= maxConcurrentDials` check can filter naive nodes.

It looks like this inaccuracy is intentional. So what's the side effect? Having to wait for 4s + refresh time for the next dial?

networks/p2p/dialsched.go
```diff
@@ -294,36 +314,52 @@ func (ds *DialSched) getCandidates() (candidates []*discover.Node, needRefresh b
...
-		num := ds.tab.ReadRandomNodes(random, targetType)
-		nodes := random[:min(num, len(random))]
-		return nodes, len(nodes) < need
-	default:
-		return nil, false
+
+	// Empty queue: signal the dial loop to run a discovery refresh.
+	// We'll come back after the refresh completes and refill the queue.
+	if len(ds.candidateQueue[targetType]) == 0 {
+		return nil, true
```
ian0371: Instead of returning, can we call `refillCandidates()` directly here?

networks/p2p/dialsched.go
blukat29: Now that I have the candidateQueue, I need not worry about retrying the hopeless candidates. Then I think the eager filtering is unnecessary. It's only inaccurate step as you've pointed out.

networks/p2p/dialsched.go
blukat29: The assumption was that when candidateQueue is empty, we have to refresh first, then refill the queue.
But there could be the case where the table is already populated (e.g. from local nodeDB) so we don't have to wait for a refresh. I'll fix that.

networks/p2p/dialsched.go
blukat29: Updated by e5839ac6b.

networks/p2p/dialsched.go
blukat29: Updated by aaeff5553.

ian0371 (approved):

PR #794, opened 2026-03-17
permless: harden vrank & add vrank cache, schema

ian0371:
### In-memory LRU cache + checkpoint DB
- `GetPFS`/`GetCFS` store results in an LRU cache keyed by block number; incremental updates start from the nearest cached ancestor, skipping full epoch re-scans
- `PostInsertBlock` eagerly warms the cache and writes a combined PFS+cpMatrix checkpoint to the DB every `scoreCheckpointInterval` blocks
- On `Init`, `catchUpScoreCaches` bootstraps from the latest DB checkpoint and replays only the tail, bounding startup cost to at most one interval of blocks

### Rewind interface
- `RewindTo` purges all in-memory caches (pfsCache, cpMatrixCache)
- `RewindDelete` removes DB checkpoints at the new chain head and rolls back the `lastCheckpoint` pointer

### Hardfork guards
- All getters (`GetPFS`, `GetCFS`, `GetPfReport`, `GetCfReport`, `TallyCfReport`) return early before the permissionless fork
- `PostInsertBlock` skips cache/DB writes before the fork

### Consensus validation hardening
- `VerifyHeader`: epoch-start block must have empty VRank; addresses must be deduplicated, sorted ascending, and all must be known candidates for block `N-1`
- Handlers use `GetCommittee` (ValActive only) instead of `GetCouncil`
- `computeCPMatrix` skips unknown addresses in `header.VRank` instead of inserting them
- `Extra.Vals` must be a correct list of validators.

### VRankPreprepare signing
- Proposer signs `keccak256("VRANK_PREPREPARE_V1" || chainID || blockNum || round || blockHash)` before broadcasting
  - This prevents a malicious validator from broadcasting invalid VRankPreprepare to candidates
- Candidate verifies recovered signer equals `GetProposer(N, round)` before responding — prevents a malicious committee member from sending a fake preprepare with a wrong block hash to poison candidates' signatures

### handleVRankCandidate gates removal
- Removed receive-time committee/candidate gating from HandleVRankCandidate.
  - The handler now acts as a bounded inbox: reject stale/too-far messages, verify signature, dedup by sender per view, and store.
  - Canonical candidate validation is deferred to TallyCfReport.
 
See further comments for rationale.

### Change TallyCfReport filtering
- don't check `v.isValidator`
- `prepreparedAt.IsZero()` returns no error
- use correct `GetCandidates(blockNum)`

### Others
- Idempotent Start/Stop: rewind will call Stop followed by Start
- api: `header.VRank` removed from `RpcOutputBlock`; the field only exists post-fork and has no stable consumer yet
- add `README.md`

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

- Candidate gate removal:
  - a candidate that becomes valid in the next epoch can be rejected by a node still anchored to the previous epoch's local view
  - this can cause a false candidate failure in the later `cfReport`
- Committee gate removal:
  - a node that becomes a committee member in the next epoch can reject early valid candidate messages before its local `prepreparedView` catches up
  - if that node later proposes the next block, it may write `header.VRank` without messages it should have counted

### ValSet call tree
ValSet getters depend on state, so queries with old blocks will not work on full node.

<details>

<summary>Call tree</summary>

```
Valset.GetProposer(N, R)
├── ✓ pfReport(N)                     R ∈ [0, header(N).Round)   — uses Extra.Validators via getBlockContext
│   ├── GetPfReport(N)
│   └── applyBlocksForPFS(start, end) N ∈ [start, end]
│       └── GetPFS(N)
├── ✓ applyBlocksForCPMatrix(start, end)  R = header(N).Round    — fast path: Engine.Author
│   ├── GetCFS(N)
│   └── computeCFS(N)
├── ✓ verifyVRankPreprepareSender(N, round)                       N ≈ head
│   └── HandleVRankPreprepare(N)
└── ✓ isProposer(N, round)                                        N ≈ head
    └── HandleIstanbulPreprepare(N)

Valset.GetCandidates(N)
├── ✓ newCPMatrix(N)                                              N ≈ head
│   ├── lookupCFSSeed(N)
│   │   └── GetCFS(N)
│   └── computeCFS(N)
├── ✓ TallyCfReport(N, round)                                     N ≈ head−1
├── ✓ BroadcastVRankPreprepare(N)                                 N ≈ head
│   └── HandleIstanbulPreprepare(N)
├── ✓ isCandidate(N)                                              N ≈ head
│   └── HandleVRankPreprepare(N)
└── ✓ VerifyHeader(N)                  validates header(N).VRank  N ≈ head

Valset.GetCommittee(N, round)
├── ⚠️ generateCFSFromCPMatrix(N, 0)   TODO: replace with AddressBook call
│   ├── GetCFS(N)
│   └── computeCFS(N)
├── ✓ BroadcastVRankCandidate(N, round)                           N ≈ head
│   └── HandleVRankPreprepare(N)
└── ✓ isCommitteeMember(N, round)                                 N ≈ head
    └── HandleIstanbulPreprepare(N)
```

</details>


kaiax/vrank/impl/init.go
```diff
@@ -72,38 +82,154 @@ type VRankModule struct {
...
+		start = epochStart
+		pfs = make(map[common.Address]uint64)
+		var err error
+		cpMatrix, err = v.newCPMatrix(epochStart)
+		if err != nil {
+			return err
+		}
+	}
+
+	for blockNum := start; blockNum <= headNum; blockNum++ {
```
hyeonLewis: It seems `start` might be before the permissionless fork. How about adding last check for `start`?

```
   86,400              Permless fork (90,000)          headNum (95,000)
-----|-------------------------|------------------------------|-------------
```

kaiax/vrank/impl/init.go
```diff
@@ -72,38 +82,154 @@ type VRankModule struct {
...
+			return err
+		}
+		v.cpMatrixCache.Add(blockNum, cloneCPMatrix(cpMatrix))
+
+		if blockNum%scoreCheckpointInterval == 0 {
+			WriteCheckpoint(v.ChainKv, blockNum, pfs, cpMatrix)
+			WriteLastCheckpoint(v.ChainKv, blockNum)
+		}
+
+		if blockNum%10000 == 0 {
```
hyeonLewis:
```suggestion
		if blockNum%10000 == 0 && start != headNum {
```

kaiax/vrank/impl/consensus.go
```diff
@@ -0,0 +1,91 @@
...
+	// Validate every address in the cfReport is an actual candidate.
+	// header(N).VRank is produced by TallyCfReport(N-1, ...) which uses
+	// GetCandidates(calcEpochStart(N)), so mirror that here.
+	// This prevents a malicious proposer from injecting arbitrary addresses (e.g. validator
+	// addresses) into header.VRank to manipulate CFS scores.
+	candidates, err := v.Valset.GetCandidates(calcEpochStart(number))
+	if err != nil {
+		return err
+	}
+	candSet := make(map[string]struct{}, len(candidates))
```
hyeonLewis: Can't we use the address as a key here?

kaiax/vrank/impl/execution.go
```diff
@@ -0,0 +1,62 @@
...
+	if _, err := v.GetCFS(blockNum); err != nil {
+		return err
+	}
+
+	if blockNum%scoreCheckpointInterval == 0 {
+		// GetCFS populates cpMatrixCache (not a separate cfsCache), so fetch from there.
+		cpRaw, hasCPS := v.cpMatrixCache.Get(blockNum)
+		var cpMatrix map[common.Address]map[common.Address]uint64
+		if hasCPS {
+			cpMatrix = cpRaw.(map[common.Address]map[common.Address]uint64)
```
hyeonLewis:
```suggestion
			cpMatrix = cloneCPMatrix(cpRaw.(map[common.Address]map[common.Address]uint64))
```

kaiax/vrank/impl/schema.go
```diff
@@ -0,0 +1,154 @@
...
+	slices.SortFunc(candidates, func(a, b common.Address) int { return bytes.Compare(a.Bytes(), b.Bytes()) })
+
+	cfsEntries := make([]cpMatrixEntry, 0)
+	for _, candidate := range candidates {
+		reporters := make([]common.Address, 0, len(cpMatrix[candidate]))
+		for reporter := range cpMatrix[candidate] {
+			reporters = append(reporters, reporter)
+		}
+		slices.SortFunc(reporters, func(a, b common.Address) int { return bytes.Compare(a.Bytes(), b.Bytes()) })
+		for _, reporter := range reporters {
```
hyeonLewis: Candidates who haven't failed VRank will be discarded here, so when restarting, it won't be cached at catchUp and will be [skipped](https://github.com/kaiachain/kaia/pull/794/changes#diff-930d80c1146e07bc80318ec6f73b0848cea3c2538746b4cc99d98deac434ef57R193). I think we could use a sentinel value or seed all missing candidates again.

kaiax/vrank/impl/consensus.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return vrank.ErrInvalidVRankFormat
+	}
+
+	// Validate every address in the cfReport is an actual candidate.
+	// header(N).VRank is produced by TallyCfReport(N-1, ...) which uses
+	// GetCandidates(calcEpochStart(N)), so mirror that here.
+	// This prevents a malicious proposer from injecting arbitrary addresses (e.g. validator
+	// addresses) into header.VRank to manipulate CFS scores.
+	candidates, err := v.Valset.GetCandidates(calcEpochStart(number))
```
hyeonLewis: Is it okay here for a full node?

kaiax/vrank/impl/init.go
ian0371: I assumed permissionless fork block should be `k*vrankEpoch`. Is this assumption not ok?
I didn't think about such case when writing KIP-227.

hyeonLewis: I was initially worried about the private network, but it will be ok since they mostly use genesis fork (if not, it should correctly set the fork number for testing)

kaiax/vrank/impl/rewind.go
```diff
@@ -0,0 +1,52 @@
...
+// along with the Kaia library. If not, see <http://www.gnu.org/licenses/>.
+
+package impl
+
+import (
+	"github.com/kaiachain/kaia/blockchain/types"
+	"github.com/kaiachain/kaia/common"
+)
+
+func (v *VRankModule) RewindTo(newBlock *types.Block) {
```
hyeonLewis, line 24: PR desc says collector is purged here, but it's not. Which one is the latest spec?

ian0371: `collector` is not purged. updated PR desc. thanks.

kaiax/vrank/impl/getter.go
hyeonLewis: This is not PR content, but we don't need safeCands separately, I think. Hdyt?

```go
var cfReport vrank.Report                                                                                                                                                                                                                           
for _, addr := range candidates {                                                                                                                                                                                                                   
  msgWithTime, ok := viewMap[addr]                                                                                                                                                                                                                
  if ok && msgWithTime.Msg != nil && msgWithTime.Msg.BlockHash == expectedBlockHash &&                                                                                                                                                            
      msgWithTime.ReceivedAt.Sub(prepreparedAt).Milliseconds() <= candidateMsgTimeoutMs {                                                                                                                                                         
      continue // safe candidate                                                                                                                                                                                                                  
  }                                                                                                                                                                                                                                               
  cfReport = append(cfReport, addr)                                                                                                                                                                                                               
}
```

kaiax/vrank/impl/init.go
```diff
@@ -72,38 +82,154 @@ type VRankModule struct {
...
+	return nil
+}
+
+func (v *VRankModule) loadCheckpointInEpoch(blockNum uint64) (uint64, map[common.Address]uint64, map[common.Address]map[common.Address]uint64, bool) {
+	epochStart := calcEpochStart(blockNum)
+	lastCP, ok := ReadLastCheckpoint(v.ChainKv)
+	if !ok || lastCP < epochStart {
+		return 0, nil, nil, false
+	}
+	cpNum := min(lastCP, calcCheckpointBlock(blockNum))
```
hyeonLewis: I think we can use `calcCheckpointBlock` directly.

i) lastcheckpoint > calcCheckpointBlock: no issue (likely happens when rewinding)
ii) lastcheckpoint == calcCheckpointBlock: no issue (likely happens in normal reboot)
iii) lastcheckpoint < calcCheckpointBlock: can't happen (DB stored checkpoint is persistent)

hyeonLewis: One case for (iii) might be: after rewinding, the headNum starts with a gap (stop at N, start at N+86,400), which can only happen with chaindata download, which means it already download checkpoint together.

kaiax/vrank/impl/consensus.go
hyeonLewis: It seems there are many places that depend on this (and the proposer). In normal operation, it seems ok since the state will be cached for N-1 blocks, but it might not be safe when the node startup/rewind.

kaiax/vrank/impl/schema.go
ian0371: nice catch! revised DB schema b28f33127

kaiax/vrank/impl/consensus.go
ian0371: fixed 089984a18

kaiax/vrank/impl/consensus.go
ian0371: > it might not be safe when the node startup/rewind.

Agreed. Should ValSet module regen states?
Is VRank the only caller of ValSet with a past block?
cc @hyunsooda

kaiax/vrank/impl/getter.go
ian0371: fixed 3d4537d09

kaiax/vrank/impl/init.go
ian0371: I agree 3d4537d09

kaiax/vrank/impl/consensus.go
ian0371: It looks like vrank was a special case which called valset with old block.
`GetCandidates(epochStart)` will use `GetCandidates(number)` now.
fixed 7f2fffab4

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
+		for _, addr := range report {
+			pfs[addr]++
+		}
 	}
+	return pfs, nil
+}
 
+// applyBlocksForCPMatrix accumulates the candidate-proposer matrix for blocks in [start, end].
+func (v *VRankModule) applyBlocksForCPMatrix(start, end uint64, seed map[common.Address]map[common.Address]uint64) (map[common.Address]map[common.Address]uint64, error) {
+	cpMatrix := cloneCPMatrix(seed)
```
hyeonLewis:
```suggestion
```

kaiax/vrank/impl/execution.go
```diff
@@ -0,0 +1,62 @@
...
+	pfs, err := v.GetPFS(blockNum)
+	if err != nil {
+		return err
+	}
+	if _, err := v.GetCFS(blockNum); err != nil {
+		return err
+	}
+
+	if blockNum%scoreCheckpointInterval == 0 {
+		// GetCFS populates cpMatrixCache (not a separate cfsCache), so fetch from there.
```
hyeonLewis:
```suggestion
		// GetCFS populates cpMatrixCache, so fetch from there.
```

kaiax/vrank/impl/execution.go
```diff
@@ -0,0 +1,62 @@
...
+		cpRaw, hasCPS := v.cpMatrixCache.Get(blockNum)
+		var cpMatrix map[common.Address]map[common.Address]uint64
+		if hasCPS {
+			cpMatrix = cloneCPMatrix(cpRaw.(map[common.Address]map[common.Address]uint64))
```
hyeonLewis: I'm sorry I initially thought we need copy here but it seems there's no concurrent read/write since `lookupCPMatrixCache` is always returns cloned matrix.

On second thought, using `lookupCPMatrixCache` here too would be consistent, but it has redundant `cloneCPMatrix`, so I think the current lookup is okay. Hdyt?

```suggestion
			cpMatrix = cpRaw.(map[common.Address]map[common.Address]uint64)
```

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
+	start, seed := v.lookupPFSSeed(blockNum)
+	if start <= blockNum {
+		var err error
+		seed, err = v.applyBlocksForPFS(start, blockNum, seed)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	v.pfsCache.Add(blockNum, cloneMap(seed))
```
hyeonLewis:
```suggestion
	v.pfsCache.Add(blockNum, seed)
```

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
-		}
 	}
-	return pfs, nil
+
+	cfs, err := v.generateCFSFromCPMatrix(blockNum, seed)
+	if err != nil {
+		return nil, err
+	}
+
+	v.cpMatrixCache.Add(blockNum, cloneCPMatrix(seed))
```
hyeonLewis:
```suggestion
	v.cpMatrixCache.Add(blockNum, seed)
```

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
+	return cloneMap(cfs), nil
```
hyeonLewis:
```suggestion
	return cfs, nil
```

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
+	cpMatrix := make(map[common.Address]map[common.Address]uint64, len(candidates))
+	for _, candidate := range candidates {
+		cpMatrix[candidate] = make(map[common.Address]uint64)
+	}
+	return cpMatrix, nil
+}
+
+// applyBlocksForPFS accumulates the pfReport for blocks in [start, end].
+func (v *VRankModule) applyBlocksForPFS(start, end uint64, seed map[common.Address]uint64) (map[common.Address]uint64, error) {
+	pfs := cloneMap(seed)
```
hyeonLewis, line 133:
```suggestion
```

kaiax/vrank/impl/scoring.go
hyeonLewis:
```suggestion
```

hyeonLewis (commented):
We can safely remove some redundant clones.

kaiax/vrank/impl/execution.go
ian0371: Right. Only mutators are `apply*` functions, so we only need copy from there.
One exception is `GetPFS`; I want to separate cache and return obj, so any update from callers doesn't effect cache. fixed 7bc34b7eb

kaiax/vrank/impl/scoring.go
ian0371: I think we need clone here. I don't want to directly write to given object.

kaiax/vrank/impl/scoring.go
ian0371: I think we need clone here. I don't want to directly write to given object.

kaiax/vrank/impl/execution.go
hyeonLewis: Ok, so I understood clone at `GetPFS` is a defensive one.

hyeonLewis (approved):

kaiax/vrank/impl/consensus.go
hyeonLewis: Should revisit here with valset implementation.

ian0371: Current situation:

```
Init
└── catchUpScoreCaches
    ├── GetPFS(head)
    │   ├── lookupPFSSeed
    │   │   ├── pfsCache          (hit → done)
    │   │   └── ReadCheckpoint DB (hit → done)
    │   └── applyBlocksForPFS  [start..head]
    │       └── pfReport(x)
    │           └── ⚠️ Valset.GetProposer(x, r<Round)        OLD BLOCK
    └── GetCFS(head)
        ├── lookupCFSSeed
        │   ├── cpMatrixCache     (hit → done)
        │   ├── ReadCheckpoint DB (hit → done)
        │   └── newCPMatrix(head)
        │       └── ✓ Valset.GetCandidates(head)
        ├── applyBlocksForCPMatrix  [start..head]
        │   └── ✓ Valset.GetProposer(x, header.Round)        fast path: Engine.Author
        └── generateCFSFromCPMatrix
            └── ⚠️ Valset.GetCommittee(blockNum, 0)          TODO: will be replaced

PostInsertBlock(head)
├── GetPFS(head)                                              (same as above)
├── GetCFS(head)                                              (same as above)
└── WriteCheckpoint → DB  (every scoreCheckpointInterval)

TallyCfReport(N, round)
└── ✓ Valset.GetCandidates(N)        N ≈ head−1

// all handlers use head+1
HandleIstanbulPreprepare(head+1)
├── isCommitteeMember(head, round)
│   └── ✓ Valset.GetCommittee(N, round)
├── isProposer(N, round)
│   └── ✓ Valset.GetProposer(N, round)
└── BroadcastVRankPreprepare  [if proposer]
    └── ✓ Valset.GetCandidates(N)

HandleVRankPreprepare(block N)
├── isCandidate(N)
│   └── ✓ Valset.GetCandidates(N)
├── verifyVRankPreprepareSender(N, round)
│   └── ✓ Valset.GetProposer(N, round)
└── BroadcastVRankCandidate  [if candidate]
    └── ✓ Valset.GetCommittee(N, round)

HandleVRankCandidate
└── (no Valset calls)
```

Fixed `GetProposer` to look headers first. Also fixed `verifyHeader` to validate `Extra.Vals`.

hyeonLewis (approved):

kaiax/vrank/impl/scoring.go
```diff
@@ -143,3 +219,39 @@ func byzantineFilter(
...
+		if cached, ok := v.pfsCache.Get(candidateNum); ok {
+			return candidateNum, cached.(map[common.Address]uint64), true
+		}
+	}
+	return 0, nil, false
+}
+
+func (v *VRankModule) lookupCPMatrixCache(blockNum uint64) (uint64, map[common.Address]map[common.Address]uint64, bool) {
+	epochStart := calcEpochStart(blockNum)
+	for i := uint64(0); i <= scoreCacheProbeLookback && i <= blockNum-epochStart; i++ {
```
hyunsooda:
```suggestion
	for i := uint64(0); i <= min(scoreCacheProbeLookback, blockNum-epochStart); i++ {
```

kaiax/vrank/impl/scoring.go
```diff
@@ -18,102 +18,178 @@ package impl
...
+	return cfs, nil
+}
+
+// lookupPFSSeed returns (start, seed) where seed holds PFS accumulated up to start-1.
+// Callers should apply [start, blockNum] on top of seed to reach blockNum.
+func (v *VRankModule) lookupPFSSeed(blockNum uint64) (start uint64, seed map[common.Address]uint64) {
+	if cachedNum, pfs, ok := v.lookupPFSCache(blockNum); ok {
+		return cachedNum + 1, pfs
+	}
+	if cpNum, pfs, _, ok := v.loadCheckpointInEpoch(blockNum); ok {
```
hyunsooda: Avoiding matrix calculations seems to result in better cache characteristics. What do you think, because it looks expensive in the worst case?

kaiax/vrank/impl/getter.go
```diff
@@ -38,9 +39,19 @@ func (v *VRankModule) GetCfReport(blockNum uint64) (vrank.Report, error) {
 	return vrank.DecodeReport(header.VRank)
 }
 
-// GetPfReport reads the committed pfReport for block blockNum from header.Extra.
-// Returns an empty report if the block was finalized at round 0.
-func (v *VRankModule) GetPfReport(blockNum uint64) (vrank.Report, error) {
+// GetCfReport reads the committed cfReport for block blockNum from header.VRank.
+// Returns an empty report if header.VRank is nil (e.g. epoch-start block).
+// Returns ErrNotPermissionless if blockNum is before the permissionless fork.
+func (v *VRankModule) GetCfReport(blockNum uint64) (vrank.Report, error) {
```
hyunsooda: GetCfReport / GetPfReport are only called from tests. Consider moving them to test helpers and removing them from the public interface, unless they're intended for future RPC/debug use.

kaiax/vrank/impl/getter.go
```diff
@@ -66,9 +77,24 @@ func (v *VRankModule) GetPfReport(blockNum uint64) (vrank.Report, error) {
 	return pfReport, nil
 }
 
+// GetPfReport reads the committed pfReport for block blockNum from header.Extra.
+// Returns an empty report if the block was finalized at round 0.
+// Returns ErrNotPermissionless if blockNum is before the permissionless fork.
+func (v *VRankModule) GetPfReport(blockNum uint64) (vrank.Report, error) {
```
hyunsooda: https://github.com/kaiachain/kaia/pull/794/changes#r2986534008

hyunsooda (commented):
Please consider introducing a CPMatrix type for `map[common.Address]map[common.Address]uint64`. It appears ~30 times
across the codebase, and a named type would improve readability. Methods like Clone() and Increment() could encapsulate nil-map guards. Conversely, Report is just `[]common.Address` with no extra methods — could be inlined.
```go
 type CPMatrix map[common.Address]map[common.Address]uint64

func (m CPMatrix) Clone() CPMatrix {
     ret := make(CPMatrix, len(m))
     for k, v := range m {
             inner := make(map[common.Address]uint64, len(v))
             maps.Copy(inner, v)
             ret[k] = inner
     }
     return ret
 }

 func (m CPMatrix) Increment(candidate, reporter common.Address) {
     if m[candidate] == nil {
             m[candidate] = make(map[common.Address]uint64)
     }
     m[candidate][reporter]++
 }
```

kaiax/vrank/impl/scoring.go
ian0371: fixed a32ee09d8

kaiax/vrank/impl/getter.go
ian0371: fixed a32ee09d8
introduced CPMatrix c0a8fd9c5

kaiax/vrank/impl/scoring.go
ian0371: Could you elaborate on "avoiding matrix calculations"? Are you suggesting that we merge PFS/CFS cache?

kaiax/vrank/impl/scoring.go
```diff
@@ -143,3 +215,31 @@ func byzantineFilter(
...
+
+func cloneMap(src map[common.Address]uint64) map[common.Address]uint64 {
+	ret := make(map[common.Address]uint64, len(src))
+	maps.Copy(ret, src)
+	return ret
+}
+
+func (v *VRankModule) lookupPFSCache(blockNum uint64) (uint64, map[common.Address]uint64, bool) {
+	epochStart := calcEpochStart(blockNum)
+	for i := uint64(0); i <= scoreCacheProbeLookback && i <= blockNum-epochStart; i++ {
```
hyunsooda: nit: same pattern with https://github.com/kaiachain/kaia/pull/794#discussion_r2985460137

kaiax/vrank/impl/scoring.go
hyunsooda: `ReadCheckpoint` deserializes both CFS and PFS. However, for most specific calls, only one is actually needed. How about separating the checkpoint reading functions?

ian0371: Sounds good.  fixed eb247444f

kaiax/vrank/impl/scoring.go
ian0371: fixed eb247444f

kaiax/vrank/impl/scoring.go
hyunsooda: Two callers: `rewind` and `cache` (`lookupPFSSeed` and `lookupCFSSeed`). The latter doesn't appear to be considered, so I imagine that `ReadCheckpoint` would be entirly disappear.

ian0371: Removed all `ReadCheckpoint`: 66b97ec60

kaiax/vrank/types.go
```diff
@@ -17,19 +17,56 @@
...
+	}
+	return cpMatrix
+}
+
+func (m CPMatrix) Clone() CPMatrix {
+	clone := make(CPMatrix, len(m))
+	for candidate, reporters := range m {
+		inner := make(map[common.Address]uint64, len(reporters))
+		maps.Copy(inner, reporters)
+		maps.Copy(inner, reporters)
```
hyunsooda: Please remove duplicate copies.

hyunsooda (approved):

kaiax/vrank/types.go
ian0371: fixed f7de44ac3

PR #809, opened 2026-03-23
consensus: fix startup race in ValidatePeerType causing delayed first block

hyunsooda:
## Proposed changes

Fix a startup race condition in `ValidatePeerType` that delays the first block consensus by ~30-40 seconds.

### Problem

The P2P server starts before `engine.Start()` sets `sb.chain`. When peers connect during this window, `ValidatePeerType` hits `sb.chain == nil` and immediately returns `errNoChainReader`, tearing down the connection. The P2P layer then waits ~30 seconds before retrying, causing multiple round changes and delaying the first block.

```
[1] p2pServer.Start()     ← peers can connect
    ...                    ← gap: sb.chain is nil
[2] engine.Start()
    ├ SetChain()           ← sb.chain set
    ├ core.Start()         ← consensus core starts
    └ coreStarted = true   ← fully ready
```

**Before fix (cn4, proposer):**
```
INFO [10:34:50] Kaia peer registration failed  err="fail to validate peer type: 0x3C44...(sb.chain is nil! --mine option might be missing)"
INFO [10:34:50] Kaia peer registration failed  err="fail to validate peer type: 0x7099...(sb.chain is nil! --mine option might be missing)"
INFO [10:34:50] Commit new mining work         number=1
WARN [10:35:00] [RC] timeoutEvent Sent!        sequence=1 round=0 preprepare is nil?=false len(prepares)=1
WARN [10:35:12] [RC] timeoutEvent Sent!        sequence=1 round=1 preprepare is nil?=true  len(prepares)=0
WARN [10:35:26] [RC] timeoutEvent Sent!        sequence=1 round=2 preprepare is nil?=true  len(prepares)=0
INFO [10:35:40] Successfully wrote mined block num=1  ← ~36s delay
```

### Solution

Replace the immediate error return with a channel-based wait (`chainInitCh`). `ValidatePeerType` now blocks on `<-sb.chainInitCh` until the engine is fully initialized (chain set + core started), instead of rejecting peers and relying on slow P2P reconnection.

```go
// Before: immediate reject → peer dropped → 30s retry
for sb.chain == nil {
    return errNoChainReader
}

// After: block until engine ready → peer registered immediately
select {
case <-sb.chainInitCh:
case <-time.After(chainInitTimeout):
    return errNoChainReader
}
```

`SignalPeerRegistrable()` closes the channel at the right moment:
- **Mining nodes**: called via `defer` in `Start()`, after `coreStarted = true`
- **WorkerDisable nodes**: called explicitly after `SetChain()`, since `Start()` is never invoked
- **Stop()**: called to unblock any waiting peers during shutdown

`sync.Once` ensures the channel is closed exactly once regardless of the code path. A 30-second timeout acts as a safety net against goroutine leaks.

### Result

- **Before**: `sb.chain is nil` errors → 3-4 round changes → ~30-40s to first block
- **After**: no errors, no round changes → first block produced immediately

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Related issues

None — this is a pre-existing race condition discovered during permissionless development with local testnet.

## Further comments

The race condition exists on the `dev` branch as well and is reproducible. The removed test case (`backend.chain = nil` after `SetChain`) tested an unrealistic scenario — `chain` is never set back to nil after initialization in production.

### DIFF
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -711,9 +711,15 @@ func (sb *backend) SetChain(chain consensus.ChainReader) {
 	sb.chain = chain
 }
 
-// SignalPeerRegistrable unblocks ValidatePeerType.
-// For mining nodes, this is called automatically at the end of Start().
-// For non-mining nodes (WorkerDisable), the caller must invoke this after SetChain.
+// SignalPeerRegistrable unblocks ValidatePeerType so that peers can be registered.
+// It is safe to call multiple times; only the first call has effect (sync.Once).
+//
+// Call sites:
+//   - Mining nodes: called automatically via defer in Start(), after coreStarted=true.
+//     Also covers Start() failure paths to prevent goroutine leaks.
+//   - WorkerDisable nodes: caller must invoke this explicitly after SetChain(),
+//     since Start() is never called.
+//   - Stop(): called to unblock any peers waiting during shutdown.
 func (sb *backend) SignalPeerRegistrable() {
 	sb.chainInitOnce.Do(func() { close(sb.chainInitCh) })
 }
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -758,6 +758,9 @@ func (sb *backend) Start(chain consensus.ChainReader, currentBlock func() *types
 	}
 	sb.commitCh = make(chan *types.Result, 1)
 
+	// Ensure peers are unblocked even if Start() fails partway through.
+	defer sb.SignalPeerRegistrable()
+
 	sb.SetChain(chain)
 	sb.currentBlock = currentBlock
 	sb.hasBadBlock = hasBadBlock
@@ -767,7 +770,6 @@ func (sb *backend) Start(chain consensus.ChainReader, currentBlock func() *types
 	}
 
 	sb.coreStarted = true
-	sb.SignalPeerRegistrable()
 	return nil
 }
 
@@ -775,6 +777,8 @@ func (sb *backend) Start(chain consensus.ChainReader, currentBlock func() *types
 func (sb *backend) Stop() error {
 	sb.coreMu.Lock()
 	defer sb.coreMu.Unlock()
+	// Unblock any peers waiting in ValidatePeerType during shutdown.
+	sb.SignalPeerRegistrable()
 	if !sb.coreStarted {
 		return istanbul.ErrStoppedEngine
 	}
--- a/consensus/istanbul/backend/engine.go
+++ b/consensus/istanbul/backend/engine.go
@@ -765,6 +765,9 @@ func (sb *backend) Start(chain consensus.ChainReader, currentBlock func() *types
 	sb.commitCh = make(chan *types.Result, 1)
 
 	// Ensure peers are unblocked even if Start() fails partway through.
+	// This defer is intentionally placed before SetChain: if Start() fails
+	// before SetChain, sb.chain remains nil, which is safely handled by
+	// the nil guard in ValidatePeerType.
 	defer sb.SignalPeerRegistrable()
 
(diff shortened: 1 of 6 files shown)


consensus/istanbul/backend/engine.go
```diff
@@ -767,6 +786,8 @@ func (sb *backend) Start(chain consensus.ChainReader, currentBlock func() *types
 func (sb *backend) Stop() error {
 	sb.coreMu.Lock()
 	defer sb.coreMu.Unlock()
+	// Unblock any peers waiting in ValidatePeerType during shutdown.
+	sb.SignalPeerRegistrable()
```
ian0371, line 790: Unless `RegisterKaiaxModules()` call is skipped somewhere, this line doesn't seem necessary given that `SignalPeerRegistrable()` is idempotent. Is there such a case?

consensus/istanbul/backend/engine.go
```diff
@@ -711,6 +711,19 @@ func (sb *backend) SetChain(chain consensus.ChainReader) {
...
+// SignalPeerRegistrable unblocks ValidatePeerType so that peers can be registered.
+// It is safe to call multiple times; only the first call has effect (sync.Once).
+//
+// Call sites:
+//   - Mining nodes: called automatically via defer in Start(), after coreStarted=true.
+//     Also covers Start() failure paths to prevent goroutine leaks.
+//   - WorkerDisable nodes: caller must invoke this explicitly after SetChain(),
+//     since Start() is never called.
+//   - Stop(): called to unblock any peers waiting during shutdown.
+func (sb *backend) SignalPeerRegistrable() {
```
ian0371, line 723: Can this logic be embedded into `SetChain`?

consensus/istanbul/backend/engine.go
hyunsooda: This is a defensive measure to immediately unblock peer goroutines waiting in `ValidatePeerType` when `Stop()` is called before `Start()` (e.g., shutdown during initialization failure). The 30-second timeout would eventually unblock them anyway, so this is not strictly necessary — just faster cleanup on shutdown. Happy to remove if you think it's unnecessary.

consensus/istanbul/backend/engine.go
hyunsooda: I initially had the close in `SetChain`, but it caused Round Changes. Inside `engine.Start()`, the call order is `SetChain()` → `core.Start()` → `coreStarted = true`. If we close `chainInitCh` in `SetChain`, peers get unblocked and registered immediately. When these peers send consensus messages, [`HandleMsg`](https://github.com/kaiachain/kaia/blob/dev/consensus/istanbul/backend/handler.go#L65) checks `sb.coreStarted` and returns `ErrStoppedEngine` since the core has not started yet. This error propagates to `processConsensusMsg`, which terminates the peer's consensus goroutine — the peer is effectively disconnected from consensus and must wait for reconnection, causing a Round Change. Confirmed this on a local 4-node testnet. Moving the signal after `coreStarted = true` (via `defer`) eliminated the RC.

consensus/istanbul/backend/engine.go
hyunsooda: To clarify the scenario:

```
1. CN.Start()              ← P2P server starts, peers connect → block on <-chainInitCh
2. (initialization failure) ← something fails before StartMining() is called
3. Stop() called           ← engine.Start() was never called, so defer never ran
```

Without `SignalPeerRegistrable()` in `Stop()`: peer goroutines wait until the 30s timeout.
With it: immediate unblock → clean shutdown.

As mentioned, the 30s timeout covers this eventually, so happy to remove if preferred.

ian0371 (approved):

hyeonLewis (approved):

PR #803, opened 2026-03-20
[PR5] Permissionless foundation: types, ABv2, genesis alloc

hyunsooda:
## Proposed changes

- Add validator state types (Registered, CandReady, ValActive, etc.) and NodeStateMap
- Add ABv2 Go layer: InstallAddressBookV2, ReadABv2Timeouts, EncodeWriteNodes, ReadGetAllValidators, ReadAddressBookV2BlsAll
- Add ERC1967Proxy v4/v5 split for ABv2 proxy compatibility
- Add genesis allocation (AllocPermissionless, deployBeaconInfra, deployCnStakingFactory)
- Add shared test helpers (MakeTestPermissionlessConfig, MakeTestBlsKey)
- Add StateBlockchainContractCaller for statedb-based contract reads
- Update KIP113 to use ABv2 as BLS source for permissionless blocks
- Update multicall import to contracts_permissionless
- Fix mock address collision with MultiCallAddr (0x402 → 0x403+)

This PR contains only standalone functions and types — no wiring into engine/worker/execution. The state transition logic and integration are in PR6.

## Types of changes

- [x] 💥 Hardfork / consensus-breaking changes
- [x] 🧪 Test improvements
- [x] ♻️ Chore / Refactor / Non-functional changes

## Checklist

- [x] 📖 I have read the [CONTRIBUTING GUIDELINES](https://github.com/kaiachain/kaia/blob/main/CONTRIBUTING.md) doc
- [x] 🟢 Lint and unit tests pass locally with my changes (`$ make test`)

## Related issues

- #748

## Context

Part of PR split for #748. See `permissionless/PR_SPLIT_PLAN.md`.

**PR chain:** PR1 (#792) → PR2 (#796) → PR3 (#797) → PR4 (#802) → **PR5 (this)** → PR6

## Deferred to PR6

- `kaiax/valset/impl/` state transition logic (state_transition.go, getters, execution, schema)
- `kaiax/staking/` ConsolidatedNodes refactoring, staking_info changes
- `consensus/engine.go` hooks (WriteStatesToContract, InstallABv2)
- `node/cn/handler.go`, `params/config.go`, `cmd/homi/`, integration
- `state_transition_test.go`, `TestGetCouncilPermissionless`, `TestGetNodeByState`, `TestReadAddressBookV2BlsAll`

> Note: `kaiax/valset/README.md` and `kaiax/staking/README.md` include descriptions of state transition and integration features that are implemented in PR6. They are included here for reference but the corresponding code arrives in PR6.

PR #792, opened 2026-03-16
Permissionless-PR1: Contract updates (OZ v5, ABv2, CnStakingV4, Proxy)

hyunsooda:
## Summary

Introduce `contracts_permissionless/` as an independent Hardhat project for permissionless-related contracts, keeping `contracts/` fully intact.

### Why a separate project?

`contracts/` uses OpenZeppelin v4 with Hardhat 2.19 (`evmVersion: paris`). Permissionless contracts require OZ v5 and `evmVersion: cancun` (for `tload`/`tstore`/`mcopy` opcodes). Hardhat cannot apply `evmVersion` overrides to files inside `node_modules`, making it impossible to mix both within a single project without affecting existing contract bytecode.

### Changes

**`contracts/`** — No functional changes. Reverted all previously added permissionless-related files to base.

**`contracts_permissionless/`** — New independent Hardhat project:
- OZ v5 (`@openzeppelin/contracts@^5.0`, `@openzeppelin/contracts-upgradeable@^5.0`)
- `evmVersion: cancun` for permissionless contracts
- Solidity contracts: `AddressBookV2`, `CnStakingV4`, `CnStakingV4Factory`, `PublicDelegation`, `ERC1967Proxy`, `UpgradeableBeacon`, `MultiCallContract`
- `generate.go` for Go binding generation (uses `../contracts/abigenw`)
- Hardhat test: `test/MultiCall/multicall.test.ts`

**`contracts_permissionless/contracts/ABv2DataContract.sol`** — `epochBlockInterval` changed from constant to immutable (set via constructor)

PR #811, opened 2026-03-23
Permissionless: State transition + integration (PR6)

hyunsooda:
## Proposed changes

State transition engine and integration wiring for the permissionless validator lifecycle. This is PR6 of the permissionless PR chain (PR1~PR6).

### State transition engine (`valset/impl/state_transition.go`, `state_transition_getter.go`)
- `WriteStatesToContract`: called at `Initialize`, computes state diffs and writes changed nodes to ABv2 via system tx
- `InstallABv2`: installs and initializes ABv2 proxy at the permissionless fork parent block's `Finalize`
- `getEpochTransition`: applies validator state transitions at VRank epoch boundaries (CandReady→CandTesting, CandTesting→ValActive, etc.)
- `getTimeoutTransition`, `getViolationTransition`: timeout/violation-based state changes
- `getOrComputeNodeStates`: reads ABv2 state with caching, falls back to parent state + transition replay

### Permissionless council & committee (`valset/impl/getter.go`, `getter_council.go`, `getter_demote.go`)
- `GetCouncil`: permissionless path returns ValActive + ValPaused + ValReady
- `GetDemotedValidators`: permissionless path returns ValReady + ValPaused
- `GetCommittee`: permissionless path returns ValActive only
- `GetCandidates`: new API for candidate (CandReady + CandTesting) addresses

### Staking info for permissionless (`staking/impl/getter.go`)
- `MultiCallStakingInfoPermissionless`: reads from ABv2 via new MultiCall function
- `parsePermissionlessCallResult`: parses ABv2 profiles into StakingInfo with KEF/KIF/KPF fund addresses

### Consensus integration (`engine.go`, `handler.go`)
- `Initialize`: calls `WriteStatesToContract` for ABv2 state writes
- `Finalize`: installs ABv2 at permissionless fork parent block
- `verifyCascadingFields`: validates VRank header field (empty pre-fork, valid encoding post-fork)

### MultiCall compatibility fix
- Compile MultiCall with solc 0.8.19 (paris EVM) for pre-cancun fork compatibility
- Fix `DefaultRegistryConfig` KIP113 address (0x403→0x404) after mock address shift in #803

### Other changes
- `SystemTxCall`: new helper for system transaction execution within EVM
- `homi`: genesis support for permissionless config
- `SetPermissionlessForkBlock`: admin API for overriding the permissionless hardfork block number at runtime (for mainnet simulation)
- Randao getter: ABv2 storage root based cache for permissionless path
- Unit tests for state transitions, epoch/timeout/violation transitions, council getters

## Types of changes

- [ ] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

- [x] 🟢 Lint and unit tests pass locally with my changes

## Further comments

For detailed design of each module, refer to the README files in the respective directories (e.g., `kaiax/valset/README.md`, `kaiax/staking/README.md`).

## Related issues

- Part of permissionless PR chain: #792 (PR1) → #796 (PR2) → #797 (PR3) → #802 (PR4) → #803 (PR5) → this PR (PR6)
- Depends on #812 (MultiCall EVM compatibility fix)


cmd/homi/setup/cmd.go
```diff
@@ -546,6 +554,63 @@ func useAddressBookMock(ctx *cli.Context, genesisJson *blockchain.Genesis) {
 	allocationFunction(genesisJson)
 }
 
+func allocatePermissionless(ctx *cli.Context, genesisJson *blockchain.Genesis, validatorAddrs []common.Address, kip113Init system.AllocKip113Init) {
```
hyeonLewis, line 557: When we install ABv2 at genesis, I think we should set `isActivated` and `isConstructed` as true so that legacy getters (e.g., `getAllAddress`) can return data correctly.

hyunsooda: Initially I assumed legacy getters wouldn't need to work when permissionless starts from genesis, but I've fixed it (https://github.com/kaiachain/kaia/pull/811/changes/39a3b0463c494c971636ee6e13476289012df78c) now — `isActivated` is set to true during ABv2 installation

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,267 @@
...
+	DefaultValPausedTimeout     = time.Hour * 8
+	DefaultValIdleTimeout       = 30 * 24 * time.Hour
+	DefaultActiveValidatorCount = 50
+)
+
+// getOrComputeNodeStates returns the node states for block `num`.
+// If parentStatedb is provided, it is used as the parent state (avoids StateAt which may fail on pruned state).
+// 1. Cache hit → return cached value.
+// 2. Block N committed (header exists) → read ABv2(N) directly.
+// 3. Block N not committed or state pruned → read ABv2(N-1) + apply transitions.
```
ian0371: I think 2 != 3:

```
ABv2(N) = ABv2(N-1) + applyTr(N) + userTx(N)
- applyTr(N): VrankViolation, Timeout, Epoch transitions for engine.Initialize() at N
- userTx(N): non-epoch txs at block N (such as pause tx).

Path 2  [statedb=nil, N committed, state not pruned]:
  = ABv2(N)
  = ABv2(N-1) + applyTr(N) + userTx(N)                          ← includes userTx(N)

Path 3a [statedb=S passed in, during Initialize(N)]:
  S = committed end of N-1  →  S.ABv2 = ABv2(N-1)
  = ABv2(N-1) + applyTr(N)                                       ← misses userTx(N)

Path 3b [statedb=nil, state pruned]:
  = ABv2(N-1) + applyTr(N)                                       ← misses userTx(N)
```

So the result seems non-deterministic for the same `N`.

Given the callers `GetNodeByState`, `writeNodesToContract`, `getCouncilPermissionless`, I think `userTx(N)` should NOT be included. In other words, path 2 should be removed.

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,267 @@
...
+	num := header.Number.Uint64()
+	currentNodes, err := v.getOrComputeNodeStates(num, statedb)
+	if err != nil {
+		return err
+	}
+
+	// Compute diff: only nodes whose state or timeout changed.
+	// Error is ignored: if parent is unavailable (nil), diffNodeStates treats all current nodes as changed (fallback to full write).
+	parentNodes, _ := v.getOrComputeNodeStates(num-1, nil)
+	diff := diffNodeStates(parentNodes, currentNodes)
```
ian0371, line 162:
```
diff(N)
= ABv2(N) - ABv2(N-1)
= { ABv2(N-1) + applyTr(N) + userTx(N) } - ABv2(N-1)
= applyTr(N) + userTx(N)
```
However, `diff(N)` should be simply `applyTr(N)` without `userTx(N)`, right?
It'll make sense if `getOrComputeNodeStates(N) = ABv2(N-1) + applyTr(N)`.

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,267 @@
...
+	DefaultValIdleTimeout       = 30 * 24 * time.Hour
+	DefaultActiveValidatorCount = 50
+)
+
+// getOrComputeNodeStates returns the node states for block `num`.
+// If parentStatedb is provided, it is used as the parent state (avoids StateAt which may fail on pruned state).
+// 1. Cache hit → return cached value.
+// 2. Block N committed (header exists) → read ABv2(N) directly.
+// 3. Block N not committed or state pruned → read ABv2(N-1) + apply transitions.
+func (v *ValsetModule) getOrComputeNodeStates(num uint64, parentStatedb *state.StateDB) (valset.NodeStateMap, error) {
```
ian0371, line 44: How about simplifying the logic? `N-1` seems to be always committed.

```
// getNodeStates(N) = ABv2(N-1) + applyTr(N)
func (v *ValsetModule) getNodeStates(num uint64) (NodeStateMap, error) {
    // 1. return cache[N] if exists

    // 2. read ABv2(N-1) from committed chain state at num-1

    // 3. apply transitions for block N: applyTr(N) = VrankViolation + Timeout + Epoch

    // 4. store in cache[N] and return
}
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil, err
+	}
+	v.nodeStatesCache.Add(num, newValidators)
+	return newValidators, nil
+}
+
+// applyAllTransitions applies VrankViolation → Timeout → Epoch transitions.
+func (v *ValsetModule) applyAllTransitions(
+	validators valset.NodeStateMap,
+	num uint64,
```
ian0371: nit: `num` and `statedb` can be removed

```
num = header.Number.Uint64()
statedb = v.Chain.StateAt(header.Root)
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,267 @@
...
+	statedb *state.StateDB,
+	header *types.Header,
+) (valset.NodeStateMap, error) {
+	newValidators := v.getViolationTransition(num, validators)
+
+	backend, err := backends.NewStateBlockchainContractBackend(v.Chain, statedb.Copy())
```
ian0371: `Copy()` can be dropped — the statedb is never committed, so mutations from EVM reads are harmless.

```suggestion
	backend, err := backends.NewStateBlockchainContractBackend(v.Chain, statedb)
```

kaiax/valset/impl/state_transition.go
hyunsooda: Good catch. You're right — Path 2 reads committed ABv2(N) which includes userTx(N), while Path 3 computes ABv2(N-1) + applyTr(N) without userTx. This makes results non-deterministic depending on cache state.

I verified this on an archive node: after restart (cache cleared), `getNodeByState(N)` returned different results than before restart for the same block N.

Fixed by removing Path 2. Added unit tests using `GenerateChain` + `InsertChain` with a real pause tx to verify the fix: 44f32d97b

kaiax/valset/impl/state_transition.go
hyunsooda: Addressed in 44f32d97b — Path 2 removal ensures `getOrComputeNodeStates(N)` always returns `ABv2(N-1) + applyTr(N)`, so the diff is now purely `applyTr(N)`.

kaiax/valset/impl/state_transition.go
hyunsooda: Simplified in 44f32d97b. Kept `parentStatedb` parameter because `StateAt` can fail when trie nodes are not in stateCache (e.g., during historical regeneration). The parameter allows callers like `writeNodesToContract` to pass the already-available parent statedb.

kaiax/valset/impl/state_transition.go
hyunsooda: `num` cannot be derived from `header` — `header` is the parent header (N-1) while `num` is the current block number (N). `statedb` is kept for the same reason as comment 3: `StateAt` may fail during historical regeneration, so callers pass the already-available parent statedb.

kaiax/valset/impl/state_transition.go
hyunsooda: Agreed. `applyAllTransitions` only reads from statedb (`ReadABv2Timeouts`, `ReadABv2MaxCounts`) and never commits, so EVM read mutations are harmless. Removed in e221a59c4.

kaiax/valset/impl/state_transition.go
ian0371: I see we need statedb.
In that case, isn't num simply:
```
num = header.Number.Uint64() + 1
```

kaiax/valset/impl/state_transition.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+
+	// Compute diff: only nodes whose state or timeout changed.
+	// Error is ignored: if parent is unavailable (nil), diffNodeStates treats all current nodes as changed (fallback to full write).
+	parentNodes, _ := v.getOrComputeNodeStates(num-1, nil)
+	diff := diffNodeStates(parentNodes, currentNodes)
+
+	// Skip call if no changes and not an epoch block (epoch blocks need epochValCount update)
+	if len(diff) == 0 && !v.isVrankEpoch(num) {
```
ian0371, line 165: `applyTr(N)` contains VrankViolation and Timeout, which should be applied every block.
This condition should be removed.
```
if !v.isVrankEpoch(num)
```

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+	return ret
+}
+
+// getViolationTransition transitions ValActive validators to ValExiting when they violate rules:
+// rule1: staking amount dropped below MinimumStake, rule2: vrank violation (TODO).
+func (v *ValsetModule) getViolationTransition(num uint64, validators valset.NodeStateMap) valset.NodeStateMap {
+	var (
+		pset          = v.GovModule.GetParamSet(num - 1) // read gov param from parent number
```
ian0371: `GetParamSet(num)` is parameter set for generating block `num`.

```suggestion
		pset          = v.GovModule.GetParamSet(num)
```

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	if !v.isVrankEpoch(num) {
+		// return early if the `num` is not vrank epoch number
+		devLog(validators)
+		return validators.Copy()
+	}
+	var (
+		newValidators        = validators.Copy()
+		activeValCompetitors []sortableValidator
+		pset                 = v.GovModule.GetParamSet(num - 1) // read gov param from parent number
```
ian0371: `GetParamSet(num)` is parameter set for generating block `num`.

```suggestion
		pset                 = v.GovModule.GetParamSet(num)
```

cmd/homi/setup/cmd.go
```diff
@@ -744,11 +809,17 @@ func Gen(ctx *cli.Context) error {
...
+	// Permissionless includes Registry allocation and does not use KIP113
+	// (BLS keys are stored in ABv2DataContract instead).
+	if permissionlessBlock := ctx.Int64(permissionlessCompatibleBlockNumberFlag.Name); permissionlessBlock == 0 {
+		allocatePermissionless(ctx, genesisJson, validatorNodeAddrs, kip113Init)
+	} else {
+		// Randao hardfork related system contracts
+		kip113ProxyAddr, kip113LogicAddr := allocateKip113(ctx, genesisJson, kip113Init)
+		allocateRegistry(ctx, genesisJson, nodeAddrs[0], kip113ProxyAddr)
+		useKip113Mock(ctx, genesisJson, kip113LogicAddr)
+		useRegistryMock(ctx, genesisJson)
```
ian0371: Is `useRegistryMock` removed after permissionless?

consensus/istanbul/backend/engine.go
```diff
@@ -517,6 +517,12 @@ func (sb *backend) Initialize(chain consensus.ChainReader, header *types.Header,
 		vmenv := vm.NewEVM(context, vm.TxContext{}, state, chain.Config(), &vm.Config{})
 		blockchain.ProcessParentBlockHash(header, vmenv, state, chain.Config().Rules(header.Number))
 	}
+
+	context := blockchain.NewEVMBlockContext(header, chain, nil)
+	vmenv := vm.NewEVM(context, vm.TxContext{}, state, chain.Config(), &vm.Config{})
+	if err := sb.valsetModule.WriteStatesToContract(vmenv, header, state); err != nil {
+		logger.Error("Failed to process transition", "number", header.Number.Uint64(), "err", err.Error())
+	}
```
ian0371: To match the above format:

```suggestion
	if chain.Config().IsPermissionlessForkEnabled(header.Number) {
    	context := blockchain.NewEVMBlockContext(header, chain, nil)
    	vmenv := vm.NewEVM(context, vm.TxContext{}, state, chain.Config(), &vm.Config{})
    	if err := sb.valsetModule.WriteStatesToContract(vmenv, header, state); err != nil {
    		logger.Error("Failed to process transition", "number", header.Number.Uint64(), "err", err.Error())
    	}
	}
```

kaiax/valset/impl/state_transition.go
hyunsooda: Removed `num` param — now derived as `nextNum := header.Number.Uint64() + 1` inside the function. Fixed in 94554a4b0.

kaiax/valset/impl/getter_council.go
```diff
@@ -113,6 +126,17 @@ func lastNumLessThan(nums []uint64, num uint64) uint64 {
 	}
 }
 
+func lastNumLessEqualThan(nums []uint64, num uint64) uint64 {
```
ian0371: This function is unused

kaiax/valset/impl/state_transition_getter.go
hyunsooda: Fixed in 199c89f02.

kaiax/valset/impl/state_transition_getter.go
hyunsooda: Fixed in 199c89f02.

consensus/istanbul/backend/engine.go
hyunsooda: Agreed. Wrapped in `IsPermissionlessForkEnabled` check. Fixed in f94eb5f1d.

kaiax/valset/impl/getter_council.go
hyunsooda: Removed in f94eb5f1d.

kaiax/valset/impl/state_transition.go
hyunsooda: The `isVrankEpoch` check here controls whether to **write to the contract**, not whether to apply transitions. VrankViolation and Timeout transitions are already computed every block in `getOrComputeNodeStates`. This condition only skips the contract write when diff is empty AND not an epoch block. At epoch blocks, we always call `processSystemTransition` (even with empty diff) to update `epochValCount`.

cmd/homi/setup/cmd.go
hyunsooda: In the permissionless path, `AllocPermissionless` deploys the real Registry — no mocks are used. Should we use mocks here too for consistency?

kaiax/valset/impl/state_transition.go
ian0371: From https://kips.kaia.io/KIPs/kip-286, `ValPaused -> ValInactive` is Anytime + System tx.
So this transition is not written to the contract until the epoch block?

cmd/homi/setup/cmd.go
ian0371: I think so because using mock registry is orthogonal to permissionless.

kaiax/valset/impl/state_transition.go
hyunsooda: Timeout transitions (e.g., ValPaused → ValInactive) produce a non-empty diff when they fire, so they are written to the contract immediately — not deferred until the epoch block. The `isVrankEpoch` check only matters when `diff == 0`: it forces a write at epoch blocks to update `epochValCount`, even if no state changed.

ian0371: Oh right. sorry for the confusion

cmd/homi/setup/cmd.go
hyunsooda: Applied `useRegistryMock` to both paths. Fixed in 150083486.

hyunsooda:
@ian0371, https://github.com/kaiachain/kaia/pull/811/changes/22b35d8f140530ca19887ccd128f6139fa934e0f addressed the following changes:
1. All callers now explicitly pass parentStatedb — Removed the nil fallback (StateAt) inside
getOrComputeNodeStates. Callers (getCouncilPermissionless, GetNodeByState, postInsertBlockPermissionless,
writeNodesToContract) resolve the statedb themselves before calling.
2. Read parentNodes from committed ABv2(N-1) directly — writeNodesToContract now uses ReadGetAllValidators(statedb) instead of getOrComputeNodeStates(num-1, nil), so the diff is purely applyTr(N).
3. Extract computeNodeStates helper — Separated ReadGetAllValidators + applyAllTransitions into its own function for clarity.
4. Assert parentStatedb root matches block N-1 — Added IntermediateRoot check to catch mismatched statedb early.

ian0371 (approved):

PR #818, opened 2026-03-27
Permissionless: move VRankEpoch to ChainConfig as single source of truth

hyunsooda:
## Proposed changes

Move `VRankEpoch` from module-local variables to `ChainConfig`. Configurable via `genesis.json` only (no CLI flag by design). Defaults to 86400 in `SetDefaults()`.

- valset: remove `vrankEpoch` field, read from `chain.Config().VRankEpoch`
- vrank module update deferred until #794 is merged

To override the default, set `vrankEpoch` in genesis.json:
```json
{
  "config": {
    "vrankEpoch": 86400
  }
}
```

## Types of changes

- [ ] 🐛 Bug fix
- [ ] ✨ Non-hardfork changes (node upgrade not required)
- [ ] 💥 Hardfork / consensus-breaking changes
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool
- [x] ♻️ Chore / Refactor / Non-functional changes

## Checklist

- [x] 📖 I have read the [CONTRIBUTING GUIDELINES](https://github.com/kaiachain/kaia/blob/main/CONTRIBUTING.md) doc
- [x] 🟢 Lint and unit tests pass locally with my changes

## Related issues

- vrank module (#794) will be updated after merge

## Further comments

PR #819, opened 2026-03-27
vrank: skip catchup when head=zero

ian0371:
## Proposed changes

Skip catchup when head block number is zero.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/kaiax/vrank/impl/init.go
+++ b/kaiax/vrank/impl/init.go
@@ -157,6 +157,9 @@ func (v *VRankModule) catchUpScoreCaches() error {
 		return nil
 	}
 	headNum := head.Number.Uint64()
+	if headNum == 0 {
+		return nil
+	}
 
 	if _, err := v.GetPFS(headNum); err != nil {
 		return err
--- a/kaiax/vrank/impl/scoring.go
+++ b/kaiax/vrank/impl/scoring.go
@@ -120,6 +120,12 @@ func (v *VRankModule) lookupCFSSeed(blockNum uint64) (start uint64, seed vrank.C
 }
 
 func (v *VRankModule) newCPMatrix(blockNum uint64) (vrank.CPMatrix, error) {
+	if blockNum == 0 {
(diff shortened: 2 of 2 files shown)


hyunsooda:
I'm investigating the same issue and the root cause was found by "missing trie node error"

On node restart, catchUpScoreCaches(headNum) calls GetPFS(headNum), which replays from epoch start (block 0). If the historical state has been pruned, it fails with "missing trie node", causing Init to return an error. Since SetupKaiaxModules logs the error but continues, kaiax modules remain nil, leading to a nil pointer panic in consensus.

The catchUpScoreCaches failure is not fatal — the node can function with cold caches. The error should be treated as a warning instead of propagating it.

Specifically, On restart, catchUpScoreCaches replays from epoch start. The GetCFS path calls GetCandidates → GetNodeByState →StateAt(parentRoot), which requires historical state. If that state has been pruned, it fails with "missing trie node".

Reproduction steps:
1. Start a network with permissionless fork at block 0
2. Let it produce several blocks, then stop all nodes
3. Restart nodes → "missing trie node" error → nil pointer panic

The fix is here at https://github.com/kaiachain/kaia/commit/3d77ac1027d382bbb03bfc20e46a498ac3041b6e#diff-39668c23e0877af4d2f01512c8ea65c47c016960f3957609b8aa6793008fc8f0R117

ian0371:
@hyunsooda Actually, my problem was this:
```
GetCandidates(0) → GetNodeByState(0) → GetHeaderByNumber(-1)
```

This issue seems to have been addressed by 818. I'll close this issue.

hyunsooda:
Ah, yes, that was the second problem. Here at: https://github.com/kaiachain/kaia/pull/818/changes/3d77ac1027d382bbb03bfc20e46a498ac3041b6e#diff-21293fdf1268c75a1c49f08454d3126e629a0e35501ace4d08a4188ad7f05c2dR130-R143

PR #834, opened 2026-04-01
consensus: migrate seal consumers to consensus.Sealer

yoomee1313:
## Proposed changes

This PR is the second in a three-PR series for the sealer refactor.

It migrates seal-related consumers from `consensus.Engine` to `consensus.Sealer`, so seal parsing and consensus-aware hash handling are no longer tied to the live engine interface. It also exposes `ValidateHeader` on `consensus.ChainReader` and updates Istanbul verification to use the chain-provided validation path directly.

The CN runtime wiring through the BFT facade is handled in the follow-up PR.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist


api/api_kaia_blockchain.go
```diff
@@ -1001,10 +1074,15 @@ func (s *KaiaBlockChainAPI) makeRPCBlockOutputWithConsensusInfo(b *types.Block,
 
 	r["committee"] = cInfo.Committee
 	r["committers"] = cInfo.Committers
+	r["headerHash"] = cInfo.HeaderHash
```
hyeonLewis: Is it intended to add those fields?

blockchain/block_validator.go
```diff
@@ -124,13 +149,10 @@ func (v *BlockValidator) Preprocess(headers []*types.Header) (chan<- struct{}, <
...
 			} else {
-				if header.Number == nil {
+				if header == nil || header.Number == nil {
 					err = consensus.ErrUnknownBlock
 				} else if header.Number.Uint64() != 0 {
-					_, err = v.engine.Author(header)
-					if err == nil {
-						_, err = v.engine.Committers(header)
-					}
+					err = v.preprocessSeals(header)
```
hyeonLewis: It changes the validation logic (e.g., `ErrEmptyCommittedSeals`). I think it's safe for most of cases but could you confirm if it's safe?

blockchain/blockchain.go
```diff
@@ -234,7 +238,7 @@ type prefetchTx struct {
 // NewBlockChain returns a fully initialised block chain using information
 // available in the database. It initialises the default Kaia validator and
 // Processor.
-func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config) (*BlockChain, error) {
+func NewBlockChain(db database.DBManager, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, consensusImpl interface{}, vmConfig vm.Config) (*BlockChain, error) {
```
hyeonLewis: For me, it adds ambiguity. I think receiving explicit type (sealer, engine) would be better, and removing the logic below would be better. Same for `NewHeaderChain`. Hdyt?

blockchain/block_validator.go
```diff
@@ -74,6 +85,7 @@ func NewBlockValidator(config *params.ChainConfig, blockchain *BlockChain) *Bloc
 		bc:     blockchain,
 		hc:     blockchain,
 		engine: blockchain.Engine(),
+		sealer: blockchain.sealer,
```
hyeonLewis, line 85: We can check whether the sealer is non-nil here and remove the pervasive nil check. Do you expect there can be a nil sealer case?

blockchain/block_validator.go
```diff
@@ -229,6 +283,80 @@ func (v *BlockValidator) validateHeader(header *types.Header, parent *types.Head
...
+	blockNum := header.Number.Uint64()
+
+	var qualified []common.Address
+	if v.mValset != nil {
+		qualified, err = v.mValset.GetQualifiedValidators(blockNum)
+		if err != nil {
+			return err
+		}
+	} else {
+		qualified, err = v.sealer.Validators(header)
```
hyeonLewis: This can't be used since we don't make a consensus on the validator list in the extra field.

hyeonLewis (commented):
I see many places that use the engine and sealer together (use the engine if the sealer is nil). Will it be handled in the next PR?

api/api_kaia_blockchain.go
yoomee1313: Oh it shouldn't be added. Thanks for catching it!

blockchain/blockchain.go
yoomee1313: Agreed. I'd changed the code to receive sealer explicitly.

blockchain/block_validator.go
yoomee1313: Since some tests expect nil, we can't add a non-nil condition.

hyeonLewis: > Since some tests expect nil

Do you mean there's a case where `sealer` should be nil? Or just for testing (passing nil argument for testing)?

If it's the latter, I think we should add a nil check here and use a non-nil mock sealer in the testing.

yoomee1313: Oh sorry, you mean sealer.. I misunderstood it. There's no such case that expects sealer nil. I added the nil check.
https://github.com/kaiachain/kaia/pull/834/commits/a50208d076750e4b6a0091e37a026e0ce6a698d6

blockchain/block_validator.go
yoomee1313: Yes it's safe. The intent behind handling `empty committed seals` is the same.

blockchain/block_validator.go
yoomee1313: Yes, right. Removed that part. The code assumes the valset, gov is always exists.

blockchain/block_validator.go
```diff
@@ -62,55 +64,40 @@ type BlockValidator struct {
...
-	if v.config == nil || v.config.Governance == nil || v.config.Governance.KIP71 == nil {
-		return nil, errors.New("missing KIP-71 config for magma header verification")
-	}
-	return v.config.Governance.KIP71, nil
+func (v *BlockValidator) ValsetModule() valset.ValsetModule {
+	return v.mValset
 }
 
 func (v *BlockValidator) ValidateHeader(header *types.Header) error {
-	if header.Number == nil {
```
hyeonLewis, line 110: Could we remove nil check here?

blockchain/block_validator.go
```diff
@@ -204,19 +192,19 @@ func (v *BlockValidator) validateHeader(header *types.Header, parent *types.Head
...
 	// Verify Magma basefee rule from governance paramset.
 	if v.config.IsMagmaForkEnabled(header.Number) {
-		kip71Config, err := v.kip71Config(header.Number.Uint64())
-		if err != nil {
-			return err
-		}
-		if err := kip71Config.VerifyMagmaHeader(header.BaseFee, parent.Number, parent.BaseFee, parent.GasUsed); err != nil {
-			return err
+		// Skip governance-dependent validation when gov module is not registered.
+		if v.mGov != nil {
```
hyeonLewis, line 196: Is it safe to skip the validation at all? It seems we had fallback previously.

blockchain/block_validator.go
hyeonLewis: Ahh, sorry, since the `NewBlockValidator` is called by `NewBlockchain` and it already checked the nil sealer, so we don't need it here.

blockchain/block_validator.go
hyeonLewis: Also for below `verifySeals`.

blockchain/block_validator.go
yoomee1313: It's okay, we can leave it since test can call the `NewBlockValidator` seperately.

blockchain/block_validator.go
yoomee1313: Yes, the `header.Number` is accessed before calling the ValidateHeader every time.

blockchain/block_validator.go
yoomee1313: We assume gov and valset module is always set, so it's safe. It would be best if we can set gov and valset at NewBlockValidator, however, it's not possible now.

hyeonLewis: Yes, I also thought that. I think we should leave it in the backlog to fix the fragile setup.

hyeonLewis (approved):

PR #835, opened 2026-04-01
consensus: route CN runtime through the BFT facade

yoomee1313:
## Proposed changes

This is the third PR in the three-PR sealer refactor series.

It routes CN runtime construction through the `consensus/bft` facade and removes the remaining legacy engine-facing seal paths.

Detailed changes
* Cleaned up the remaining consensus engine construction path after the sealer refactor.
* Updated CN initialization to use bft.NewEngine() explicitly instead of constructing the Istanbul backend directly.
* Narrowed the protocol manager dependency from the full consensus engine to consensus.Handler, matching the actual interface it uses.
* Removed unused DB and ChainConfig dependencies from bft.Opts and istanbul/backend.BackendOpts.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist


consensus/bft/engine.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Opts bundles the inputs needed to instantiate a consensus engine.
+type Opts struct {
+	IstanbulConfig *istanbul.Config
+	PrivateKey     *ecdsa.PrivateKey
+	NodeType       common.ConnType
+}
+
+// NewEngine creates the consensus engine selected by the given configuration.
+func NewEngine(opts *Opts) consensus.Engine {
```
hyeonLewis, line 20: How about adding defensive nil check here?

cmd/utils/nodecmd/util.go
```diff
@@ -163,61 +162,64 @@ func decodeGov(bytes []byte) (map[string]interface{}, error) {
...
+		committers = []common.Address{}
+	}
+
+	headerHash := s.HeaderHash(header)
+	sigHash := s.SigHash(header)
+
+	result := make(map[string]interface{})
+	result["hash"] = headerHash.Hex()
+	result["sigHash"] = sigHash.Hex()
+	result["validators"] = validators
```
hyeonLewis, line 218: How about using the same output format for `committedSeal` and `validators`?

blockchain/chain_makers.go
```diff
@@ -297,43 +287,6 @@ func GenerateChain(config *params.ChainConfig, parent *types.Block, sealer conse
 	return blocks, receipts
 }
 
-func generateChainSigner(config *params.ChainConfig) (consensus.Sealer, common.Address) {
```
hyeonLewis, line 300: Is it correct that we don't need to seal for `chain_makers` (fake sealer for testing only)?

consensus/bft/engine.go
yoomee1313: Added at here https://github.com/kaiachain/kaia/pull/835/commits/d216ae00f6185c5e12810285138ee761ac078a1e

cmd/utils/nodecmd/util.go
yoomee1313: Changed `committedSeal` output format as hex format so that both are returned as hex format. https://github.com/kaiachain/kaia/pull/835/commits/d216ae00f6185c5e12810285138ee761ac078a1e

blockchain/chain_makers.go
yoomee1313: No, we don't need sealing. `GenerateChain` takes a sealer as a parameter, but in current usage it is mostly called with faker-based sealers, so block generation works fine without actual sealing. In the dev branch, the `GenerateChain` path did not perform actual sealing also.

hyeonLewis (approved):

PR #842, opened 2026-04-06
permless: read slot factor from ABv2 for Byzantine filtering

ian0371:
## Proposed changes

- rename `epochValCount` to `slotFactor`
- add slotFactor to MultiCall

vrank changes
- `getSlotFactorAt(blockNum)` reads SlotFactor from ABv2 state at the given block via `ReadNodeStates`. This requires
StateAt and a full EVM context, so the chain interface is widened to embed `backends.BlockChainForCaller` instead of the
 two-method minimal interface.
- `GetCFS(blockNum)` now calls `getSlotFactorAt(blockNum)` first, then delegates to `getCFS(blockNum, slotFactor)`.
- `GetCFSWithSlotFactor(blockNum, slotFactor uint64)` is added to avoid missing-trie-node error from `Initialize()` epoch transition path.
- `getCPMatrix(blockNum)` is extracted as a standalone helper that computes and caches the cpMatrix without needing a
slotFactor. `catchUpScoreCaches` and `PostInsertBlock` use it directly, so cache warm-up succeeds even when ABv2 state is unavailable.
- `getCFS` is now a two-liner: `getCPMatrix` + `generateCFSFromCPMatrix`.
- `generateCFSFromCPMatrix` is now a pure function.

valset changes
- `computeNodeStates` passes `res.SlotFactor` through applyAllTransitions → getEpochTransition → isPassVrankTest.
- `isPassVrankTest` calls `GetCFSWithSlotFactor()` instead of `GetCFS` to avoid missing-trie-node error.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist


kaiax/valset/impl/state_transition_getter_test.go
```diff
@@ -255,18 +255,42 @@ func TestGetEpochTransition_CandTestingCFS(t *testing.T) {
...
-			result := v.getEpochTransition(testMinStake, validators, testIdleTimeout, testMaxValCount, testBlockTime, testCFSBlockNum, testCFSThreshold)
+			result := v.getEpochTransition(testMinStake, validators, testIdleTimeout, testMaxValCount, testBlockTime, testCFSBlockNum, testCFSThreshold, 0)
 			assert.Equal(t, tc.expectedState, result[addr1].State)
 		})
 	}
 }
 
+// TestGetEpochTransition_SlotFactorForwarded verifies that slotFactor is forwarded as-is
+// to GetCFSWithSlotFactor — not re-read from chain.
+func TestGetEpochTransition_SlotFactorForwarded(t *testing.T) {
```
hyunsooda: This test seems redundant — `TestGetEpochTransition_CandTestingCFS` already calls `getEpochTransition` with a specific `slotFactor` and the mock verifies `GetCFSWithSlotFactor` is called. The forwarding is implicitly covered there. Also, it makes removal of `testChain` extension for that interfcae in teh testing(https://github.com/kaiachain/kaia/pull/842/changes?diff=unified#diff-d9cb1f3a4d4524815c189ced4bd1b99efce99b398bd6832175a03f187fb704efR60).

kaiax/vrank/impl/getter_test.go
```diff
@@ -53,10 +57,36 @@ func (c *testChain) GetHeaderByNumber(number uint64) *types.Header {
 	return c.headers[number]
 }
 
+func (c *testChain) Config() *params.ChainConfig {
```
hyunsooda: Use mock?

kaiax/vrank/impl/scoring_test.go
```diff
@@ -1039,15 +1016,44 @@ func TestApplyBlocksForCFS(t *testing.T) {
...
-		cfs, err := computeCFS(v, 0, 0)
+		cfs, err := computeCFS(v, 0, 0, 1)
 		require.NoError(t, err)
 		// C1 is pre-seeded from GetCandidates but no block reported any failure, so CFS=0.
 		score, hasC1 := cfs[C1]
 		assert.True(t, hasC1, "C1 should appear in CFS map (pre-seeded from candidates)")
 		assert.Equal(t, uint64(0), score, "C1 CFS should be 0 (no failures reported)")
 	})
+
+	t.Run("uses_epoch_val_count_snapshot_instead_of_live_committee_len", func(t *testing.T) {
```
hyunsooda: `uses_epoch_val_count_snapshot_instead_of_live_committee_len` seems to test the same logic as the existing `computeCFS(v, 5, 9, 4)` case above — both verify that slotFactor determines F for byzantine filtering. The only difference is the F value (1 vs 2). Also, the test name still references epoch_val_count instead of slotFactor.

kaiax/vrank/impl/scoring.go
```diff
@@ -57,12 +58,42 @@ func (v *VRankModule) GetPFS(blockNum uint64) (map[common.Address]uint64, error)
...
+func (v *VRankModule) getSlotFactorAt(blockNum uint64) (uint64, error) {
+	header := v.Chain.GetHeaderByNumber(blockNum)
+	if header == nil {
+		return 0, vrank.ErrHeaderNotFound
+	}
+	statedb, err := v.Chain.StateAt(header.Root)
+	if err != nil {
+		return 0, err
+	}
+	res, err := system.ReadNodeStates(statedb, v.Chain, header)
```
hyunsooda: `getSlotFactorAt` calls `system.ReadNodeStates` which executes a full multicall (profiles, staking amounts, timeouts, thresholds, slot limits) just to read `SlotFactor`. Consider using `ABv2.GetSlotFactor()` directly via the existing binding — it's a single view call, avoids the overhead of the full multicall, and would remove the need to widen the chain interface to BlockChainForCaller.

kaiax/valset/impl/state_transition.go
```diff
@@ -93,15 +93,15 @@ func (v *ValsetModule) applyAllTransitions(
...
 }
 
 // isPassVrankTest returns true if the candidate's CFS is below the threshold.
 // CFS < cfsThreshold → pass (CandTesting → ValActive promotion eligible).
-func (v *ValsetModule) isPassVrankTest(addr common.Address, num, cfsThreshold uint64) bool {
-	cfsScores, err := v.VRankModule.GetCFS(num)
+func (v *ValsetModule) isPassVrankTest(addr common.Address, num, cfsThreshold, slotFactor uint64) bool {
+	cfsScores, err := v.VRankModule.GetCFSWithSlotFactor(num, slotFactor)
 	if err != nil { // CFS unavailable → pass to avoid blocking promotion
 		logger.Warn("isPassVrankTest: GetCFS failed", "num", num, "err", err)
```
hyunsooda: nit: Could you change the related logs? `GetCFS -> GetCFSWithSlotFactor`

kaiax/valset/impl/state_transition_getter_test.go
ian0371: Right. fixed 26804e8c2

kaiax/vrank/impl/scoring_test.go
ian0371: Right. fixed 26804e8c2

kaiax/vrank/impl/scoring.go
ian0371: fixed 70a1b3ce2

kaiax/valset/impl/state_transition.go
ian0371: fixed 93de080e5

hyunsooda (approved):
Let me sync the contract changes into system-contracts once it's merged.

PR #847, opened 2026-04-07
permless: fallback when `len(ValActive)` becomes zero

ian0371:
## Proposed changes

If all ValActive are below stake, `len(ValActive) = 0`. In this case, remove stake filter (minStake check) for collecting `activeValCompetitors`.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/kaiax/valset/impl/state_transition_getter.go
+++ b/kaiax/valset/impl/state_transition_getter.go
@@ -93,24 +93,26 @@ func (v *ValsetModule) getEpochTransition(
 			potentialActiveVal.IdleTimeout = now.Add(idleTimeout)
 		}
 	}
+	return newValidators
+}
 
-	// Fallback: if no validator ended up ValActive (e.g., all top competitors were ValPaused),
-	// promote the entire staking-competition group to ValActive to preserve len(ValActive)>0.
-	// The group is guaranteed non-empty because intra-epoch checks ensure len(ValActive)>=1 at epoch-1.
-	hasActive := slices.ContainsFunc(activeValCompetitors, func(sv sortableValidator) bool {
-		// Optimization: check `activeValCompetitors` instead of `newValidators` — after the loop above,
-		// any ValActive in `newValidators` must have come through `activeValCompetitors`.
-		return sv.State == valset.ValActive
-	})
-	if !hasActive {
-		logger.Warn("All top competitors are ValPaused; force-promoting to ValActive", "count", min(maxValidatorCount, len(activeValCompetitors)))
-		for idx, sv := range activeValCompetitors {
-			if idx >= maxValidatorCount {
-				break
+// getFallbackTransition is the last-resort committee fallback when getEpochTransition produces no ValActive.
+// It promotes epoch-1's staking competition group members (CandTesting/ValReady/ValActive/ValPaused)
+// to ValActive regardless of stake. CandTesting validators are still subject to the vrank test.
+func (v *ValsetModule) getFallbackTransition(validators valset.NodeStateMap, num, cfsThreshold, slotFactor uint64) valset.NodeStateMap {
+	newValidators := validators.Copy()
+	for addr, val := range newValidators {
+		switch val.State {
+		case valset.CandTesting:
+			if v.isPassVrankTest(addr, num, cfsThreshold, slotFactor) {
+				val.State = valset.ValActive
+				val.IdleTimeout = time.Time{}
+				val.PausedTimeout = time.Time{}
 			}
-			sv.State = valset.ValActive
-			sv.IdleTimeout = time.Time{}
-			sv.PausedTimeout = time.Time{}
+		case valset.ValReady, valset.ValActive, valset.ValPaused:
+			val.State = valset.ValActive
+			val.IdleTimeout = time.Time{}
+			val.PausedTimeout = time.Time{}
 		}
 	}
 	return newValidators
--- a/kaiax/valset/impl/state_transition_getter.go
+++ b/kaiax/valset/impl/state_transition_getter.go
@@ -54,6 +54,29 @@ func (v *ValsetModule) getEpochTransition(
 	now time.Time,
 	num, cfsThreshold, slotFactor uint64,
 ) valset.NodeStateMap {
+	result := v.doEpochTransition(validators, minStake, idleTimeout, maxValidatorCount, now, num, cfsThreshold, slotFactor, false)
+	for _, val := range result {
+		if val != nil && val.State == valset.ValActive {
+			return result
+		}
+	}
+
+	logger.Warn("Epoch transition produced no ValActive; retrying with staking filter lifted", "num", num)
+	return v.doEpochTransition(validators, minStake, idleTimeout, maxValidatorCount, now, num, cfsThreshold, slotFactor, true)
+}
+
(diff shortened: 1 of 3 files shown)


kaiax/valset/impl/state_transition_getter.go
```diff
@@ -93,6 +93,26 @@ func (v *ValsetModule) getEpochTransition(
 			potentialActiveVal.IdleTimeout = now.Add(idleTimeout)
 		}
 	}
+
+	// Fallback: if no validator ended up ValActive (e.g., all top competitors were ValPaused),
+	// promote the entire staking-competition group to ValActive to preserve len(ValActive)>0.
+	// The group is guaranteed non-empty because intra-epoch checks ensure len(ValActive)>=1 at epoch-1.
+	hasActive := slices.ContainsFunc(activeValCompetitors, func(sv sortableValidator) bool {
```
hyunsooda: Slot limits cap ValPaused count, so all-ValPaused competitors is unreachable. The only trigger is empty `activeValCompetitors` (all below minStake), which makes the loop a no-op. Consider broadening to all CandTesting/ValReady/ValActive/ValPaused validators regardless of stake.

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -93,6 +93,26 @@ func (v *ValsetModule) getEpochTransition(
...
+		// Optimization: check `activeValCompetitors` instead of `newValidators` — after the loop above,
+		// any ValActive in `newValidators` must have come through `activeValCompetitors`.
+		return sv.State == valset.ValActive
+	})
+	if !hasActive {
+		logger.Warn("All top competitors are ValPaused; force-promoting to ValActive", "count", min(maxValidatorCount, len(activeValCompetitors)))
```
hyunsooda: Incorrect log: Unreachable — slot limits prevent all-ValPaused. Actual trigger is empty competitors.

kaiax/valset/impl/state_transition_getter_test.go
```diff
@@ -113,7 +113,7 @@ func TestGetEpochTransition_StateTransitions(t *testing.T) {
 		{"T3b: CandTesting + stake<min → ValInactive", valset.CandTesting, belowMinStake, valset.ValInactive, true},
 		{"ValReady + stake>=min → ValActive", valset.ValReady, aboveMinStake, valset.ValActive, false},
 		{"ValActive + stake>=min → ValActive", valset.ValActive, aboveMinStake, valset.ValActive, false},
-		{"ValPaused + stake>=min → ValPaused (preserved)", valset.ValPaused, aboveMinStake, valset.ValPaused, false},
+		{"ValPaused + stake>=min → ValActive (fallback)", valset.ValPaused, aboveMinStake, valset.ValActive, false},
```
hyunsooda: This change removed the "ValPaused preserved" coverage. Consider adding a new test with ValPaused + another ValActive validator to verify ValPaused is preserved when fallback doesn't trigger.

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -93,6 +93,26 @@ func (v *ValsetModule) getEpochTransition(
...
+		for idx, sv := range activeValCompetitors {
+			if idx >= maxValidatorCount {
```
hyunsooda: nit: No test covers the `idx >= maxValidatorCount` branch in the fallback loop.

kaiax/valset/impl/state_transition_getter.go
ian0371: Right. Actually the previous did not implement "fallback to epoch-1's competitors", which implies that there's no violation/timeout/transition at `epoch` block, only fallback. So moved fallback logic to `applyAllTransitions`.

kaiax/valset/impl/state_transition_getter.go
ian0371: Actually the previous did not implement "fallback to epoch-1's competitors", so moved fallback logic to `applyAllTransitions`.

kaiax/valset/impl/state_transition_getter_test.go
ian0371: Restored this test.

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -96,6 +96,40 @@ func (v *ValsetModule) getEpochTransition(
...
+	}
+	slices.SortFunc(activeValCompetitors, func(a, b sortableValidator) int {
+		return cmp.Or(
+			cmp.Compare(b.StakingAmount, a.StakingAmount),
+			bytes.Compare(a.addr[:], b.addr[:]), // tie-breaking: address order
+		)
+	})
+	for idx, potentialActiveVal := range activeValCompetitors {
+		if idx >= maxValidatorCount {
+			break
```
hyunsooda: I think validators beyond `maxValidatorCount` should be demoted to `ValInactive` with `IdleTimeout`, same as `getEpochTransition` does. Keeping the two functions semantically aligned would prevent subtle state divergence. What do you think?

kaiax/valset/impl/state_transition_getter.go
```diff
@@ -96,6 +96,40 @@ func (v *ValsetModule) getEpochTransition(
...
+		switch val.State {
+		case valset.CandTesting:
+			if v.isPassVrankTest(addr, num, cfsThreshold, slotFactor) {
+				activeValCompetitors = append(activeValCompetitors, sortableValidator{addr, val})
+			}
+		case valset.ValReady, valset.ValActive, valset.ValPaused:
+			activeValCompetitors = append(activeValCompetitors, sortableValidator{addr, val})
+		}
+	}
+	slices.SortFunc(activeValCompetitors, func(a, b sortableValidator) int {
```
hyunsooda: The staking competition sort (stake desc, address asc) is now in both `getEpochTransition` and `getFallbackTransition`. Would it make sense to extract a shared helper? That way future sort criteria changes only need one update.

kaiax/valset/impl/state_transition_getter.go
ian0371: It's required for proper capping. fixed 8b5fb13af

kaiax/valset/impl/state_transition_getter.go
ian0371: fixed 8b5fb13af

hyunsooda (approved):

ian0371:
@hyunsooda
PTAL again. Logic changed. This time, only the stake filter is ignored at fallback.
Previously, ValPaused was promoted to ValActive as a side effect. Now ValPaused stays the same.

Please note that if we implement your suggestion to move stake filter to violation, we may not need this PR at all.

Also, I found out `getViolationTransition` is non-deterministic; it depends on the order of map iteration. It randomly makes addr1/addr2 ValInactive.
https://github.com/kaiachain/kaia/blob/54451be3df99998f202f548f21b0a4f506ec166c/kaiax/valset/impl/state_transition_getter_test.go?plain=1#L755-L756

Can be confirmed by:
```
go test ./kaiax/valset/... -run=TestApplyAllTransitions -count=10
```

ian0371:
@hyunsooda should I merge it or close it?

hyunsooda:
Please close it. My following PR will cover the scope of this PR.

PR #860, opened 2026-04-14
Permissionless: reformulate slot factor

hyunsooda:
## Proposed changes

- **SF = VA only**: Redefine Slot Factor as `len(ValActive)` only, removing ValPaused.
  - CFS Byzantine filter F=(SF-1)/3 counts proposer faults; only ValActive can propose, so including VP inflates F
  - Removes feedback loop where VA→VP transition doesn't reduce SF
- **SlotMath ceil(2n/3)**: Remove `f(n)` and `budgetMin`. All formulas derive from a single BFT threshold `ceil(2n/3)`.
  - `minActiveCount(n) = (2n+2)/3`
  - `maxSlotAvailable(n) = (n/3+1)/2`
- **n<4 guard in SlotMath**: For n < 4, BFT f=0 requires all nodes active.
  `minActiveCount(n<4) = n`, `maxSlotAvailable(n<4) = 0`
- **epochSlotFactor from core**: `processSystemTransition` now receives `epochSF` (VA count after epoch, before violation) from core and stores it at epoch blocks. Contract no longer self-computes slotFactor; core is the single source of truth.
- **Test expansion**: Extend n=10 TC to 4 violators, proving individual slot limits (maxSlot=2) and minActive budget cap work together

## Types of changes

- [ ] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

- [x] 📖 I have read the CONTRIBUTING GUIDELINES doc
- [x] 📝 I have signed in the PR comment
- [ ] 🟢 Lint and unit tests pass locally with my changes ($ make test)

## Further comments

After this PR is merged, kaiachain/kips#107 should be merged to update the KIP-286 spec accordingly.


contracts_permissionless/contracts/AddressBookV2/NodeActions.sol
```diff
@@ -201,7 +201,7 @@ abstract contract NodeActions is AddressBookV2Base {
             _batchTransition(nodeIds, newStates, timeoutAts);
         }
         if (_isEpochBlock()) {
-            uint256 slotFactor = _getStateCount(State.ValActive) + _getStateCount(State.ValPaused);
+            uint256 slotFactor = _getStateCount(State.ValActive);
```
ian0371: slotFactor definition is different from core and contract.

Core: slotFactor before violation transition (i.e., after epoch transition)
Contract: slotFactor after violation transition (i.e., after epoch-violation-timeout transition)

Does this discrepancy only affect violation transition at epoch block (no2 below)?
1. e(k*epoch) <-- defines newSF
2. v(k*epoch) <-- uses newSF
3. Contract redefines slotFactor and stores. This value may be different from newSF above.
4. v(k*epoch+1) <-- uses slotFactor stored in contract

contracts_permissionless/contracts/libraries/SlotMath.sol
```diff
@@ -2,39 +2,23 @@
...
-    /// @notice Returns the maximum number of total changeable slots per epoch
-    function f(uint256 n) internal pure returns (uint256) {
-        if (n == 0) return 0;
-        return (n - 1) / 3;
+    /// @notice Returns the minimum number of active validators required for BFT liveness
+    /// @dev ceil(2n/3) = (2n + 2) / 3 in integer arithmetic.
+    /// @param n The epoch validator count (ValActive count at epoch start)
+    /// @return The minimum active validator count
+    function minActiveCount(uint256 n) internal pure returns (uint256) {
+        return (2 * n + 2) / 3;
```
ian0371, line 23:
```
if n < 4: return n
```
same as RequiredMessageCount.

contracts_permissionless/contracts/libraries/SlotMath.sol
```diff
@@ -2,39 +2,23 @@
...
     }
 
     /// @notice Returns the maximum number of validators allowed in ValPaused or ValExiting state each
-    /// @dev Returns max(1, F/2) where F is the maximum number of total changeable slots per epoch
-    ///      N = epoch validator count (VA + VP at epoch start). Returns 0 if N < 2.
-    /// @param n The number of epoch validators
+    /// @dev ceil(floor(n/3) / 2) = (n/3 + 1) / 2 in integer arithmetic.
```
ian0371: Is this equal to `ceil((N - minActive) / 2)`?

ian0371:
Also is there `count(ValPaused) + count(ValExiting) <= SF - minActive` check?

contracts_permissionless/contracts/AddressBookV2/NodeActions.sol
hyunsooda: Thanks for the detailed analysis. The issue: at an epoch block, core computes `epochSF` (VA count after epoch, before violation) and uses it as the slot budget for violation. The contract, however, received the final batch transitions and inferred `slotFactor` from the post-all-transitions VA count — which is `epochSF - (demoted by violation)` when violation fires at the same epoch block.

This means user transactions in the rest of the epoch read a lower `slotFactor` from storage than what the epoch's violation transition actually used as the budget.

Fixed by passing `epochSlotFactor` explicitly: `processSystemTransition` now receives `epochSF` from core and stores it at epoch blocks. Core computes the SF, contract only stores it.

contracts_permissionless/contracts/libraries/SlotMath.sol
hyunsooda: Fixed. Added `if (n < 4) return n / return 0`.

contracts_permissionless/contracts/libraries/SlotMath.sol
hyunsooda: Yes. `n - minActive = n - ceil(2n/3) = floor(n/3)`, so `ceil((n - minActive) / 2) = ceil(floor(n/3) / 2) = (n/3 + 1) / 2`.

hyunsooda:
No explicit combined check, but `paused + exiting <= totalBudget(SF - minActive)` is enforced by the slot constraint on both user and system paths.

contracts_permissionless/contracts/libraries/SlotMath.sol
ian0371: I see. can you add in the comment that the original formula is `n - minActive` and the result is an optimization?

hyunsooda: Done. https://github.com/kaiachain/kaia/pull/860/changes/f43cbec502f20cee5571bab58874a9b7358d504a

ian0371 (approved):

PR #863, opened 2026-04-14
kaiabft: add KaiaBFT consensus engine with speculative execution

hyeonLewis:
## Proposed changes

- Introduce the consensus/kaiabft package — a BFT consensus engine that speculatively executes proposed blocks      
during pre-prepare, populating a cache that InsertChain uses to skip re-execution
- Add --consensus.engine CLI flag to select between istanbul (default) and kaiabft                                  
- Route consensus/engine.NewEngine() to the correct engine based on the flag

## What is KaiaBFT?                                                                                                 
                                         
KaiaBFT is a **wire-compatible enhancement** of Istanbul BFT. It uses the same message format, same header          
extra-data layout, and the same PBFT consensus protocol (pre-prepare → prepare → commit → round-change). Istanbul
and KaiaBFT nodes can reach consensus together in a mixed network — a KaiaBFT node is "just a fast node."           
                                        
The key behavioral change is **speculative execution on the validator side**:                                       
  
| | Istanbul BFT | KaiaBFT |                                                                                        
|---|---|---|                          
| **Validator receives pre-prepare** | Accept proposal, send prepare | Accept proposal, send prepare, **also start background execution** |
| **Consensus completes (commit)** | `InsertChain` re-executes the block from scratch | `InsertChain` finds cached result, **skips re-execution** |                                                                                    
| **Round change** | Cancel current round | Cancel current round + cancel in-flight speculative execution |
| **Speculative exec fails** | N/A | Graceful fallback to sync re-execution (Istanbul behavior) |                   
                                        
## Architecture                                                                                                     
                                        
consensus/kaiabft/
- backend.go     ← Engine/Handler impl, Seal, Commit, Verify, Gossip (~650 lines)
- machine.go     ← single runLoop, all PBFT handlers, speculative exec (~1250 lines)                              
  
**Key design decisions:**                                                                                           
                                        
- **Single-runloop state machine** — all PBFT state transitions in one dispatch loop in `machine.go`, not scattered 
across 7+ files like Istanbul. One breakpoint in `runLoop` sees every event.
- **No Istanbul import** — `kaiabft` imports `consensus/bft` (shared wire types) but never `consensus/istanbul`.    
Package topology: `bft` = leaf, `istanbul` and `kaiabft` both import `bft`, never each other.                       
- **`consensus.Sealer` interface** — kaiabft takes the sealer via interface, not the concrete
`*istanbul.IstanbulSealer`. The engine factory creates and injects it.                                              
- **Speculative execution is additive** — failure degrades to synchronous re-execution at `InsertChain`.
`ValidateState` always runs before writing, so a malicious proposer cannot poison the cache.                        
- **Eager commit** — validator sends prepare/commit without waiting for speculative execution to finish. No new
consensus semantics.                                                                                                
- **Context-based cancellation** — round changes cancel in-flight speculative goroutines via `context.Context`.
- **Switchable mode** — istBFT and kaiaBFT is switchable by `--consensus.engine` flag. The default one is `istBFT` for now, and this will be updated later.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist


consensus/kaiabft/machine.go
```diff
@@ -0,0 +1,1417 @@
...
+		m.unlockHash()
+		m.sendNextRoundChange("commit failure")
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Speculative execution (the kaiabft-specific addition)
+// ---------------------------------------------------------------------------
+
+func (m *machine) startSpeculativeExecution(proposal bft.Proposal) {
```
yoomee1313: Can we move speculative execution out of machine and into backend? I agree that machine decides when to run speculative execution, but it also performs the execution itself in `startSpeculativeExecution`.
* machine: trigger decision only
* backend: execution

consensus/kaiabft/backend.go
```diff
@@ -0,0 +1,651 @@
...
+	select {
+	case <-b.chainInitCh:
+	case <-time.After(chainInitTimeout):
+		return errNoChainReader
+	}
+	if b.chain == nil {
+		return errNoChainReader
+	}
+	if b.valsetModule == nil {
+		return errInvalidPeer
```
2dvorak:
```suggestion
		return errNoModule
```

consensus/kaiabft/backend.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+	council, err := b.valsetModule.GetCouncil(b.chain.CurrentHeader().Number.Uint64() + 1)
+	if err != nil {
+		return errInvalidPeer
```
2dvorak:
```suggestion
		return err
```
How about this?

consensus/kaiabft/backend.go
hyeonLewis: [417734b](https://github.com/kaiachain/kaia/pull/863/commits/417734b3e95a8d886fcb042c89aad7123a8a6e49)

consensus/kaiabft/backend.go
hyeonLewis: [417734b](https://github.com/kaiachain/kaia/pull/863/commits/417734b3e95a8d886fcb042c89aad7123a8a6e49)

consensus/kaiabft/machine.go
hyeonLewis: Seems clear to me. 94c42d9

yoomee1313 (approved):

PR #895, opened 2026-05-12
storage/database: surface errors from offline DB migration

2dvorak:
## Proposed changes

DB migration does not correctly return errors:
- `copyDB` returned `nil` error when the quit channel was closed by SIGINT/SIGTERM, so an interrupted migration was indistinguishable from a completed one.
  - Introduce `ErrDBMigrationInterrupted` and return it from copyDB on interruption
- `StartDBMigration`'s non-single source path collected per-DB copy errors into a channel but always returned `nil` after logging them.
  - Propagate the first error so the CLI exits non-zero when any per-DB copy fails (including interruption)
- Add regression tests covering interrupt-on-copy, the successful path, and non-single error propagation via a failing-batch stub.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/storage/database/db_migration.go
+++ b/storage/database/db_migration.go
@@ -19,13 +19,13 @@
 package database
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"os/signal"
 	"path"
 	"syscall"
 	"time"
-
-	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


storage/database/db_migration.go
```diff
@@ -146,17 +152,20 @@ func (dbm *databaseManager) StartDBMigration(dstdbm DBManager) error {
 			}()
 		}
 
+		var firstErr error
```
ian0371: Can you use [errors.Join](https://pkg.go.dev/errors#Join) to merge all errors and return?

2dvorak: Thanks. PTAL 621113cbfe2e933f564c946c8a7f352377fd1870

blukat29 (approved):

ian0371 (approved):

PR #897, opened 2026-05-13
valset: transition orchestration

ian0371:
## Proposed changes

- Orchestrator (`kaiax/valset/impl/transition.go`): per-block pipeline that reads ABv2 + gov + vrank for the parent state, applies the KIP-286 transitions via `TransitionContext`, and caches the `TransitionResult` by block number.
- ABv2 writers:
  - `WriteTransitionToABv2` — every post-fork block, writes the diff of changed nodes via `processSystemTransition`.
  - `InstallABv2` — at the HF-1 block's Finalize step, deploys the ABv2 proxy and runs `initialize()`.
- `ValsetModule` now satisfies `kaiax.BlockStateModule` (`InitializeState` / `FinalizeState`). Registration in `node/cn` will be included in the future PR.
- Supporting infrastructure:
  - `blockchain.SystemTxCall` — shared helper for kaiax-driven system txs.
  - `backends.NewStateBlockchainContractBackend` — `ContractCaller` over an arbitrary statedb, used during init/finalize when `StateAt` isn't available (e.g., state from `Initialize()` after reboot).
- Type renames: `system.NodeStatesResult` → `ABv2StateSnapshot`, field `Validators` → `Nodes`; doc/log strings switched from "validator" to "node" where the value covers all node states.
  - Replace `ABv2Context` with `ABv2TransitionParam`
- Unit tests for the orchestrator: cache hit, parent-read errors, pre-fork no-op, `fetchVRankCtx` (epoch / non-epoch / fail-open), `diffNodeStates`.
- vrank: VerifyHeader validates that epoch-start block should contain `CandTesting(N)`

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

ApplyTransition(N):

```mermaid
flowchart TD
  ABV2["ABv2 contract states<br/>state(N-1)"]
  MC["MultiCallContractNodeStatesResult"]
  Snapshot["system.ABv2Snapshot"]
  Param["system.ABv2TransitionParam"]
  NodeMap0["valset.NodeMap<br/>ABv2(N-1)"]
  GovCtx["impl.GovContext"]
  VRankCtx["impl.VRankContext"]
  BlockCtx["impl.BlockContext"]
  EpochCtx["impl.EpochTransitionContext"]
  TCtx["impl.TransitionContext"]
  TResult["impl.TransitionResult"]
  NodeMap1["valset.NodeMap<br/>NodeStates(N)"]
  Diff["valset.NodeMap<br/>changed states/timeouts"]
  Tx["types.Transaction / types.Message"]
  StateDB["state.StateDB<br/>ABv2(N)"]

  ABV2 -->|"multiCallNodeStatesPermissionless"| MC
  MC -->|"ReadABv2Snapshot"| Snapshot
  Snapshot -->|"Nodes.MarkSuspended"| NodeMap0
  Snapshot -->|"TransitionParam"| Param

  BlockCtx -->|"SetBlockCtx"| TCtx
  GovCtx -->|"SetGovCtx"| TCtx
  Param -->|"SetABv2TransitionParam"| TCtx
  EpochCtx -->|"SetSlotsCtx"| TCtx
  VRankCtx -->|"SetVRankCtx"| TCtx

  TCtx -->|"ApplyAllTransitions"| TResult
  NodeMap0 -->|"ApplyAllTransitions"| TResult
  TResult -->|"Nodes"| NodeMap1

  NodeMap0 -->|"diffNodeStates"| Diff
  NodeMap1 -->|"diffNodeStates"| Diff

  Diff -->|"EncodeProcessSystemTransition"| Tx
  Tx -->|"SystemTxCall"| StateDB
```

### DIFF
--- a/kaiax/valset/impl/transition.go
+++ b/kaiax/valset/impl/transition.go
@@ -65,6 +65,7 @@ func (v *ValsetModule) getTransitionResult(num uint64, parentStatedb *state.Stat
 	if cached, ok := v.transitionResultCache.Get(num); ok {
 		return cached.(*TransitionResult), nil
 	}
+
 	parentHeader := v.Chain.GetHeaderByNumber(num - 1)
 	if parentHeader == nil {
 		return nil, errParentHeaderNotFound(num)
@@ -86,31 +87,31 @@ func (v *ValsetModule) getTransitionResult(num uint64, parentStatedb *state.Stat
 //
 // Caller must ensure statedb == StateAt(header.Root).
 func (v *ValsetModule) applyTransition(header *types.Header, statedb *state.StateDB) (*TransitionResult, error) {
-	headNum := header.Number.Uint64()
-	nextNum := headNum + 1
-
-	// ABv2 read from state `headNum`
+	// ABv2 read from state(header.Root), i.e. ABv2(N).
 	abv2result, err := system.ReadABv2Snapshot(statedb, v.Chain, header)
 	if err != nil {
 		return nil, err
 	}
 	abv2result.Nodes.MarkSuspended(abv2result.SuspendedValidators)
 
-	// Gov read from `nextNum`
-	pset := v.GovModule.GetParamSet(nextNum)
+	ctx := v.newTransitionContext(header, abv2result)
+	return ctx.ApplyAllTransitions(abv2result.Nodes), nil
+}
 
-	// VRank reads `headNum` (fail-open via nil maps on error).
+func (v *ValsetModule) newTransitionContext(header *types.Header, abv2result *system.ABv2Snapshot) *TransitionContext {
+	headNum := header.Number.Uint64()
+	nextNum := headNum + 1
+
+	pset := v.GovModule.GetParamSet(nextNum)
 	cfs, pfs, pfReport := v.fetchVRankCtx(headNum)
 
-	// Build the transition context.
 	ctx := NewTransitionContext()
 	ctx.SetBlockCtx(header, v.isVrankEpoch(nextNum))
 	ctx.SetGovCtx(pset)
 	ctx.SetABv2TransitionParam(abv2result.TransitionParam())
 	ctx.SetSlotsCtx(slotLimitsFor(abv2result.EpochVACount))
 	ctx.SetVRankCtx(cfs, pfs, pfReport)
-
-	return ctx.ApplyAllTransitions(abv2result.Nodes), nil
+	return ctx
 }
 
 // fetchVRankCtx pulls the three vrank scores transitions need at block num.
--- a/kaiax/valset/impl/transition.go
+++ b/kaiax/valset/impl/transition.go
@@ -114,8 +114,7 @@ func (v *ValsetModule) applyTransition(header *types.Header, statedb *state.Stat
 }
 
 // fetchVRankCtx pulls the three vrank scores transitions need at block num.
-// Errors are swallowed here so the orchestrator stays robust to transient
-// vrank failures; the resulting nil/empty maps cause transitions to fail open.
+// Errors are swallowed here so the orchestrator stays robust to transient vrank failures.
 func (v *ValsetModule) fetchVRankCtx(num uint64) (cfs, pfs map[common.Address]uint64, pfReport []common.Address) {
 	if v.VRankModule == nil {
 		logger.Error("VRankModule is nil")
@@ -155,8 +154,8 @@ func (v *ValsetModule) isVrankEpoch(num uint64) bool {
 // WriteTransitionToABv2 computes the node-state transition for the current
 // block and writes any diffs to AddressBookV2 via SystemTx (ABv2.processSystemTransition).
 // No-op pre-fork.
-// WriteTransitionToABv2(n) is for generating the next block.
-// WriteTransitionToABv2(n) = Write NodeStates(N) - NodeStates(N-1) at Initialize(N).
+// WriteTransitionToABv2(N) is for generating the next block.
+// WriteTransitionToABv2(N) = Write NodeStates(N) - NodeStates(N-1) at Initialize(N).
 func (v *ValsetModule) WriteTransitionToABv2(
 	vmenv *vm.EVM,
 	header *types.Header,
--- a/kaiax/valset/impl/transition.go
+++ b/kaiax/valset/impl/transition.go
@@ -123,21 +123,21 @@ func (v *ValsetModule) fetchVRankCtx(num uint64) (cfs, pfs map[common.Address]ui
 	if nextNum := num + 1; v.isVrankEpoch(nextNum) {
 		c, err := v.VRankModule.GetCFS(num)
 		if err != nil {
-			logger.Warn("fetchVRankCtx: GetCFS failed", "num", num, "err", err)
+			logger.Error("fetchVRankCtx: GetCFS failed", "num", num, "err", err)
 		} else {
 			cfs = c
 		}
 	}
 	r, err := v.VRankModule.GetPfReport(num)
 	if err != nil {
-		logger.Warn("fetchVRankCtx: GetPfReport failed", "num", num, "err", err)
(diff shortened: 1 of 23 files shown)


kaiax/valset/impl/transition.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, err
+	}
+	return result.Nodes, nil
+}
+
+func (v *ValsetModule) getTransitionResult(num uint64, parentStatedb *state.StateDB) (*TransitionResult, error) {
+	if cached, ok := v.transitionResultCache.Get(num); ok {
+		return cached.(*TransitionResult), nil
+	}
+	parentHeader := v.Chain.GetHeaderByNumber(num - 1)
```
hyunsooda, line 90: I noticed the comment // Read ABv2(N-1) + apply transitions for block N was removed here. Was this omitted by accident? It provides good context, so I think it's worth keeping.

kaiax/valset/impl/transition.go
```diff
@@ -0,0 +1,285 @@
...
+	nextNum := headNum + 1
+
+	// ABv2 read from state `headNum`
+	abv2result, err := system.ReadNodeStates(statedb, v.Chain, header)
+	if err != nil {
+		return nil, err
+	}
+	abv2result.Nodes.MarkSuspended(abv2result.SuspendedValidators)
+
+	// Gov read from `nextNum`
```
hyunsooda: Looking at a new builder function may be introduced by wrapping around L100 - L111. How do you see?

kaiax/valset/impl/transition.go
```diff
@@ -0,0 +1,285 @@
...
+	DefaultMaxNodeCount            = 100
+	DefaultMaxValActivePausedCount = 50
+)
+
+// #region getter
+
+// getNodes returns the transitioned node map for producing block `num`.
+// Resolves parent state from chain (StateAt(parent.Root)). Use this for
+// post-commit reads (RPC, peer set queries).
+func (v *ValsetModule) getNodes(num uint64) (valset.NodeMap, error) {
```
hyunsooda, line 45: Missing the handle of `num == 0` case. Because there's no TODO comment for that, did you intentionally remove it? If so, it makes underflow at L49 if the genesis block is a permissionless fork.

kaiax/valset/impl/transition.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return nil
+}
+
+// diffNodeStates returns nodes whose state or timeout changed between parent and current.
+// Other Node fields are intentionally omitted: processSystemTransition writes
+// only lifecycle state and timeoutAt. StakingAmount comes from staking reads,
+// and Suspended is derived from ABv2's suspended validator list.
+func diffNodeStates(parent, current valset.NodeMap) valset.NodeMap {
+	diff := make(valset.NodeMap)
```
hyunsooda, line 241: The `Node` struct can be deep-copied thanks to the primitive list of fields, but whenever I read it, I lookup the struct and check the copy correctness. How do you see that to define a new Copy function like

```go
func (n *Node) Copy() *Node {
    if n == nil {
        return nil
    }
    copied := *n
    return &copied
}
```

kaiax/valset/impl/transition_context.go
```diff
@@ -89,12 +76,11 @@ type VRankContext struct {
 }
 
 // TransitionResult is the output of the orchestrator's transition pipeline.
-// EpochVACount is the count of ValActive after applyEpochTransition (before
-// violation); 0 for non-epoch blocks. The orchestrator writes it back to the
-// AddressBookV2 contract via processSystemTransition.
+// epochVACountForWrite is set only for epoch blocks, where it is written back
```
hyunsooda, line 79: Missing `Purge()` call at rewind module.

kaiax/vrank/errors.go
```diff
@@ -38,6 +38,7 @@ var (
 
 	ErrUnexpectedVRankBeforePermissionless = errors.New("unexpected vrank before permissionless fork")
 	ErrUnexpectedVRankAtEpochStart         = errors.New("unexpected vrank at epoch start block")
+	ErrMismatchedEpochStartVRank           = errors.New("vrank at epoch start does not match CandTesting")
```
hyunsooda: Based on the error usage, how aobut `ErrEpochVRankCandidateMismatch`?

kaiax/valset/impl/transition.go
hyunsooda, line 45: Based on the KaiaBFT refactoring, EIP-2935 is now executed before all BlockStateModule initializations in StateProcessor.InitializeState. Previously, we had an assertion that checked
IntermediateRoot(parentStatedb) == parentHeader.Root to verify the statedb was the committed parent state.

Under the new ordering, this assertion would always fail because EIP-2935 modifies the statedb (writing to history storage contract) before ValsetModule.InitializeState is called —
meaning the intermediate root no longer matches parentHeader.Root at that point.

Since EIP-2935 writes to the history storage contract and the valset transition reads from ABv2, the two are independent and the ordering does not affect correctness. So removing the
assertion is the right call.

Could you confirm that the assertion was removed specifically because the new ordering makes it always fail, rather than as a separate design decision?

kaiax/valset/impl/transition.go
ian0371: fixed c7d2d6563

kaiax/valset/impl/transition.go
ian0371: added `newTransitionContext` 63c2f89a4

kaiax/valset/impl/transition.go
ian0371: fixed 5bbbbac52. I initially thought `getOrComputeNodeStates` -> `getNodes` and missed it

kaiax/valset/impl/transition_context.go
ian0371: It's in Start() along with other cache purge logic

kaiax/vrank/errors.go
ian0371: I think it should contain "epoch start"; what about `ErrEpochStartVRankMismatch`?

kaiax/valset/impl/transition.go
ian0371: I removed `IntermediateRoot(parentStatedb)` check because it calls `statedb.Finalise()` internally which can have a side effect; it isn't a read-only operation.
I wasn't aware of the ordering though. Thanks for mentioning it.

kaiax/valset/impl/transition.go
ian0371: added b3ba401f9

kaiax/valset/impl/transition_context.go
hyunsooda: When it comes to rewind, I thought, semantically, each module should clear the internal cache explicitly. Do you mean that we don't need to do it because it's always deterministic?

kaiax/vrank/errors.go
hyunsooda: I thoguth that the maining of `epoch start` is implicitly subsumed by `VRankCandidate` because the candidate list is stamped at only vrank epoch start. Either is fine.

kaiax/valset/impl/transition_context.go
ian0371: `setHead()` always calls `module.Stop()` and `module.Start()`, so it's reset after rewind.

Although, some modules clear cache in `rewindTo` (staking, randao, etc.) and some in `Start` (supply). Let's unify the policy later.

kaiax/vrank/errors.go
ian0371: It may be misread as `VRankCandidate` message type, so I'll go with `ErrEpochStartVRankMismatch`. fixed f64e27b52

hyunsooda (approved):

hyeonLewis (approved):

blukat29 (approved):

PR #924, opened 2026-06-05
consensus: harden IBFT header and message verification

ian0371:
## Proposed changes

For permissionless phase, make IBFT verification accept only headers/messages the live consensus could actually have produced. Four independent fixes.

**1. Bind data into the signature-recovery cache key.**
`cacheSignatureAddress` keyed its LRU on the signature alone, but
`ecrecover(data, sig)` depends on `data` — so a header reusing a cached
signature over different data recovered the original signer, letting
`VerifyHeader` accept a forged header. Now keyed on `keccak256(data, sig)`.

**2. Verify committed seals against the committee with the consensus quorum.**
`verifySeals` counted seals against the council/qualified set (a superset of
the round's committee) and required only `2f+1`, while consensus commits
against the committee with `ceil(2N/3)`. `IstanbulSealer.Quorum` now returns
`ceil(2N/3)`; post-permissionless, `verifySeals` counts against the committee
and requires that quorum. Gated behind the Permissionless fork; the
pre-permissionless path is preserved exactly (`2f+1`).

**3. Verify a COMMIT's `CommittedSeal`.** `handleCommit` never checked the inner
`CommittedSeal` was the sender's signature before copying it into the sealed
block. Now recovers the signer and requires it equals the sender.

**4. Restrict round-change admission to the committee.** ROUND CHANGE was
admitted from any qualified validator while the thresholds are committee-sized.
Now gated on committee membership, mirroring `handleCommit`.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/blockchain/block_validator.go
+++ b/blockchain/block_validator.go
@@ -255,9 +255,34 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 		if err != nil {
 			return err
 		}
+		// author == proposer implies the author is the committee-selected proposer,
+		// so a separate qualified-membership check is redundant here.
 		if author != proposer {
 			return consensus.ErrUnauthorized
 		}
+
+		// Count committed seals only from the round's committee, matching the set the
+		// live consensus commits against (handleCommit rejects non-committee senders),
+		// rather than the broader qualified/council set.
+		committee, err := v.mValset.GetCommittee(blockNum, uint64(round))
+		if err != nil {
+			return err
+		}
+		committeeSet := valset.NewAddressSet(committee)
+		validSeal := 0
+		for _, addr := range committers {
+			if committeeSet.Remove(addr) {
+				validSeal++
+			} else {
+				return istanbul.ErrInvalidCommittedSeals
+			}
+		}
+		// Require the quorum over the committee. Post-permissionless the committee
+		// equals the qualified set, so pass its size for both arguments.
+		if validSeal < v.sealer.Quorum(blockNum, len(committee), len(committee)) {
+			return istanbul.ErrInvalidCommittedSeals
+		}
+		return nil
 	}
 
 	qualified, err := v.mValset.GetQualifiedValidators(blockNum)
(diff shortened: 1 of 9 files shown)


consensus/istanbul/core/roundchange.go
```diff
@@ -95,15 +95,21 @@ func (c *core) handleRoundChange(msg *bft.Message, src common.Address) error {
...
 	if err := c.checkMessage(bft.MsgRoundChange, rc.View); err != nil {
 		return err
 	}
 
+	// Restrict round-change admission to committee members, mirroring handleCommit.
+	// The thresholds (requiredMessageCount, f) are committee-sized, but messageSet
+	// admits any qualified validator; gating on the committee keeps admission
+	// consistent with the thresholds. The committee equals the qualified set (now and
+	// post-permissionless), so this is round-agnostic.
+	if !c.current.committee.Contains(src) {
```
hyunsooda, line 107: Fork-gating seems omitted

blockchain/block_validator.go
```diff
@@ -255,9 +255,34 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
...
 		}
+
+		// Count committed seals only from the round's committee, matching the set the
+		// live consensus commits against (handleCommit rejects non-committee senders),
+		// rather than the broader qualified/council set.
+		committee, err := v.mValset.GetCommittee(blockNum, uint64(round))
+		if err != nil {
+			return err
+		}
+		committeeSet := valset.NewAddressSet(committee)
```
hyunsooda, line 272: How about making a closure or an unexported function for L272 - 278, which are duplicated calculations. logic with the L305 - 312

consensus/istanbul/core/roundchange.go
ian0371: It's intentional; I don't want fork gating in consensus code. It never existed. It should be ok because non-committee members will not send RC messages.

blockchain/block_validator.go
ian0371: added aa8bc7be5

hyunsooda (approved):

yoomee1313 (approved):

PR #930, opened 2026-06-09
kaiax/valset: serve committee/qualified validators from the header

hyunsooda:
## Proposed changes

`GetCommittee` and `GetQualifiedValidators` recomputed the validator set from state (`getNodes → StateAt(parent.Root)`), which throws `missing trie node` on pruned blocks (RPC, historical re-validation, reward recompute).

Post-fork the canonical header already carries this set (written by `WriteValidators`, checked by `VerifyHeader`), so this serves it straight from the header and only falls back to state when the header isn't available yet (block being produced) or pre-fork — the same pattern `getBlockContext` already uses for `GetProposer`. Also removes the now-unused internal `getCommittee`.

Not changed: `VerifyHeader` still recomputes from state (malicious-proposer guard intact); consensus core stays state-based (in-flight block isn't canonical); `GetCouncil`/`GetDemotedValidators`/etc. still need state (header doesn't carry their data).

## Types of changes

- [x] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Related issues

N/A

### DIFF
--- a/kaiax/valset/impl/getter.go
+++ b/kaiax/valset/impl/getter.go
@@ -47,16 +47,30 @@ func (v *ValsetModule) GetDemotedValidators(num uint64) ([]common.Address, error
 	return demoted.List(), nil
 }
 
-func (v *ValsetModule) GetQualifiedValidators(num uint64) ([]common.Address, error) {
-	var (
-		qualified *valset.AddressSet
-		err       error
-	)
-	if v.Chain.Config().IsPermissionlessForkEnabled(new(big.Int).SetUint64(num)) {
-		qualified, err = v.getQualifiedValidators(num)
-	} else {
-		qualified, err = v.getQualifiedPermissioned(num)
+// qualifiedFromHeaderOrState returns the qualified validator set for block `num`.
+//
+// Post-fork, a canonical header already carries the verified validator set (== the
+// committee in permissionless), so we read it from the header to avoid a state trie
+// lookup that fails with "missing trie node" on pruned blocks. We compute from state
+// instead when pre-fork, or when the header isn't available yet (block being
+// produced) or unparsable.
+//
+// Safe: VerifyHeader independently rechecks the header against state-computed
+// qualified validators, so the header we trust here is already validated.
+func (v *ValsetModule) qualifiedFromHeaderOrState(num uint64) (*valset.AddressSet, error) {
+	if !v.Chain.Config().IsPermissionlessForkEnabled(new(big.Int).SetUint64(num)) {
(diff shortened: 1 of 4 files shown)


kaiax/valset/impl/getter.go
```diff
@@ -47,16 +47,34 @@ func (v *ValsetModule) GetDemotedValidators(num uint64) ([]common.Address, error
...
-		qualified, err = v.getQualifiedValidators(num)
-	} else {
-		qualified, err = v.getQualifiedPermissioned(num)
+// qualifiedFromHeaderOrState returns the qualified validator set for block `num`.
+//
+// Post-fork, a canonical header already carries the verified validator set (== the
+// committee in permissionless), so we read it from the header to avoid a state trie
+// lookup that fails with "missing trie node" on pruned blocks. We compute from state
+// instead when pre-fork, when the header isn't canonical yet (block being produced),
+// or at genesis (num == 0, not seal-produced — resolve like the other genesis getters).
```
ian0371: Doesn't num=0 also contain istanbul.Validator at extra? Does it need to be "seal-produced"?

kaiax/valset/impl/getter.go
```diff
@@ -47,16 +47,34 @@ func (v *ValsetModule) GetDemotedValidators(num uint64) ([]common.Address, error
...
+// qualified validators, so the header we trust here is already validated.
+func (v *ValsetModule) qualifiedFromHeaderOrState(num uint64) (*valset.AddressSet, error) {
+	if !v.Chain.Config().IsPermissionlessForkEnabled(new(big.Int).SetUint64(num)) {
+		return v.getQualifiedPermissioned(num)
+	}
+	if num > 0 {
+		if header := v.Chain.GetHeaderByNumber(num); header != nil {
+			vals, err := v.Chain.Sealer().Validators(header)
+			if err != nil {
+				return nil, err
```
ian0371: maybe fallback to state in this case?

kaiax/valset/impl/getter.go
hyunsooda: Excluding the case `num ==0` looks more consistent. Removed.

kaiax/valset/impl/getter.go
hyunsooda: Thanks for the spot. Falls back to calc. with state instead of returning the error.

ian0371 (approved):

hyeonLewis (approved):

PR #938, opened 2026-06-11
txpool: fix queue eviction order and missing-nonce fill at full pool

hyeonLewis:
## Proposed changes

Two txpool bugs that together caused the CN queue explosion under sustained load: the queue ratcheted to `NonExecSlotsAll`, pending fluctuated, and mining degraded until restart.

**1. Queue-cap eviction evicted the freshest accounts first (`txpool: evict stalest accounts first and keep queue frontier`)**

`addresssByHeartbeat` sorts oldest-beat-first, but the eviction loop pops victims from the end of the slice — evicting the account with the *newest* heartbeat first. Since beats refresh on promotion, this deleted the queues of healthy, actively-promoting accounts while preserving stuck ones (upstream geth uses `sort.Reverse` here; the comment "drop oldest ones" describes the intended behavior). Each overflow event manufactured fresh nonce gaps in healthy accounts, feeding back into more queue growth.

Fixes:
- Restore stalest-first victim ordering (`sort.Reverse`).
- Evict from the highest nonce down and always keep the account's lowest queued tx and its queue entry. Deleting the entry closed the account's missing-nonce admission path at full pool (`queue[from] == nil` refusal), making the account unhealable.
- Refresh the beat when an account's queue entry is recreated, so it doesn't inherit a stale beat and become the immediate next victim.
- Add `txpool/queued/capevict` and `txpool/queued/lifetimeevict` counters (the lifetime eviction path previously had no metric).

**2. Missing-nonce fill at full pool destroyed the tx it was meant to help (`txpool: fix missing-nonce fill at full pool`)**

At full pool, a tx filling a missing nonce evicted the account's max queued tx *before* running the underpriced check. `Underpriced` treated equal gas price as underpriced (`>=`), so under uniform pricing every fill attempt deleted one queued tx and then got refused — and when the evicted tx was the account's last queued one, the queue entry was deleted as well, permanently closing the heal path. Redelivery (resend/gossip) repeated this, grinding the account's delivered lookahead to nothing.

Fixes:
- Decide the refusal before evicting (refuse-without-collateral).
- `Underpriced`: `>=` → `>`. Equal price is no longer refused; this is safe because the function's single caller admits via a strictly intra-account swap (the sender's own max queued tx is evicted), so no third party is displaced. With this, a redelivered missing-nonce tx heals the account at full pool.
- Guard the subsequent `Discard` call: a multi-slot evicted tx made its count negative, which panics in `make` (latent crash, reachable before this PR).

**Validation**: stress test (11k tx/s, 10k unique senders, 4 CN / 8 EN, fixed gas price) — No queue increasing > 1h. Please note that the queue explosion happens with higher load, which applies to KaiaBFT, so the stress test is done based on the KaiaBFT branch.

image

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

The `Underpriced` semantic change (`>=` → `>`) is intentional and reviewed: the function has exactly one production caller (the full-pool missing-nonce fill), where the eviction is confined to the sender's own queue. Equal-price churn is self-limiting — each swap consumes one of the sender's own queued txs and closes one of their own nonce holes, and new holes cannot be created at full pool since non-missing-nonce txs are refused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/blockchain/tx_pool.go
+++ b/blockchain/tx_pool.go
@@ -119,6 +119,10 @@ var (
 	underpricedTxCounter = metrics.NewRegisteredCounter("txpool/underpriced", nil)
 	refusedTxCounter     = metrics.NewRegisteredCounter("txpool/refuse", nil)
 	slotsGauge           = metrics.NewRegisteredGauge("txpool/slots", nil)
+
+	// Queue eviction metrics, split by path.
+	queueCapEvictionCounter = metrics.NewRegisteredCounter("txpool/queued/capevict", nil)
+	lifetimeEvictionCounter = metrics.NewRegisteredCounter("txpool/queued/lifetimeevict", nil)
 )
 
 // TxStatus is the current status of a transaction as seen by the pool.
@@ -411,6 +415,7 @@ func (pool *TxPool) loop() {
 					if pool.queue[addr] != nil {
 						for _, tx := range pool.queue[addr].Flatten() {
 							pool.removeTx(tx.Hash(), true)
+							lifetimeEvictionCounter.Inc(1)
 						}
 					}
 					delete(pool.beats, addr)
@@ -1245,6 +1250,9 @@ func (pool *TxPool) enqueueTx(hash common.Hash, tx *types.Transaction) (bool, er
 	from, _ := types.Sender(pool.signer, tx) // already validated
 	if pool.queue[from] == nil {
 		pool.queue[from] = newTxList(false)
+		// Refresh the beat on (re)creation so the account doesn't inherit a
+		// stale one and become the next eviction victim.
+		pool.beats[from] = time.Now()
 	}
 	inserted, old := pool.queue[from].Add(tx, pool.config.PriceBump, pool.rules.IsMagma)
 	if !inserted {
@@ -1264,8 +1272,6 @@ func (pool *TxPool) enqueueTx(hash common.Hash, tx *types.Transaction) (bool, er
 		pool.all.Add(tx)
 		pool.priced.Put(tx)
 	}
-
-	pool.checkAndSetBeat(from)
 	return old != nil, nil
 }
 
@@ -1640,15 +1646,6 @@ func (pool *TxPool) GetBlobSidecarFromPool(txHash common.Hash) (*types.BlobTxSid
 	return sidecar, nil
 }
 
-// checkAndSetBeat sets the beat of the account if there is no beat of the account.
-func (pool *TxPool) checkAndSetBeat(addr common.Address) {
-	_, exist := pool.beats[addr]
-
-	if !exist {
-		pool.beats[addr] = time.Now()
-	}
-}
-
 // removeTx removes a single transaction from the queue, moving all subsequent
 // transactions back to the future queue.
 func (pool *TxPool) removeTx(hash common.Hash, outofbound bool) {
@@ -1849,37 +1846,29 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) {
 	queued := pool.queuedCount
 
 	if queued > pool.config.NonExecSlotsAll {
-		// Sort all accounts with queued transactions by heartbeat
+		// Sort by heartbeat with the stalest account last: the eviction loop below
+		// pops from the tail, so the least-recently-active accounts are dropped first.
 		addresses := make(addresssByHeartbeat, 0, len(pool.queue))
 		for addr := range pool.queue {
 			if !pool.locals.contains(addr) { // don't drop locals
 				addresses = append(addresses, addressByHeartbeat{addr, pool.beats[addr]})
 			}
 		}
-		sort.Sort(addresses)
+		sort.Sort(sort.Reverse(addresses))
 
-		// Drop transactions until the total is below the limit or only locals remain
+		// Drop highest-nonce txs from the stalest accounts, keeping each
+		// account's lowest queued tx so missing-nonce admission can heal it.
 		for drop := queued - pool.config.NonExecSlotsAll; drop > 0 && len(addresses) > 0; {
 			addr := addresses[len(addresses)-1]
 			list := pool.queue[addr.address]
 
 			addresses = addresses[:len(addresses)-1]
 
-			// Drop all transactions if they are less than the overflow
-			if size := uint64(list.Len()); size <= drop {
-				for _, tx := range list.Flatten() {
-					pool.removeTx(tx.Hash(), true)
-				}
-				drop -= size
-				queuedRateLimitCounter.Inc(int64(size))
-				continue
-			}
-			// Otherwise drop only last few transactions
 			txs := list.Flatten()
-			for i := len(txs) - 1; i >= 0 && drop > 0; i-- {
+			for i := len(txs) - 1; i >= 1 && drop > 0; i-- {
 				pool.removeTx(txs[i].Hash(), true)
 				drop--
-				queuedRateLimitCounter.Inc(1)
+				queueCapEvictionCounter.Inc(1)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


blockchain/tx_pool.go
```diff
@@ -1170,30 +1175,35 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
...
 			logger.Trace("Rejecting a new Tx, because TxPool is full and a new TX does not have missing nonce", "hash", tx.Hash())
 			refusedTxCounter.Inc(1)
 			return false, fmt.Errorf("txpool is full and the new tx does not have missing nonce: %d", uint64(pool.all.Count()))
 		}
 
-		// (4) discard underpriced transactions
-		// If the new transaction is underpriced, don't accept it
+		// (3) discard underpriced transactions. Decided before the eviction in
+		// (4); the previous evict-then-check order destroyed maxTx on every
+		// refusal and deleted single-tx queue entries with it.
```
hyunsooda: Nit: This narrates the previous (removed) behavior — after merge there is no `evict-then-check order` left in the code to compare against, so it reads as dangling history. 

```suggestion
		// (3) discard underpriced transactions, before the eviction in (4): a refused
		// tx must leave the sender's queue untouched, so the eviction only runs once
		// the tx is certain to be admitted.
```

blockchain/tx_pool.go
```diff
@@ -1849,37 +1851,29 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) {
 	queued := pool.queuedCount
 
 	if queued > pool.config.NonExecSlotsAll {
-		// Sort all accounts with queued transactions by heartbeat
+		// Sort all accounts with queued transactions by heartbeat, stalest last.
+		// Without sort.Reverse the loop below evicted the freshest accounts first.
```
hyunsooda: Same pattern — `Without sort.Reverse ... evicted the freshest accounts first` describes code that no longer exists. Stating why the current ordering is required is clearer and won't rot:

```suggestion
		// Sort by heartbeat with the stalest account last: the eviction loop below
		// pops from the tail, so the least-recently-active accounts are dropped first.
```

hyunsooda (dismissed):
LGTM

blockchain/tx_pool.go
hyeonLewis: Thanks, applied.

blockchain/tx_pool.go
hyeonLewis: Thanks, applied.

hyunsooda (approved):

blockchain/tx_pool.go
```diff
@@ -1170,30 +1175,35 @@ func (pool *TxPool) add(tx *types.Transaction, local bool) (bool, error) {
...
-		for _, tx := range drop {
-			logger.Trace("Discarding freshly underpriced transaction", "hash", tx.Hash(), "price", tx.GasPrice())
-			underpricedTxCounter.Inc(1)
-			pool.removeTx(tx.Hash(), false)
+
+		// (4) remove an old Tx with the largest nonce from queue to make a room for a new Tx with missing nonce
+		pool.removeTx(maxTx.Hash(), true)
+		logger.Trace("Removing an old Tx with the max nonce to insert a new Tx with missing nonce, because TxPool is full", "account", from, "new nonce(previously missing)", tx.Nonce(), "removed max nonce", maxTx.Nonce())
+
+		// New transaction is better than our worse ones, make room for it.
```
ian0371, line 1198: It seems we're redefining the definition of "a better tx": from gas price to nonce-gap filling. Previously, we could discard a tx only when its gas price is strictly lower; now we will discard a tx even if the gas price is the same.

Do you think it can introduce other nonce-gapped txs? Or can we ignore this because in most cases `numSlots(newTx) = numSlots(maxTx)` where `Discard` won't happen?

hyeonLewis: This is only applied within an account, and the changed replacement logic aligns with the priced heap `Less` logic. ~~ I don't think this intra-account change itself will cause another nonce gap issue. ~~

ian0371: `Discard` doesn't happen within an account; it happens across the global tx pool.

A scenario where `largeMissingNonce` replaces third party tx of equal gas price:

```go
// Tests that an equal-priced missing-nonce tx is not allowed to evict
// third-party txs when the same-account max queued tx does not free enough
// slots for the incoming tx.
func TestTransactionFullPoolEqualPriceLargerMissingNonceDoesNotDiscardThirdParty(t *testing.T) {
	t.Parallel()

	statedb, _ := state.New(common.Hash{}, state.NewDatabase(database.NewMemoryDBManager()), nil, nil)
	blockchain := &testBlockChain{statedb: statedb, gasLimit: 10000000, chainHeadFeed: new(event.Feed), txMap: make(map[common.Hash]*types.Transaction)}

	config := testTxPoolConfig
	config.ExecSlotsAccount = 8
	config.ExecSlotsAll = 2
	config.NonExecSlotsAccount = 8
	config.NonExecSlotsAll = 1

	pool := NewTxPool(config, params.TestChainConfig.Copy(), blockchain, &dummyGovModule{chainConfig: params.TestChainConfig.Copy()})
	defer pool.Stop()

	var (
		keyA, _  = crypto.GenerateKey()
		keyB, _  = crypto.GenerateKey()
		keyC, _  = crypto.GenerateKey()
		addrA    = crypto.PubkeyToAddress(keyA.PublicKey)
		gasPrice = big.NewInt(10)

		maxTx             = pricedTransaction(1, 100000, gasPrice, keyA)
		largeMissingNonce = pricedDataTransaction(0, 5000000, gasPrice, keyA, txSlotSize)
		thirdParty1       = pricedTransaction(0, 100000, gasPrice, keyB)
		thirdParty2       = pricedTransaction(0, 100000, gasPrice, keyC)
	)
	testAddBalance(pool, addrA, big.NewInt(10000000000))
	testAddBalance(pool, crypto.PubkeyToAddress(keyB.PublicKey), big.NewInt(10000000000))
	testAddBalance(pool, crypto.PubkeyToAddress(keyC.PublicKey), big.NewInt(10000000000))

	// initial txpool: [maxTx, thirdParty1, thirdParty2]
	require.Equal(t, 1, numSlots(maxTx))
	require.Equal(t, 1, numSlots(thirdParty1))
	require.Equal(t, 1, numSlots(thirdParty2))
	require.NoError(t, pool.AddRemote(maxTx))
	require.NoError(t, pool.AddRemote(thirdParty1))
	require.NoError(t, pool.AddRemote(thirdParty2))
	require.Equal(t, 3, pool.all.Slots())

	// The missing nonce tx has the same gas price but consumes more slots than
	// maxTx. Admitting it requires an extra global Discard after maxTx removal,
	// which can evict one of the third-party txs.
	require.Greater(t, numSlots(largeMissingNonce), numSlots(maxTx))

	// add largeMissingNonce to txpool
	err := pool.AddRemote(largeMissingNonce)
	require.NoError(t, err)

	require.NotNil(t, pool.all.Get(thirdParty1.Hash()), "thirdParty1 is removed")
	require.NotNil(t, pool.all.Get(thirdParty2.Hash()), "thirdParty2 is removed")
}
```

hyeonLewis: Ah, yes, `Discard` already existed (I thought you claimed that it was intra-account logic), but this PR could broaden the cases. But as you mentioned, the impact is bounded, so it seems ok.

hyeonLewis: In either case (what this PR fixes vs what this PR broadens), a nonce gap can occur, but the intra-account fix has priority, I think.

ian0371: Got it thanks

ian0371 (approved):

PR #939, opened 2026-06-11
p2p: reserve cross-type connection slots for the CN<->EN link

hyunsooda:
## Proposed changes

In a PN-less topology, an own-type mesh (CN↔CN or EN↔EN) was bounded only by `MaxPhysicalConnections`, so a busy own-type mesh could consume every slot and leave no room for the cross-type CN↔EN link.

This caps each own-type mesh at `MaxPhysicalConnections − (cross-type reservation)`, turning the cross-type `peerTargets` (CN→EN, EN→CN) into **guaranteed floors** instead of starvable ceilings — a CN always keeps room for ENs, and an EN always keeps room for CNs.

Also in this PR:
- **EN/PN no longer enforce the CN allowlist** — `SetCNPeers` is a no-op for them, so they admit and serve any CN (the post-fork CN↔EN sync path). Only CNs filter by `CNPeers`.
- Reservations are **operator-configurable** via `--reserved-en-conns` / `--reserved-cn-conns` (defaults `3` / `2`); each node's reservation governs only its own slot allocation, so no network-wide agreement is needed.
- Startup rejects a `MaxPhysicalConnections` below the reservation (own-mesh cap would be ≤ 0); runtime warns when `|CNPeers|` exceeds full-mesh + reservation capacity.

## Types of changes

- [ ] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

- [x] 📝 I have signed the CLA

## Related issues

- KIP-311 spec update: https://github.com/kaiachain/kips/pull/122


cmd/utils/flags.go
```diff
@@ -1140,6 +1140,20 @@ var (
...
+		Aliases:  []string{"p2p.reserved-en-conns"},
+		Category: "NETWORK",
+	}
+	ReservedCNConnsFlag = &cli.IntFlag{
+		Name:     "reserved-cn-conns",
+		Usage:    "Number of CN connection slots an EN/PN reserves so its EN mesh cannot starve the CN<->EN link (0 = use default)",
+		Value:    0,
+		Aliases:  []string{"p2p.reserved-cn-conns"},
+		Category: "NETWORK",
+	}
```
ian0371: Do you think we need two flags? Isn't a `reservedCrossTypeSlots` sufficient?

```suggestion
	ReservedCrossTypeSlotsFlag = &cli.IntFlag{
		Name:     "reserved-cross-type-slots",
		Usage:    "Number of connection slots reserved for the opposite CN/EN node type so same-type peers cannot starve CN<->EN
		connectivity (0 = use node-type default)",
		Value:    0,
		Aliases:  []string{"p2p.reserved-cross-type-slots"},
		Category: "NETWORK",
	}
```

If you think two flags are better, then consider separating them into `KCNFlags` & `KENFlags` instead of `CommonNodeFlags`.

networks/p2p/server_util.go
```diff
@@ -32,15 +32,77 @@ import (
...
+	if v > 0 {
+		return v
+	}
+	return defaultReservedCNForEN
+}
+
+// MinPhysicalConnections returns the smallest MaxPhysicalConnections a node of
+// cfg's connection type may use while still honoring its cross-type reservation.
+// Unset reservations fall back to the package defaults.
+func MinPhysicalConnections(cfg Config) int {
```
ian0371: Can we rename `MinPhysicalConnections`? Since it is only used to validate `MaxPhysicalConnections`, the current name sounds a bit like a separate p2p connection policy parallel to `MaxPhysicalConnections`.

Maybe a name like `MaxPhysicalConnectionsLowerBound`, `MinMaxPhysicalConnections`, or `RequiredMaxPhysicalConnections` would make the intent clearer: this is the minimum valid value for `MaxPhysicalConnections` after accounting for the reserved cross-type slots.

networks/p2p/server_util.go
```diff
@@ -32,15 +32,77 @@ import (
...
+)
+
+// peerTargetFor returns how many peers of peerType a selfType node may keep, given
+// maxconnections (maxPhys) and the node's cross-type reservations. Pass
+// EffectiveConnType values. The reservation turns into accept caps:
+//   - other type (CN->EN, EN->CN): the reserved number
+//   - same type  (CN->CN, EN->EN): maxconnections minus the reservation
+//
+// ok=false means "no per-type cap" — only maxconnections / inbound limits apply.
+func peerTargetFor(selfType, peerType common.ConnType, maxPhys, reservedENForCN, reservedCNForEN int) (target int, ok bool) {
```
ian0371: If reservedENForCN and reservedCNForEN are merged into ReservedCrossTypeSlotsFlag as mentioned above, we can consider:

```suggestion
func peerTargetFor(isSameType bool, maxPhys, reservedCrossTypeSlot int) (target int, ok bool) {
```

cmd/utils/flags.go
hyunsooda: Merged. https://github.com/kaiachain/kaia/pull/939/changes/73286827cfa37b4a21d53c475c77e166c2208fce

networks/p2p/server_util.go
hyunsooda: Renamed. https://github.com/kaiachain/kaia/pull/939/changes/73286827cfa37b4a21d53c475c77e166c2208fce

networks/p2p/server_util.go
hyunsooda: Restructured. https://github.com/kaiachain/kaia/pull/939/changes/73286827cfa37b4a21d53c475c77e166c2208fce

ian0371 (approved):

networks/p2p/server_util.go
```diff
@@ -32,15 +32,57 @@ import (
...
+// ok=false means no per-type cap (e.g. a bootnode, or a non-CN/EN node).
+func peerTargetFor(selfType, peerType common.ConnType, maxPhys, reservedCrossTypeSlots int) (target int, ok bool) {
+	isCNorEN := func(t common.ConnType) bool {
+		return t == common.CONSENSUSNODE || t == common.ENDPOINTNODE
+	}
+	if !isCNorEN(selfType) || !isCNorEN(peerType) {
+		return 0, false // unrelated pair (e.g. bootnode): no per-type cap
+	}
+	if selfType == peerType {
+		return max(0, maxPhys-reservedCrossTypeSlots), true // own mesh
```
yoomee1313, line 72: I noticed that max physical connection is a double of peer count when it's multichannel. Is there difference between multichannel and single channel?
```
klaytn_p2p_PeerCountGauge 3
klaytn_p2p_ConnectionCountGauge 6
```

yoomee1313 (approved):

networks/p2p/server_util.go
hyunsooda: The metrics themselves are expected. That said, I revisited the multichannel mechanism and it doesn't quite align with the KIP-311 spec: inbound admission is guarded by the inbound connection (socket) count, not the peer count. With multichannel enabled, each peer opens 2 sockets, so a node could end up admitting only ~half the intended peers — e.g. with maxconnections=100 and no outbound dialing (as on a CN), it would accept only ~50 inbound peers instead of 100.

KIP-311 defines these budgets in terms of peers, so I think the inbound check should count peers regardless of channel mode. This is pre-existing and orthogonal to this PR, so I'll address
it separately. 

cc. @ian0371

yoomee1313: Got it, thanks for clarifying. That makes sense to handle separately.

PR #947, opened 2026-06-18
kaiax/gov: apply DeprecatedAt policy to contract governance ingestion

hyunsooda:
## Proposed changes

Contract governance applied `GovParam` params without `gov.DeprecatedAt`, so it could
override immutable params like `reward.deferredtxfee` — desyncing reward finalization
(param set) from tx execution (genesis), corrupting fee accounting. Now contract-gov
drops deprecated params via `gov.DeprecatedAt` — always-deprecated params at any block,
permissionless-deprecated ones after the fork.

## Types of changes

- [x] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

Deprecated/to-be-deprecated params must never be set via `GovParam` (operational
invariant); contract-gov drops them defensively if present.

### DIFF
--- a/kaiax/gov/contractgov/impl/getter.go
+++ b/kaiax/gov/contractgov/impl/getter.go
@@ -81,6 +81,17 @@ func (c *contractGovModule) contractGetAllParamsAtFromAddr(blockNum uint64, addr
 	}
 
 	ret := ParseContractCall(names, values)
+
+	rules := config.Rules(new(big.Int).SetUint64(blockNum))
+	if rules.IsPermissionless {
+		for name := range ret {
+			if gov.DeprecatedAt(name, rules) {
+				logger.Warn("Ignoring deprecated parameter from contract governance", "name", name, "blockNum", blockNum)
+				delete(ret, name)
+			}
+		}
+	}
+
 	return ret, nil
 }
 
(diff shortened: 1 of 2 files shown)


kaiax/gov/contractgov/impl/getter.go
```diff
@@ -81,6 +81,17 @@ func (c *contractGovModule) contractGetAllParamsAtFromAddr(blockNum uint64, addr
 	}
 
 	ret := ParseContractCall(names, values)
+
+	rules := config.Rules(new(big.Int).SetUint64(blockNum))
+	if rules.IsPermissionless {
```
ian0371: Since we never voted forbidden params via GovParam, we can remove permissionless check.

hyunsooda: The gating was to avoid a gap between updated and outdated nodes. Without the gate, an updated node would start filtering immediately while an outdated one wouldn't — so if such a param ever came from GovParam, the two would branch.

Do you want to remove the gating and instead guarantee at the operational level that such params never come from GovParam?

ian0371: Yep, we should never vote deprecated or to-be-deprecated params.

hyunsooda: Removed. https://github.com/kaiachain/kaia/pull/947/changes/6083ff775c9aad967cc0b80d61a37e9a3278facc

ian0371 (approved):

2dvorak (approved):

PR #948, opened 2026-06-18
kaiax/gov: fix nil-header panic when rebuilding gov/vote cache

hyunsooda:
## Proposed changes

`readGovDataFromDB`/`readVoteDataFromDB` dereferenced index-listed headers without a nil
check, panicking on startup (every restart) when the index lists a block whose header is
absent. Now a block above the current head is skipped (no longer canonical); a missing
header at or below head is fatal (chain corruption). Shared loop extracted into
`forEachIndexedHeader`.

## Types of changes

- [x] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

Healthy DB is unaffected (no stale blocks → nothing skipped); block processing is unchanged.

### DIFF
--- a/kaiax/gov/headergov/impl/init.go
+++ b/kaiax/gov/headergov/impl/init.go
@@ -7,6 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/kaiachain/kaia/blockchain/types"
 	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/consensus"
 	"github.com/kaiachain/kaia/kaiax/gov"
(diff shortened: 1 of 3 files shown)


kaiax/gov/headergov/impl/init.go
```diff
@@ -280,43 +281,55 @@ func (h *headerGovModule) accumulateVotesInEpoch(epochIdx uint64) {
...
+	head := chain.CurrentBlock().NumberU64()
+	for _, blockNum := range blockNums {
 		header := chain.GetHeaderByNumber(blockNum)
-		parsedVote, err := headergov.VoteBytes(header.Vote).ToVoteData()
-		if err != nil {
-			panic(err)
+		if header == nil {
+			if blockNum > head {
+				logger.Warn("Skipping stale block above head", "kind", kind, "num", blockNum, "head", head)
+				continue
```
ian0371, line 294: Can we return in this case?

2dvorak (approved):

kaiax/gov/headergov/impl/init.go
hyunsooda: return here requires `blockNums` to be sorted ascending, delegating that guarantee to the caller (`continue` doesn't assume any order). If you're fine taking that implicit assumption, I'll
revise it to return.

ian0371: I see, let's leave it thanks.

ian0371 (approved):

PR #966, opened 2026-06-25
kaiax/valset: pin base proposer list to qualified set at updateNum+1

hyeonLewis:
## Proposed changes

# valset: pin base proposer list to qualified set at updateNum+1

## Summary

`valset.calcBaseProposers` caches the per-interval base proposer list under `updateNum` (first-writer-wins, purged on every `Start()`), but builds the list from `c.qualified` — the qualified validator set at the **requesting** block `c.num`, not at the proposer-update block. Because the qualified set changes within an interval (validator votes, staking-driven demotion/requalification), the cached list ends up depending on *which block first misses the cache* rather than on the canonical chain alone.

Two honest nodes — or one node before vs. after a mid-interval restart — can therefore cache different base lists for the same `updateNum` and compute different proposers for the same `(num, round)`. This is a determinism break in consensus-critical proposer selection.

Reachable only on the **WeightedRandom + pre-Randao** path (post-Randao live consensus uses the Randao committee path and never touches this cache). Practical exposure: a node that resumes sync mid-interval over a pre-Randao range where the validator set changed.

## Fix

Make the base list a pure function of `updateNum` by pinning the two `c`-dependent inputs to the interval's first block, `updateNum+1`:

- `qualified` ← `getQualifiedPermissioned(updateNum+1)` (was `c.qualified`)
- `IsKore` ← `Rules(updateNum+1).IsKore` (was `c.rules.IsKore`; otherwise the uniform-vs-weighted choice stays cache-order dependent in the Kore-boundary interval)

`GetStakingInfo(updateNum)` (weights), the seed (`hash@updateNum`) and `useGini` (immutable) are already deterministic and left untouched.

### Why `updateNum+1`, not `updateNum`

`updateNum+1` reproduces both the legacy istanbul behavior and the current code's correct path:

- Legacy `snapshot.apply` computed the list for interval `U` once, at header `N=U`, from `RefreshValSet(N+1)` = the validator set at `U+1`. That sealed every historical pre-Randao block.
- In sequential processing the first block to miss `cache[U]` is always `U+1` (block `U` belongs to the previous interval), so the current code already populates the cache with `qualified@(U+1)`.

`qualified@U` would differ from this whenever a validator vote lands exactly on a proposer-update block (`getCouncilDB` reflects a vote at `U` in `council@(U+1)`, not `council@U`) — i.e. that variant *would* change canonical results.

## Not a hardfork / no new state requirement

- **No hardfork:** the change is identical to the current code's sequential behavior. A fresh sequential re-sync (how nodes bootstrap) already uses `qualified@(U+1)`; mainnet re-syncs succeed today, so all history is already consistent with it. The fix only stops the poisoned-cache path from producing a non-canonical proposer.
- **No pruned-state dependency:** `getQualifiedPermissioned` is DB-backed (council from `ChainKv` / istanbul snapshot; demotion via `GetStakingInfo`, which for pre-Kaia reads cache → DB and persists, hitting account state only as a last-resort fallback). The weighted path already calls `GetStakingInfo(updateNum)` for an old block in production, proving old-interval staking is DB-resident.

## Testing

- New `TestCalcBaseProposers_PinnedToUpdateNumPlus1`: poisons the cache from a late block whose `c.qualified` is the demoted set and asserts the list is still derived from `qualified@(U+1)` and independent of cache-population order.
- Updated `TestGetProposer` / `TestGetCommittee` with the council-DB/gov/config setup the recompute now requires (expected proposers unchanged).
- `./kaiax/valset/...` green.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/kaiax/valset/impl/getter_proposers.go
+++ b/kaiax/valset/impl/getter_proposers.go
@@ -18,7 +18,7 @@ func (v *ValsetModule) getProposerList(c *blockContext) ([]common.Address, uint6
 	)
 
 	// calculate base pList at updateNum
-	pList, err := v.calcBaseProposers(c, updateNum, useGini)
+	pList, err := v.calcBaseProposers(updateNum, useGini)
 	if err != nil {
 		return nil, 0, err
 	}
@@ -30,7 +30,7 @@ func (v *ValsetModule) getProposerList(c *blockContext) ([]common.Address, uint6
 	return pList, updateNum, nil
 }
 
-func (v *ValsetModule) calcBaseProposers(c *blockContext, updateNum uint64, useGini bool) ([]common.Address, error) {
+func (v *ValsetModule) calcBaseProposers(updateNum uint64, useGini bool) ([]common.Address, error) {
 	// Lookup proposers at updateNum from cache
 	if list, ok := v.proposerListCache.Get(updateNum); ok {
 		return list.([]common.Address), nil
@@ -61,7 +61,7 @@ func (v *ValsetModule) calcBaseProposers(c *blockContext, updateNum uint64, useG
 		}
 		list = generateProposerListWeighted(qualified, si, useGini, updateHeader.Hash())
 	}
-	logger.Debug("GetProposerList", "number", c.num, "updateNum", updateNum, "list", valset.NewAddressSet(list).String())
+	logger.Debug("GetProposerList", "updateNum", updateNum, "list", valset.NewAddressSet(list).String())
 
 	// Store to cache
 	v.proposerListCache.Add(updateNum, list)
--- a/kaiax/valset/impl/getter_proposers.go
+++ b/kaiax/valset/impl/getter_proposers.go
@@ -2,6 +2,7 @@ package impl
 
 import (
 	"math"
+	"math/big"
 	"slices"
 
 	"github.com/kaiachain/kaia/common"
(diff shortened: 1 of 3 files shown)


kaiax/valset/impl/getter_proposers.go
```diff
@@ -40,17 +41,27 @@ func (v *ValsetModule) calcBaseProposers(c *blockContext, updateNum uint64, useG
...
 	} else {
 		si, err := v.StakingModule.GetStakingInfo(updateNum)
 		if err != nil {
 			return nil, err
 		}
-		list = generateProposerListWeighted(c.qualified, si, useGini, updateHeader.Hash())
+		list = generateProposerListWeighted(qualified, si, useGini, updateHeader.Hash())
 	}
-	logger.Debug("GetProposerList", "number", c.num, "list", valset.NewAddressSet(list).String())
+	logger.Debug("GetProposerList", "number", c.num, "updateNum", updateNum, "list", valset.NewAddressSet(list).String())
```
ian0371: Can we remove `c` from parameters? It seems this is the only usage

hyeonLewis: Applied. [f4ccb40](https://github.com/kaiachain/kaia/pull/966/commits/f4ccb408d8c403e076d99772c7c3127e0f5b9349)

ian0371 (approved):

hyunsooda (approved):

PR #978, opened 2026-07-13
blockchain/types: guard tx JSON decoders against empty signatures

2dvorak:
## Proposed changes

Return an error instead of panicking when the "signatures" field is missing, empty, or null while decoding a JSON transaction.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/api/api_kaia_transaction_test.go
+++ b/api/api_kaia_transaction_test.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"math/big"
 	"reflect"
 	"testing"
@@ -63,6 +64,34 @@ var (
 	senderPrvKey, _   = crypto.HexToECDSA("95a21e86efa290d6665a9dbce06ae56319335540d13540fb1b01e28a5b2c8460")
 	feePayerPrvKey, _ = crypto.HexToECDSA("aebb680a5e596c1d1a01bac78a3985b62c685c5e995d780c176138cb2679ba3e")
 )
+
+func TestSendTxArgsUnmarshalSignatures(t *testing.T) {
+	testcases := []struct {
+		name    string
+		input   string
+		wantErr bool
+	}{
+		{"missing", `{}`, false},
+		{"null", `{"signatures":null}`, false},
+		{"empty", `{"signatures":[]}`, false},
+		{"null_element", `{"signatures":[null]}`, true},
+		{"empty_element", `{"signatures":[{}]}`, true},
+		{"partial_element", `{"signatures":[{"V":"0x1"}]}`, true},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			var args SendTxArgs
+			err := json.Unmarshal([]byte(tc.input), &args)
+			if tc.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Empty(t, args.TxSignatures)
+			}
+		})
+	}
+}
 
 // TestTxTypeSupport tests tx type support of APIs in KaiaTransactionAPI.
 func TestTxTypeSupport(t *testing.T) {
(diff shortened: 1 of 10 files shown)


ian0371 (dismissed):

ian0371 (commented):
Do Kaia tx-types not require this guard?

2dvorak:
@ian0371 Thanks, PTAL eb812b1

ian0371 (commented):
| Category     | Item                                                                             | Location                                                   | Action                                                                                                                                                          |
| ------------ | -------------------------------------------------------------------------------- | ---------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Required** | Empty-list rejection in the shared JSON type                                     | `TxSignaturesJSON.UnmarshalJSON`, tx_signatures.go:172-174 | Remove the 3-line `len(sigs) == 0` branch                                                                                                                       |
| **Required** | Regression test for `SendTxArgs`                                                 | `api/`, new test                                           | `{}`, `{"signatures":null}`, `{"signatures":[]}` decode without error; `[null]`, `[{}]`, `[{"V":"0x1"}]` are rejected                                           |
| Optional     | Per-decoder guards `len == 0 \|\| [0] == nil` in Legacy and the 4 Ethereum types | blockchain/types/tx_internal_data_ethereum_*.go            | Remove `len(js.TxSignatures) == 0 \|\| js.TxSignatures[0] == nil` checks. Unreachable through the serializer once the two checks above are in place             |
| Optional     | Error name                                                                       | `errEmptyTxSignatures` in tx_signatures.go                 | Add a separate error such as `errInvalidTxSignatureJSON` for the element check in `TxSignaturesJSON.UnmarshalJSON`; the serializer keeps `errEmptyTxSignatures` |
| Optional     | Test coverage for malformed elements                                             | `tx_internal_data_json_signatures_test.go`                 | Add an `[{}]` case                                                                                                                                              |


For the first item, `len(sigs) == 0` is already checked by `serializer.UnmarshalJSON`, so `TxSignaturesJSON.UnmarshalJSON` is not required. It instead introduces a side effect where it has to be empty (`kaia_signTransactionAsFeePayer`). 
```
Transaction.UnmarshalJSON                                  transaction.go:278
└ serializer.UnmarshalJSON                                 tx_internal_data_serializer.go:110
    ├ json.Unmarshal(b, &txInternalDataJSON)               (pre-decode)
    │   └ TxSignaturesJSON.UnmarshalJSON                   tx_signatures.go:167  (nil check + len==0)
    ├ len(dec.TxSignatures) == 0 → errEmptyTxSignatures    serializer.go:115
    └ TxInternalDataValueTransfer.UnmarshalJSON            tx_internal_data_value_transfer.go:254  [per-tx-type decoder]
    ├ json.Unmarshal(b, js)                                tx_internal_data_value_transfer.go:256
    │   └ TxSignaturesJSON.UnmarshalJSON                   tx_signatures.go:167  (nil check + len==0)
    └ t.TxSignatures = js.TxSignatures.ToTxSignatures()    tx_internal_data_value_transfer.go:266  ← panic if nil
```

edwin082 (commented):
I agree with the two required changes above. `SendTxArgs` uses the shared `TxSignaturesJSON` type, and `SignTransactionAsFeePayer` explicitly allows absent sender signatures so the fee payer can sign first. The existing transaction pool tests also use fee-payer-first signing and verify acceptance of the completed transaction.

Please preserve omitted, `null`, and empty `signatures` in `SendTxArgs`, while keeping the required-signature checks in the transaction serializer. The proposed API regression tests should cover this distinction.

2dvorak:
@ian0371 @edwin082 Thank you for the comments, PTAL 374ab10f9

edwin082 (approved):
Approved.

---
Non-blocking note: the new fee-payer presence check changes how Go clients can decode partially signed transactions.

Specifically, when a Go client calls `kaia_signTransaction` for a fee-delegated transaction and decodes the response into `api.SignTransactionResult`, its `Tx *types.Transaction` field invokes `Transaction.UnmarshalJSON` → `TxInternalDataSerializer.UnmarshalJSON`. The response has a valid sender signature but empty `feePayerSignatures`, so the new check rejects it. This decoding previously succeeded.

The signing API still returns the response normally. Clients using `json.RawMessage`, plain JSON objects, or their own structs without `types.Transaction` are unaffected. Fee-payer signature validation remains enforced separately when transactions enter the pool or are processed.

I couldn’t find an in-repository RPC caller using that response-decoding path, so I’m noting this as a compatibility change rather than a blocker.

ian0371 (approved):

PR #985, opened 2026-07-20
consensus/istanbul, blockchain: bind committed seal to consensus round

hyunsooda:
## Proposed changes

The final consensus round is stored in a header vanity byte excluded from the block hash and seal preimages, so it is unauthenticated and a peer can alter it on an otherwise-valid block. From the permissionless fork, bind the round into the committed seal so a mutated round byte fails committer recovery and the block is rejected.

- `PrepareCommittedSealWithRound` / `MakeCommittedSealFromHashWithRound` / `CommittersWithRound`
- `finalizeMessage`/`handleCommit` sign/verify the round-bound seal, `verifySeals` recovers with it (post-fork)
- `Backend.IsPermissionlessAt` gates it; pre-fork keeps the legacy preimage; the block hash stays round-independent

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/blockchain/block_validator.go
+++ b/blockchain/block_validator.go
@@ -252,6 +252,11 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 		if err != nil {
 			return err
 		}
+		// Committed seals bind the round post-permissionless.
+		committers, err := v.sealer.CommittersWithRound(header)
+		if err != nil {
+			return err
+		}
 		proposer, err := v.mValset.GetProposer(blockNum, uint64(round))
 		if err != nil {
 			return err
--- a/blockchain/block_validator.go
+++ b/blockchain/block_validator.go
@@ -227,19 +227,26 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 	if err != nil {
 		return err
 	}
-	committers, err := v.sealer.Committers(header)
-	if err != nil {
-		return err
-	}
-	if len(committers) == 0 {
-		return istanbul.ErrEmptyCommittedSeals
-	}
 
 	blockNum := header.Number.Uint64()
 
 	var rules params.Rules
 	if v.config != nil {
 		rules = v.config.Rules(new(big.Int).SetUint64(blockNum))
+	}
+
+	// Post-permissionless committed seals bind the round, so recover with the matching preimage.
+	var committers []common.Address
+	if rules.IsPermissionless {
+		committers, err = v.sealer.CommittersWithRound(header)
+	} else {
+		committers, err = v.sealer.Committers(header)
+	}
+	if err != nil {
+		return err
+	}
+	if len(committers) == 0 {
+		return istanbul.ErrEmptyCommittedSeals
 	}
 
 	// Skip module-dependent seal validation when gov/valset modules are not registered.
@@ -249,11 +256,6 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 
 	if rules.IsPermissionless {
 		round, err := v.sealer.Round(header)
-		if err != nil {
-			return err
-		}
-		// Committed seals bind the round post-permissionless.
-		committers, err := v.sealer.CommittersWithRound(header)
 		if err != nil {
 			return err
 		}
@@ -296,14 +298,12 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 		return consensus.ErrUnauthorized
 	}
 
-	signerSet := qualifiedSet.Copy()
-	if !rules.IsPermissionless {
-		council, err := v.mValset.GetCouncil(blockNum)
-		if err != nil {
-			return err
-		}
-		signerSet = valset.NewAddressSet(council).Copy()
+	// Reached only pre-permissionless (post-fork returns above); count seals against the council.
+	council, err := v.mValset.GetCouncil(blockNum)
+	if err != nil {
+		return err
 	}
+	signerSet := valset.NewAddressSet(council).Copy()
 	validSeal, err := countValidCommittedSeals(committers, signerSet)
 	if err != nil {
 		return err
--- a/blockchain/block_validator.go
+++ b/blockchain/block_validator.go
@@ -235,13 +235,8 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 		rules = v.config.Rules(new(big.Int).SetUint64(blockNum))
 	}
 
-	// Post-permissionless committed seals bind the round, so recover with the matching preimage.
-	var committers []common.Address
-	if rules.IsPermissionless {
-		committers, err = v.sealer.CommittersWithRound(header)
-	} else {
-		committers, err = v.sealer.Committers(header)
-	}
+	// Committers is fork-aware: post-permissionless it recovers with the round-bound preimage.
+	committers, err := v.sealer.Committers(header)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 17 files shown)


consensus/istanbul/backend/backend.go
```diff
@@ -145,6 +145,13 @@ func (sb *backend) NodeType() common.ConnType {
 	return sb.nodetype
 }
 
+func (sb *backend) IsPermissionlessAt(num uint64) bool {
+	if sb.chain == nil {
```
ian0371: Is this for test? If so, please add a comment

blockchain/block_validator.go
```diff
@@ -252,6 +252,11 @@ func (v *BlockValidator) verifySeals(header *types.Header) error {
 		if err != nil {
 			return err
 		}
+		// Committed seals bind the round post-permissionless.
```
ian0371: Is it ok to leave line 230 calling `v.sealer.Committers`?

blockchain/block_validator_test.go
```diff
@@ -63,6 +63,10 @@ func (s *verifySealsTestSealer) Committers(*types.Header) ([]common.Address, err
 	return s.committers, nil
 }
 
+func (s *verifySealsTestSealer) CommittersWithRound(*types.Header) ([]common.Address, error) {
```
ian0371: Could you add `verifySeals` pre-HF and post-HF unit tests?

consensus/istanbul/core/commit_test.go
```diff
@@ -96,6 +96,7 @@ func TestSendCommitForOldBlockSealMatchesDigest(t *testing.T) {
 	mockBackend := mock_istanbul.NewMockBackend(mockCtrl)
 	mockBackend.EXPECT().Address().Return(sealAddr).AnyTimes()
 	mockBackend.EXPECT().Sealer().Return(istanbul.NewSealerImpl(sealKey)).AnyTimes()
+	mockBackend.EXPECT().IsPermissionlessAt(gomock.Any()).Return(false).AnyTimes()
```
ian0371, line 99: Could you add coverage for the new post-permissionless round-binding?
- `finalizeMessage()` with IsPermissionlessAt=true
- `handleCommit()` with IsPermissionlessAt=true

consensus/istanbul/backend/backend.go
hyunsooda: Removed. https://github.com/kaiachain/kaia/pull/985/commits/785fc7442914e37c2b88de7d7ef9b586e7ed58bc

blockchain/block_validator.go
hyunsooda: Thanks. Refactored. https://github.com/kaiachain/kaia/pull/985/changes/5a6ca45cb9f4410e8f4726470d5c547ddc2392f0
Note that a dead fork-gating has been removed together (https://github.com/kaiachain/kaia/pull/985/changes/5a6ca45cb9f4410e8f4726470d5c547ddc2392f0#diff-27df61836478aecbd859d1b87b0be8da21055c5d36afde50fef1da4d2f04e4d8L300)

blockchain/block_validator_test.go
hyunsooda: Added. [bcc2a6ff6](https://github.com/kaiachain/kaia/pull/985/changes/bcc2a6ff6a3334e27a10d1803b7d3c4a6469f49c)

consensus/istanbul/core/commit_test.go
hyunsooda: Added. https://github.com/kaiachain/kaia/pull/985/commits/49c9d26b4b9f95d683009ff6d62a9c8b76fbbcdc

2dvorak (commented):
Shouldn't we add fork gated behaviors in other sites referencing `Committers()`?
- `api.GetConsensusInfo()`
- `BlockValidator.Preprocess()`
- `nodecmd.decodeExtra()`

hyunsooda:
@2dvorak I made the sealer's `Committers()` fork-aware in [3e89dd826](https://github.com/kaiachain/kaia/pull/985/commits/3e89dd826bc325d61c9b60d209c3749e06c619b5), so `GetConsensusInfo`, block preprocessing, and any future caller recover correctly without their own fork check. `verifySeals` now just calls it too. Thank you.

ian0371 (dismissed):

consensus/istanbul/backend/backend.go
```diff
@@ -145,6 +145,10 @@ func (sb *backend) NodeType() common.ConnType {
 	return sb.nodetype
 }
 
+func (sb *backend) IsPermissionlessAt(num uint64) bool {
```
2dvorak, line 148:
```suggestion
// IsPermissionlessAt reports whether the permissionless fork is enabled at num.
//
// Requires SetChain to have been called: sb.chain is read without a nil guard.
// Start() calls SetChain before starting the core, so the current callers
// (finalizeMessage/handleCommit, which run on the core's handleEvents goroutine)
// always observe a non-nil chain. Callers outside that goroutine must not assume it.
func (sb *backend) IsPermissionlessAt(num uint64) bool {
```
Could we add a comment above `IsPermissionlessAt`? It reads `sb.chain` without a nil guard, and nothing enforces that `SetChain()` ran first except the call order inside `Start()` (engine.go:177 before :181) — correct today, just worth making explicit.

hyunsooda: Adding it here means commenting the other usages too for consistency, which seems redundant given it's trivial.

consensus/engine/sealer.go
```diff
@@ -73,9 +73,20 @@ func (s *dynamicSealer) Committers(header *types.Header) ([]common.Address, erro
 	if header == nil || header.Number == nil {
 		return nil, consensus.ErrUnknownBlock
 	}
+	// Post-permissionless committed seals bind the round; recover with the matching preimage.
+	if s.chainConfig.IsPermissionlessForkEnabled(header.Number) {
```
yoomee1313, line 77: Since `Committers()` now chooses legacy vs round-bound recovery from `chainConfig`, `decode-extra` will always use the pre-permissionless path because it constructs the sealer with `TestChainConfig` (osaka). Maybe we need a flag there to identify the fork mode/block. Could we add a TODO at `decode-extra`?

2dvorak (commented):
The new tests cover the pieces well — the preimage is round-sensitive, the sealer selects it per fork, and validators sign/verify it live. None asserts the verdict, though: that `verifySeals` rejects a post-fork block whose round byte was mutated.

The existing `verifySeals` tests use `verifySealsTestSealer`, a stub returning canned addresses and a hardcoded `Round() = 0`, so no test runs real signature recovery through `verifySeals`. "Mutated round → wrong committers → below quorum → rejected" is inferred across tests rather than executed.

Worth covering because `rules.IsPermissionless` in `verifySeals` and `s.chainConfig.IsPermissionlessForkEnabled(...)` in `dynamicSealer` are two independent fork checks that must agree — if they ever diverge the wrong way, the round silently stops being bound and every test stays green.

consensus/engine/sealer.go
hyunsooda: Thanks, #1000 bumps `TestChainConfig` to the latest fork (permissionless), which was missing. Once both PRs are in, decode-extra takes the permissionless path, and I added a NOTE there saying the reported committers hold only for post-fork headers.

hyunsooda:
@2dvorak Thanks,  https://github.com/kaiachain/kaia/pull/985/changes/23dca4b886e38b5992b435c952da66db8a84c0c6 added

2dvorak (approved):

ian0371 (approved):

PR #994, opened 2026-07-27
kaiax/vrank, node/cn: bound VRankCandidate/Preprepare signatures to fixed size

hyunsooda:
## Proposed changes

`VRankCandidate.Sig`/`BlsSig` and `VRankPreprepare.Sig` were unbounded `[]byte`. A peer could pad them up to the generic 12 MiB protocol limit, and the receiver decoded the whole payload before any length check — a cheap (well-compressing), repeatable allocation load against consensus nodes.

They are now fixed-size `[65]byte` (ECDSA) / `[96]byte` (BLS), so RLP rejects an oversized field at decode — before allocating it — and the peer is dropped on the decode error. The now-redundant non-empty check in `node/cn` is removed.

Wire format is unchanged: RLP encodes `[N]byte` and a same-length `[]byte` identically, so upgraded and non-upgraded nodes still interoperate on valid messages.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

Fixed-size arrays make the signature fields self-validating at RLP decode, so no per-message size limit or explicit length check is needed for these messages.

### DIFF
--- a/kaiax/vrank/impl/handler.go
+++ b/kaiax/vrank/impl/handler.go
@@ -105,8 +105,8 @@ func (v *VRankModule) HandleVRankPreprepare(msg *vrank.VRankPreprepare) error {
 			BlockNumber: block.NumberU64(),
 			Round:       uint8(view.Round.Uint64()),
 			BlockHash:   block.Hash(),
-			Sig:         sig,
-			BlsSig:      blsSig,
+			Sig:         [65]byte(sig),
+			BlsSig:      [96]byte(blsSig),
 		})
 	}
 	return nil
@@ -141,7 +141,7 @@ func (v *VRankModule) HandleVRankCandidate(msg *vrank.VRankCandidate) error {
 	}
 
 	sigHash := v.vrankCandidateSigHash(msg.BlockNumber, msg.Round, msg.BlockHash)
-	sender, err := v.recoverVRankCandidateSender(sigHash, msg.Sig)
+	sender, err := v.recoverVRankCandidateSender(sigHash, msg.Sig[:])
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


ian0371 (dismissed):

kaiax/vrank/impl/handler.go
```diff
@@ -105,8 +105,8 @@ func (v *VRankModule) HandleVRankPreprepare(msg *vrank.VRankPreprepare) error {
 			BlockNumber: block.NumberU64(),
 			Round:       uint8(view.Round.Uint64()),
 			BlockHash:   block.Hash(),
-			Sig:         sig,
-			BlsSig:      blsSig,
+			Sig:         [65]byte(sig),
+			BlsSig:      [96]byte(blsSig),
```
2dvorak:
```suggestion
			Sig:         [crypto.SignatureLength]byte(sig),
			BlsSig:      [blstypes.SignatureLength]byte(blsSig),
```
How about this? Or we could have something like this instead:

```
const (
      sigLen    = crypto.SignatureLength   // 65, ECDSA + recovery id
      blsSigLen = blstypes.SignatureLength // 96, compressed BLS12-381 signature
)
```

It's just a style so please discard if you don't feel like changing.

2dvorak (commented):
Could we add a few tests? The current one covers only oversized `VRankCandidate`.
- empty sig / empty BLS sig — the case the removed `node/cn` length check used to catch
- undersized sig (64 bytes) and BLS sig (95 bytes)
- old `[]byte` shape ↔ new fixed-size shape: same encoded bytes, decodes both directions
- same set for `VRankPreprepare` — it changed too and has no RLP test

kaiax/vrank/impl/handler.go
hyunsooda: All done. PTAL https://github.com/kaiachain/kaia/pull/994/changes/fb39e648419670e765932dbbc7864734aeddf349 

Regarding the regression test for the type shape, I consider it a dangling test since the type is already anchored.

2dvorak (approved):

ian0371 (approved):

PR #997, opened 2026-07-28
p2p: expire multichannel peers that never complete their channels

hyunsooda:
## Proposed changes

A multichannel peer becomes a `Peer` only once every channel arrives. Until then its first connection sits in `CandidateConns`:

- absent from `srv.peers`, so it escapes `MaxPhysicalConnections`, the inbound limit and `peerTargets`
- no read loop, never closed, and still reachable from the server, so the GC cannot reclaim its socket
- released only on successful assembly — no deadline, no eviction

So a peer that opens one channel, declares `Multichannel` and never sends the rest holds a file descriptor for the lifetime of the process, and can repeat with fresh node keys until the node can no longer accept or dial.

Fix:

- candidates carry an open time and are dropped, sockets and all, past the assembly deadline
- a channel replacing an occupied slot closes the one it replaces
- a `portOrder` outside our listener set is rejected instead of silently dropped
- shutdown closes whatever is still under assembly

Expiry runs at the start of `handleAddPeerConn`, not on a timer: p2p has no periodic loop, and the call that needs the room is the one that reclaims it.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/networks/p2p/server_multi.go
+++ b/networks/p2p/server_multi.go
@@ -71,10 +71,11 @@ func (srv *MultiChannelServer) Stop() {
 	}
 
 	// Peers under assembly have no read loop to notice the shutdown.
+	var candidates []*conn
 	for id, connSet := range srv.CandidateConns {
 		for _, c := range connSet {
 			if c != nil {
-				c.close(DiscQuitting)
+				candidates = append(candidates, c)
 			}
 		}
 		delete(srv.CandidateConns, id)
@@ -85,6 +86,10 @@ func (srv *MultiChannelServer) Stop() {
 
 	// Unlock here to allow Server loops and dialsched loop to finish their remaining jobs, dialsched to finish its dial goroutines.
 	srv.lock.Unlock()
+
+	for _, c := range candidates {
+		c.close(DiscQuitting)
+	}
 
 	if srv.dialSched != nil {
 		srv.dialSched.Close() // Wait for dial attempts to finish.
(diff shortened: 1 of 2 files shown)


networks/p2p/server_multi.go
```diff
@@ -66,6 +70,17 @@ func (srv *MultiChannelServer) Stop() {
 		listener.Close()
 	}
 
+	// Peers under assembly have no read loop to notice the shutdown.
+	for id, connSet := range srv.CandidateConns {
+		for _, c := range connSet {
+			if c != nil {
+				c.close(DiscQuitting)
```
ian0371: `c.close` can take up to 1s. Can it be done without holding the `srv.lock`?

networks/p2p/server_multi.go
```diff
@@ -338,22 +354,30 @@ func (srv *MultiChannelServer) handleAddPeerConn(c *conn) error {
 	if !srv.running {
 		return errServerStopped
 	}
+	srv.expireCandidates(mclock.Now())
```
ian0371: `expireCandidates` calls `c.close` and has the same issue when called with `srv.lock` held.

networks/p2p/server_multi.go
```diff
@@ -338,22 +354,30 @@ func (srv *MultiChannelServer) handleAddPeerConn(c *conn) error {
...
 		if connSet == nil {
 			connSet = make([]*conn, len(srv.ListenAddrs))
 			srv.CandidateConns[c.id] = connSet
+			srv.candidateSince[c.id] = mclock.Now()
 		}
 
-		if int(c.portOrder) < len(connSet) {
-			connSet[c.portOrder] = c
+		if old := connSet[c.portOrder]; old != nil {
+			old.close(DiscAlreadyConnected)
```
ian0371: same here

networks/p2p/server_multi.go
hyunsooda: Fixed at https://github.com/kaiachain/kaia/pull/997/changes/37ef50f0e1faa3bc91d20eeffd5036617e197380. Thanks.

networks/p2p/server_multi.go
hyunsooda: Fixed at https://github.com/kaiachain/kaia/pull/997/changes/37ef50f0e1faa3bc91d20eeffd5036617e197380. Thanks.

networks/p2p/server_multi.go
hyunsooda: Fixed at https://github.com/kaiachain/kaia/pull/997/changes/37ef50f0e1faa3bc91d20eeffd5036617e197380. Thanks.

ian0371 (approved):

2dvorak (approved):

PR #1009, opened 2026-08-04
fix(blockchain/types): bound transaction signature lists

da1suk8:
## Proposed changes

- Limit sender and fee-payer signature lists to `accountkey.MaxNumKeysForMultiSig`.
- Reject oversized lists before signature validation or public-key recovery.
- Add boundary tests for RLP and JSON decoding.

## Types of changes

- [x] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

Tested with:

```sh
go test ./blockchain/types -count=1
```

### DIFF
--- a/api/api_kaia_transaction.go
+++ b/api/api_kaia_transaction.go
@@ -615,6 +615,8 @@ func (s *KaiaTransactionAPI) RecoverFromTransaction(ctx context.Context, encoded
 	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
 		return common.Address{}, err
 	}
+	// NOTE: Deliberately not fork-gated because the block number is
+	// caller-supplied, so gating on it would let the caller bypass the limit.
 	if err := tx.ValidateSignatureListLength(); err != nil {
 		return common.Address{}, err
 	}
--- a/api/api_kaia_transaction.go
+++ b/api/api_kaia_transaction.go
@@ -615,6 +615,9 @@ func (s *KaiaTransactionAPI) RecoverFromTransaction(ctx context.Context, encoded
 	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
 		return common.Address{}, err
 	}
+	if err := tx.ValidateSignatureListLength(); err != nil {
+		return common.Address{}, err
+	}
 
 	state, header, err := s.b.StateAndHeaderByNumber(ctx, blockNumber)
 	if err != nil {
--- a/blockchain/types/tx_signatures_test.go
+++ b/blockchain/types/tx_signatures_test.go
@@ -24,7 +24,6 @@ import (
 	"testing"
 
 	"github.com/kaiachain/kaia/blockchain/types/accountkey"
-	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/kerrors"
 	"github.com/kaiachain/kaia/rlp"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


2dvorak (changes requested):
`SanityCheckSignatures` is called from `DecodeRLP`, which runs on every block body decode, and `RecoverPubkey` is reached from `ValidateSender` during block execution. An unconditional cap in these two places changes how pre-Istanbul blocks decode and execute, so a node syncing from genesis could stop on such a block. And it can't be fixed by adding a fork gate here, since `DecodeRLP` has no block number to gate on.

da1suk8:
@2dvorak 

Thanks, you're right. These are consensus/history paths, so the cap must not be unconditional here.

I’ll remove the checks from DecodeRLP/SanityCheckSignatures and RecoverPubkey, and enforce the bound at tx-pool and RPC ingress before sender recovery. Any consensus-path check would need to live at a block-number-aware validation point, not in DecodeRLP.

api/api_kaia_transaction.go
```diff
@@ -615,6 +615,9 @@ func (s *KaiaTransactionAPI) RecoverFromTransaction(ctx context.Context, encoded
 	if err := rlp.DecodeBytes(encodedTx, tx); err != nil {
 		return common.Address{}, err
 	}
+	if err := tx.ValidateSignatureListLength(); err != nil {
```
2dvorak, line 620:
```suggestion
	// NOTE: Deliberately not fork-gated because the block number is
	// caller-supplied, so gating on it would let the caller bypass the limit.
	if err := tx.ValidateSignatureListLength(); err != nil {
```

2dvorak: Without fork gate this may return wrong result for pre-Istanbul blocks, however adding fork gate would allow the caller to bypass the limit. So it would be better to leave a comment about that here instead of adding a fork gate.

da1suk8: Thanks, added a comment to clarify why this check is deliberately not fork-gated.

2dvorak (dismissed):

blockchain/types/tx_signatures_test.go
```diff
@@ -0,0 +1,112 @@
...
+	return sigs
+}
+
+func TestSanityCheckSignaturesLength(t *testing.T) {
+	maxSignatures := int(accountkey.MaxNumKeysForMultiSig)
+
+	require.False(t, SanityCheckSignatures(nil, TxTypeValueTransfer))
+	require.True(t, SanityCheckSignatures(validTestSignatures(1), TxTypeValueTransfer))
+	require.True(t, SanityCheckSignatures(validTestSignatures(maxSignatures), TxTypeValueTransfer))
+	require.True(t, SanityCheckSignatures(validTestSignatures(maxSignatures+1), TxTypeValueTransfer))
```
ian0371, line 52:
```suggestion
	require.True(t, SanityCheckSignatures(validTestSignatures(maxSignatures+1), TxTypeValueTransfer)) // // no length cap here (by design); enforced at ingress
```

ian0371 (dismissed):

blockchain/types/tx_signatures_test.go
da1suk8: Applied, thank you. I moved it above the line and added the reason.

2dvorak (approved):

ian0371 (approved):

PR #1012, opened 2026-08-04
node/cn: verify a blob tx sidecar only once

hyunsooda:
## Proposed changes

A replayed blob transaction is decoded into a fresh sidecar object, so the sidecar's own validation cache never hits and the handler verified its KZG proofs again on every copy. The handler now skips verification for blob tx hashes it already verified; the tx pool still verifies the sidecar independently before admitting the transaction.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -99,7 +99,25 @@ var (
 	errUnknownProcessingError  = errors.New("unknown error during the msg processing")
 	errUnsupportedEnginePolicy = errors.New("unsupported engine or policy")
 	errKZGVerificationError    = errors.New("KZG verification error")
+	errBloblessBlobTx          = errors.New("blobless blob transaction")
 )
+
+// blobSidecarKey identifies what ValidateWithBlobHashes consumes. Commitments are left
+// out because each hash is the sha256 of one, so the hashes already pin them.
+func blobSidecarKey(hashes []common.Hash, sc *types.BlobTxSidecar) common.Hash {
+	parts := make([][]byte, 0, len(hashes)+1+len(sc.Blobs)+len(sc.Proofs))
+	for i := range hashes {
+		parts = append(parts, hashes[i][:])
+	}
+	parts = append(parts, []byte{sc.Version})
+	for i := range sc.Blobs {
+		parts = append(parts, sc.Blobs[i][:])
+	}
+	for i := range sc.Proofs {
+		parts = append(parts, sc.Proofs[i][:])
+	}
+	return crypto.Keccak256Hash(parts...)
+}
 
 func errResp(code errCode, format string, v ...interface{}) error {
 	return fmt.Errorf("%v - %v", code, fmt.Sprintf(format, v...))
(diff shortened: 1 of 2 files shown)


node/cn/handler_msg_test.go
```diff
@@ -335,17 +330,37 @@ func TestHandleTxMsg_KZGVerificationError(t *testing.T) {
...
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	pm, mockPeer, blobTx := prepareBlobTxMsg(t, mockCtrl)
+	require.NoError(t, handleTxMsg(pm, mockPeer, generateMsg(t, TxMsg, types.Transactions{blobTx})))
+	require.True(t, pm.verifiedBlobTxs.Contains(blobTx.Hash()))
+
+	// The tx hash does not cover the sidecar, so the replay carries a broken proof
+	// under the same hash and passes only because the verification is skipped.
+	blobTx.BlobTxSidecar().Proofs[0][0] ^= 0xFF
```
ian0371: The cache key doesn't cover the sidecar, so this happens. Is this ok?

2dvorak (commented):
1. `tx.Hash()` as the key is bypassable at zero cost. The expensive work is per-sidecar, but the key is per-transaction, and the attacker controls that mapping: one precomputed sidecar attaches to unlimited distinct transactions by bumping a nonce — identical sidecar bytes, different tx hashes — so every one of them misses the set and pays a full verification.
2. Reject blob txs with zero blob hashes. `ValidateWithBlobHashes` passes vacuously when there are no blobs so a 113-byte blob tx with an empty sidecar validates. The pool rejects these unconditionally (`tx_pool.go:1071`), so the handler accepts and forwards a transaction that can never be valid and never disconnects the peer for it. It also mints a cache entry per tx, cheap enough to flush the whole set for ~113 KiB — though that part goes away once the key changes per (1).

hyunsooda:
@2dvorak The key now covers what the verification consumes, and rejects blobless tx earlier before the veficiation runs.

node/cn/handler_msg_test.go
hyunsooda: Changed to `keccak(blobHashes ‖ Version ‖ Blobs ‖ Proofs) `

node/cn/handler.go
```diff
@@ -96,8 +99,26 @@ var (
 	errUnknownProcessingError  = errors.New("unknown error during the msg processing")
 	errUnsupportedEnginePolicy = errors.New("unsupported engine or policy")
 	errKZGVerificationError    = errors.New("KZG verification error")
+	errBloblessBlobTx          = errors.New("blobless blob transaction")
 )
 
+// blobSidecarKey identifies what ValidateWithBlobHashes consumes. Commitments are left
+// out because each hash is the sha256 of one, so the hashes already pin them.
```
2dvorak: Commitments are excluded from the key on the grounds that the blob hashes already pin them — but the check that enforces that binding, `ValidateBlobCommitmentHashes`, is exactly what a cache hit skips. So an attacker can swap the commitment of an already-verified sidecar: the key still matches, validation is skipped, and the tx reaches the pool with the peer left connected. No CPU is burnt — the skip is the cheap path — but the misbehavior signal is silenced, where dev today disconnects on any malformed sidecar.

In addition, prefixing the element counts would make the key encoding unambiguous.

What do you think?

hyunsooda: Both fixed in [afa66aa33](https://github.com/kaiachain/kaia/pull/1012/commits/afa66aa33007da6f3addd0dd3400945426f15b7e). Commitments are in the key and the element counts are prefixed, so the key is now an injective encoding of every input `ValidateWithBlobHashes` reads.

2dvorak (approved):

ian0371 (approved):

PR #1013, opened 2026-08-05
discover: admit only CNPeers members into the CN storage

hyunsooda:
## Proposed changes

The node type in a discovery packet is supplied by the sender and the CN storage is unbounded, so any reachable key pair can fill the list the dial scheduler samples for CN candidates. The CN storage now accepts an entry only when its address is in `CNPeers` or its NodeId is a configured bootnode, bounding the list by an authenticated set instead of a capacity limit. Discovery stays open elsewhere: the filter is off before the hard fork and whenever `CNPeers` is unavailable, bootnodes are exempt so they keep seeding CN lookups, and peer admission is unchanged.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/networks/p2p/discover/discovery.go
+++ b/networks/p2p/discover/discovery.go
@@ -20,6 +20,7 @@ import (
 	"crypto/ecdsa"
 	"net"
 
+	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/networks/p2p/netutil"
 	"github.com/kaiachain/kaia/networks/rpc"
 )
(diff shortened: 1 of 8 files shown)


ian0371 (dismissed):

2dvorak (commented):
We could have better coverage with the following test cases:

1. `BaseServer` → `ntab forward` — `nil` ntab in both existing tests means the three new added lines are dead to CI.
2. Table-level shrink — `dropsOutsiders` only covers `nil → {member}`, i.e. startup. No allowlist-to-allowlist case exists.

And one minor thing is: `TestServerENBypassesCNPeerFilter` says `SetCNPeers` is no-op for PN/EN, but it became stale comment.

hyunsooda:
@2dvorak All added. Thank you

2dvorak (approved):

ian0371 (approved):

PR #1019, opened 2026-08-06
vrank: read epoch candidates from the header when state is gone

hyunsooda:
## Proposed changes

`newCPMatrix` seeded the CP matrix from `GetCandTesting`, which reads node states
one block below its argument — so an epoch block's transition needed the state two
blocks below it, which a node that synced the range without executing it does not
have. The epoch-start header carries the same candidate list and is now used when
the state read fails.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/kaiax/vrank/impl/consensus.go
+++ b/kaiax/vrank/impl/consensus.go
@@ -65,7 +65,7 @@ func (v *VRankModule) VerifyHeader(header *types.Header, _ *types.Header) error 
 	}
 	// Failures score against the reporter's own byzantine-filterable column regardless of content,
 	// so only the failed list is checked (CandTesting is epoch-stable within the epoch).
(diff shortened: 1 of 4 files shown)


da1suk8 (dismissed):

ian0371 (commented):
Should we also address this line? https://github.com/kaiachain/kaia/blob/f4cea3815bc67202580625d05f2aca02b733f73b/kaiax/vrank/impl/consensus.go?plain=1#L68

hyunsooda:
@ian0371 No need to apply the same fallback there. During on non-epoch duration, the number `n` looks correct, instead of `n-1`.

ian0371 (approved):

da1suk8 (approved):

PR #1023, opened 2026-08-07
metrics: serve Prometheus /metrics on a private mux

2dvorak:
## Proposed changes

- Starting the exporter with a `nil` handler served `http.DefaultServeMux`, exposing net/http/pprof's /debug/pprof/* (including cmdline) on the Prometheus port even with `--pprof` off
- Use a private mux; listen address unchanged. Add a regression test.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/metrics/utils/metrics_test.go
+++ b/metrics/utils/metrics_test.go
@@ -1,0 +1,99 @@
+// Copyright 2026 The Kaia Authors
+// This file is part of the Kaia library.
+//
+// The Kaia library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The Kaia library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the Kaia library. If not, see <http://www.gnu.org/licenses/>.
+
+package metricutils
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"net/http"
+	_ "net/http/pprof" // registers /debug/pprof/* on http.DefaultServeMux, like the node binaries
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v2"
+)
+
+// freePort returns an ephemeral TCP port that is free at call time.
+func freePort(t *testing.T) int {
+	t.Helper()
+	var lc net.ListenConfig
+	l, err := lc.Listen(context.Background(), "tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	port := l.Addr().(*net.TCPAddr).Port
+	require.NoError(t, l.Close())
+	return port
+}
+
+// httpGet issues a context-aware GET and returns the response.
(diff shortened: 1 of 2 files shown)


hyunsooda (dismissed):

metrics/utils/metrics_test.go
```diff
@@ -0,0 +1,99 @@
...
+// /metrics but not net/http/pprof's /debug/pprof/*, guarding against a regression
+// to a nil (DefaultServeMux) handler that would leak them on the Prometheus port.
+func TestPrometheusExporterDoesNotExposePprof(t *testing.T) {
+	// Enabled/EnabledPrometheusExport are normally set by init() from os.Args.
+	prevEnabled, prevExport := Enabled, EnabledPrometheusExport
+	Enabled, EnabledPrometheusExport = true, true
+	t.Cleanup(func() { Enabled, EnabledPrometheusExport = prevEnabled, prevExport })
+
+	port := freePort(t)
+	StartMetricCollectionAndExport(newExporterCtx(port))
```
yoomee1313: It's a trivial, but the metric server doesn't stop when we run repeated test runs. for example, `go test run -count=200` or `race`. 
Why don't you run this test in child goroutine and close after test finished?

edwin082 (commented):
The private-mux change looks correct to me. I found two issues in the newly added regression test on Go 1.25.3 (macOS/arm64):

- **Data race:** cleanup writes `Enabled` at `metrics_test.go:67` while `CollectProcessMetrics` reads it at `metrics.go:93`.
- **Duplicate registration panic:** repeated test runs leave background metric updaters sharing the same Prometheus registry, eventually causing `panic: duplicate metrics collector registration attempted`.

To reproduce the data race:

```sh
go test -race ./metrics/utils \
  -run '^TestPrometheusExporterDoesNotExposePprof$' -count=20 -timeout=60s
```

To reproduce the duplicate registration panic:

```sh
go test ./metrics/utils \
  -run '^TestPrometheusExporterDoesNotExposePprof$' -count=200 -timeout=60s
```

The test starts the metrics HTTP server and background collection tasks, but cleanup only restores the flags. Those tasks remain active after the test finishes. These failures concern the new test’s lifecycle, rather than the production routing change.

metrics/utils/metrics_test.go
2dvorak: Thanks, PTAL 89974f6.

metrics/utils/metrics_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"net/http"
+	"net/http/httptest"
+	_ "net/http/pprof" // registers /debug/pprof/* on http.DefaultServeMux, like the node binaries
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPrometheusExporterDoesNotExposePprof checks that the exporter serves
```
edwin082: Approved

---
nit: After the refactor, the test starts an `httptest` server with `newPrometheusMux()` instead of calling `StartMetricCollectionAndExport`. It verifies the mux’s routes, but would still pass if the production server’s `Handler` were changed back to `nil`. The comment would be more accurate with the narrower scope:

```go
// TestPrometheusExporterDoesNotExposePprof verifies that newPrometheusMux
// serves /metrics and does not expose the pprof routes registered on
// http.DefaultServeMux.
```

edwin082 (dismissed):

metrics/utils/metrics_test.go
2dvorak: Thanks, PTAL 3a1aba76b

edwin082 (approved):

ian0371 (approved):

yoomee1313 (approved):

PR #1024, opened 2026-08-10
datasync/fetcher: bound the memory held by the block import queue

hyunsooda:
## Proposed changes

Problem

- `handleNewBlockMsg` queues a decoded block before its header, parent, body or size is checked.
- The queue admits on counts only — `blockLimit` per peer, `maxQueueDist` ahead — with no bound on the bytes they hold.
- Entries are not released when the peer disconnects, so a new identity gets a fresh allowance while the old blocks stay.

Fix

- Track the encoded size of the queued blocks and refuse propagated ones once it exceeds 64 MiB, the budget the downloader already uses for block caching.
- Release a peer's entries when it disconnects.
- Exempt blocks the consensus engine committed: a validator that did not propose imports its own committed block through this queue.
- Bound `newBlockData.TD`, an unbounded `big.Int` that `SetHead` keeps on the peer (second commit).

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/datasync/fetcher/fetcher.go
+++ b/datasync/fetcher/fetcher.go
@@ -51,6 +51,10 @@ const (
 var (
 	errTerminated = errors.New("terminated")
 	logger        = log.NewModuleLogger(log.DatasyncFetcher)
+
+	// queuedBytesLimit caps the encoded size of the blocks held in the import queue,
+	// matching the downloader's blockCacheMemory. blockLimit alone bounds the count.
+	queuedBytesLimit uint64 = 64 * 1024 * 1024
 )
 
 // blockRetrievalFn is a callback type for retrieving a block from the local chain.
@@ -111,8 +115,9 @@ type bodyFilterTask struct {
 
 // inject represents a schedules import operation.
 type inject struct {
-	origin string
-	block  *types.Block
+	origin  string
+	block   *types.Block
+	trusted bool // locally produced by consensus; exempt from the queue byte budget
 }
 
 // Fetcher is responsible for accumulating block announcements from various peers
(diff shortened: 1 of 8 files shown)


datasync/fetcher/fetcher.go
```diff
@@ -391,7 +415,20 @@ func (f *Fetcher) loop() {
...
 			propBroadcastInMeter.Mark(1)
-			f.enqueue(op.origin, op.block)
+			f.enqueue(op.origin, op.block, op.trusted)
+
+		case peer := <-f.forgetPeer:
+			// The peer is gone; release what it queued instead of holding it until
+			// the chain reaches those heights.
+			for hash, op := range f.queued {
+				if op.origin == peer {
+					f.forgetBlock(hash)
```
yoomee1313, line 425: Even if a committed block is later delivered from the local consensus layer via `EnqueueTrusted`, if the same hash was already queued from a peer, the duplicate path simply ignores the new enqueue and does not record that the block is trusted. As a result, when the peer disconnects, `ForgetPeer` can remove the queued block and the locally committed block can be lost from the fetcher queue.

hyunsooda: Right, fixed in [cad7aacd](https://github.com/kaiachain/kaia/pull/1024/changes/cad7aacdfb04802b82713fecdfd2c9528b56f746) by carrying the mark over and skipping trusted blocks on disconnect. PTAL. Thank you.

yoomee1313 (approved):

ian0371 (approved):

PR #1025, opened 2026-08-10
node/cn: bound the length of decoded response lists

hyunsooda:
## Proposed changes

Problem

- The synchronisation response handlers decoded a peer's list in full before asking whether it answered anything.
- The message cap bounds the encoded packet, not the object graph reflection builds from it: a 12 MiB `NodeDataMsg` of one-byte elements decodes to 12.5 million slice headers.
- A delivery that finds no pending request is only logged, so the packet can be repeated on the same connection.

Fix

- Decode these lists one element at a time and refuse one longer than the number we could have requested, reusing the limit each message already applies when it is requested and when it is served.
- A longer list cannot come from an implementation of this protocol, so the handler returns an error and the peer is dropped by the existing path.
- Deliveries that find no pending request keep being logged: without request IDs a late but well-formed response is indistinguishable from an unsolicited one.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/node/cn/handler.go
+++ b/node/cn/handler.go
@@ -949,12 +949,35 @@ func handleBlockHeadersRequestMsg(pm *ProtocolManager, p Peer, msg p2p.Msg) erro
 	return p.SendBlockHeaders(headers)
 }
 
+// decodeResponseList decodes a response list one element at a time, refusing a list
+// longer than limit. Decoding the whole list at once lets a peer turn a message-sized
+// packet into an arbitrarily larger object graph, and we never request more than limit.
+func decodeResponseList[T any](msg p2p.Msg, limit int) ([]T, error) {
+	stream := rlp.NewStream(msg.Payload, uint64(msg.Size))
+	if _, err := stream.List(); err != nil {
+		return nil, errResp(ErrDecode, "msg %v: %v", msg, err)
+	}
+	var items []T
+	for {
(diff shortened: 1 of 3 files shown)


node/cn/handler.go
```diff
@@ -952,12 +949,35 @@ func handleBlockHeadersRequestMsg(pm *ProtocolManager, p Peer, msg p2p.Msg) erro
 	return p.SendBlockHeaders(headers)
 }
 
+// decodeResponseList decodes a response list one element at a time, refusing a list
+// longer than limit. Decoding the whole list at once lets a peer turn a message-sized
+// packet into an arbitrarily larger object graph, and we never request more than limit.
+func decodeResponseList[T any](msg p2p.Msg, limit int) ([]T, error) {
```
yoomee1313, line 955: `BlobSidecarsMsg` may not amplify memory as much as `NodeDataMsg`, but it still has the same problem this PR is trying to remove.
Can we also apply the same response-list limit to `BlobSidecarsMsg` as well?

hyunsooda: Missed one. Thank you.

yoomee1313 (approved):

ian0371 (approved):

PR #1034, opened 2026-08-12
node/cn: derive the CN peer allowlist from the canonical head

hyunsooda:
## Proposed changes

Problem

- The sync loop derived the allowlist from the block in the `ChainHeadEvent`. Those events are posted after the insertion lock is released, so an older one can arrive after a newer one and overwrite the current allowlist with a stale one.

Fix

- Read the canonical head instead. The event is only a signal that the head moved.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/node/cn/cnpeers.go
+++ b/node/cn/cnpeers.go
@@ -145,8 +145,8 @@ func (s *CN) cnPeerSyncLoop(ch <-chan blockchain.ChainHeadEvent, sub event.Subsc
 
 	for {
 		select {
-		case ev := <-ch:
-			s.cnPeerUpdater.syncHead(ev.Block)
+		case <-ch:
+			s.cnPeerUpdater.syncHead(s.blockchain.CurrentBlock())
 		case err, ok := <-sub.Err():
 			if ok && err != nil {
 				logger.Warn("CN peer sync stopped", "err", err)
--- a/kaiax/valset/impl/error.go
+++ b/kaiax/valset/impl/error.go
@@ -31,6 +31,7 @@ var (
 	errNoVoteBlockNums        = errors.New("no validator vote block nums")
 	errMismatchedValidators   = errors.New("header extra validators do not match qualified validators")
 	errPermissionlessDisabled = errors.New("permissionless fork is not enabled")
+	errVRankModuleNotSet      = errors.New("VRankModule is not set")
 
 	// rpc related errors
 	errPendingNotAllowed       = errors.New("pending is not allowed")
--- a/kaiax/valset/impl/transition.go
+++ b/kaiax/valset/impl/transition.go
@@ -119,16 +119,22 @@ func (v *ValsetModule) applyTransition(header *types.Header, statedb *state.Stat
 	}
 	abv2result.Nodes.MarkSuspended(abv2result.SuspendedValidators)
 
-	ctx := v.newTransitionContext(header, abv2result)
+	ctx, err := v.newTransitionContext(header, abv2result)
+	if err != nil {
+		return nil, err
+	}
 	return ctx.ApplyAllTransitions(abv2result.Nodes), nil
 }
 
-func (v *ValsetModule) newTransitionContext(header *types.Header, abv2result *system.ABv2Snapshot) *TransitionContext {
+func (v *ValsetModule) newTransitionContext(header *types.Header, abv2result *system.ABv2Snapshot) (*TransitionContext, error) {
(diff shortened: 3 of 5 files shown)


node/cn/cnpeers.go
```diff
@@ -145,8 +145,8 @@ func (s *CN) cnPeerSyncLoop(ch <-chan blockchain.ChainHeadEvent, sub event.Subsc
 
 	for {
 		select {
-		case ev := <-ch:
-			s.cnPeerUpdater.syncHead(ev.Block)
+		case <-ch:
+			s.cnPeerUpdater.syncHead(s.blockchain.CurrentBlock())
```
yoomee1313, line 149: There is a small chance that `CurrentHeader` and `CurrentBlock` can be inconsistent. The problem is that the call path from `syncHead` can refer to `CurrentHeader`.
To avoid this situation, what about checking the consistency between `CurrentHeader` and `CurrentBlock`? Once the block is fully inserted, it will be indirectly retried by the next `ChainHeadEvent`.

2dvorak (dismissed):

ian0371:
Regarding the `CurrentHeader`/`CurrentBlock` inconsistency raised above: the window is real (`insert()` stores `currentBlock` before `SetCurrentHeader`), but the damage it can cause lives in the valset module, not in this loop. `GetCNPeers(num)` computes the node-state transition via `getTransitionResult`, whose `fetchVRankCtx` absorbs vrank read errors into an empty context. If the sync loop reads `CurrentBlock()` inside that window — and block N had a proposal failure (round > 0) or N+1 is a vrank epoch block — `GetPFS`/`GetCFS` return `ErrFutureBlock` (they gate on `CurrentHeader`), the error is swallowed, and a transition computed without vrank evidence is cached in `transitionResultCache`, the same cache `writeTransitionToABv2` consumes during block processing. A microsecond race could then permanently pin a wrong transition for that height: the affected node later computes a state root for N+1 that disagrees with the rest of the network. This PR makes the window reachable from the sync loop on every block, so it should not merge without the fix.

```
insert(block N)                          cnPeerSyncLoop (this PR)
---------------                          ------------------------
currentBlock.Store(N)      --+
                             |  window:  CurrentBlock() -> N
                             |           GetCNPeers(N+1)
                             |            +- fetchVRankCtx(N)
                             |                +- GetPFS(N): N > CurrentHeader (= N-1)
                             |                   -> ErrFutureBlock, swallowed -> PFS = nil
SetCurrentHeader(N)        --+            +- transitionResultCache.Add(N+1, degraded)

later, when the affected node processes block N+1 (InitializeState, before user txs):
writeTransitionToABv2(N+1) -> transitionResultCache hit (degraded)
                           -> wrong ABv2 system tx -> state root differs from other nodes
```

Please merge this patch into the PR: https://gist.github.com/ian0371/90b12da2d4f5f4be688fa668e24ef45e (`git am` on the branch). It makes `fetchVRankCtx` propagate errors instead of absorbing them; `getTransitionResult` already skips caching on error, so a window hit now degrades to `GetCNPeers` returning nil, the updater keeps the last allowlist, and the next `ChainHeadEvent` refreshes it. That is the same outcome as the consistency check suggested above (skip the sync when `CurrentHeader` is behind `CurrentBlock`), and it also covers other transient vrank failures (for example during reorgs), so no additional guard is needed in the sync loop. One behavior change to be aware of: on the block-processing path a genuine vrank read failure now fails the block loudly (the ABv2 system tx is skipped or the insertion aborts) instead of committing a silently degraded transition — safer, because nothing wrong is cached and a retry recomputes. Tests are included, and `kaiax/valset`, `kaiax/vrank`, and `node/cn` pass.

node/cn/cnpeers.go
hyunsooda: Thanks for the catch. Went with ian's patch instead of a check here. The RPC getters reach the same path, so a guard in this loop wouldn't be enough. Same effect as what you suggested though: nil comes back, last allowlist is kept, next event refreshes.

hyunsooda:
@ian0371 Applied at https://github.com/kaiachain/kaia/pull/1034/changes/dd56622fc0bc46cc4d67d714dc1c9192c75dce15 Thanks.

ian0371 (approved):

yoomee1313 (approved):

PR #1035, opened 2026-08-12
p2p: fix dial scheduler and multichannel dial cleanup

yoomee1313:
## Proposed changes

This PR addresses p2p dial issues introduced by the DialSched refactor.

It contains three fixes:

063bb7274 p2p: serialize dial candidate collection
- Fixes a dial scheduler race where `getCandidates` read scheduler bookkeeping across inconsistent lock snapshots.
- Keeps candidate demand calculation, static peer snapshot, dynamic queue refill, and CN allowlist filtering under one scheduler lock.
- This makes each dial candidate batch reflect one consistent view of connected/dialing/static/CN peer state.

b0690b2b2 p2p: preserve connected peers on dial failure
- Fixes a simultaneous dial case where outbound dial failure could remove bookkeeping for an already live inbound peer.
- markDialFailure now returns early when the peer is already present in connectedAll.
- This prevents failed outbound attempts from tearing down live peer accounting.

7fb265652 p2p: clean up incomplete multichannel dials
- Fixes partial multichannel dial cleanup when some channels succeed but the full peer is never assembled.
- dialMulti now verifies peer assembly after all channels, cleans only outbound pending candidates, and preserves inbound candidates from simultaneous dials.
- Incomplete multichannel dials no longer remove static peers, and the warning log includes expected/dialed/cleaned channel counts.

One known limitation remains: multichannel candidate assembly is still keyed only by peer ID, so inbound and outbound candidate state can share the same candidate set. This is an existing design limitation rather than a regression introduced by this PR, so it is left out of scope.

## Types of changes

- [ ] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/networks/p2p/dialsched_test.go
+++ b/networks/p2p/dialsched_test.go
@@ -613,11 +613,13 @@ func TestDialSched_DialMulti(t *testing.T) {
 	var (
 		multiNode = discover.NewNode(testNodeID(401), net.ParseIP("10.0.0.41"), 30303, 30303, []uint16{30304, 30305}, discover.NodeTypeEN)
 		md        = &mockDialer{dialErr: make(map[discover.NodeID]error)}
-		mb        = &mockSetupBackend{setupErr: make(map[discover.NodeID]error)}
+		mb        = &multiDialLifecycleBackend{}
 		ds        = NewDialSched(DialConfig{}, nil, mb)
 	)
-	ds.dialer = md
+	t.Cleanup(mb.Close)
 	mb.dialsched = ds
+	mb.connectOnPort = uint16(len(multiNode.TCPs) - 1)
+	ds.dialer = md
 
 	ds.dialOnce(multiNode, dynDialedConn)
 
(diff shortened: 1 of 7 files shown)


2dvorak:
@yoomee1313 Please rebase onto the latest dev and resolve the conflicts. It looks like “p2p: preserve connected peers on dial failure” was already addressed by #1036.

networks/p2p/dialsched_test.go
```diff
@@ -639,6 +641,87 @@ func TestDialSched_DialMulti(t *testing.T) {
 	}
 }
 
+func TestDialSched_DialMultiRollbackOnSetupFailure(t *testing.T) {
+	multiNode := discover.NewNode(testNodeID(402), net.ParseIP("10.0.0.42"), 30303, 30303, []uint16{30304, 30305}, discover.NodeTypeEN)
+	setupErr := errors.New("setup failed")
+	mb := &multiDialLifecycleBackend{
+		setupErrByPort: map[uint16]error{0: setupErr},
```
2dvorak:
```suggestion
		setupErrByPort: map[uint16]error{1: setupErr},
```
Would it be better to fail on port 1 here? That would verify that `CleanConn` closes the channel already accepted on port 0, which seems to be the intended partial-dial rollback scenario.

networks/p2p/dialsched_test.go
```diff
@@ -639,6 +641,87 @@ func TestDialSched_DialMulti(t *testing.T) {
...
+			_ = remote.Close()
+		})
+	}
+	ds := NewDialSched(DialConfig{}, nil, mb)
+	ds.dialer = &fixedMultiDialer{conns: dialConns}
+
+	err := ds.dialMulti(multiNode, dynDialedConn)
+
+	require.ErrorIs(t, err, setupErr)
+	require.Len(t, mb.calls, 1, "SetupConn should stop after the first failure")
```
2dvorak:
```suggestion
	require.Len(t, mb.calls, 2, "SetupConn should stop after the first failure")
```
Failing on port 1 would result in two SetupConn calls.

2dvorak (approved):
One minor thing is that the PR description is stale.

ian0371 (approved):
One structural note for a follow-up, not for this PR: `CleanConn` exists because `DialBackend.SetupConn` takes one fd, so the multichannel "bundle" is known only inside `dialMulti`, and the server has to be told afterwards when a partially accepted bundle failed.
If we later add an outbound-only `SetupMultiConn(fds, flags, dest)` that the server handles all-or-nothing (assemble the peer on full success, close every fd on any failure), `CleanConn` and `errIncompleteMultiChannelDial` go away and `CandidateConns` becomes inbound-only, which also removes the inbound/outbound mixing noted as a known limitation. Worth an issue under the KIP-311 multichannel work.

PR #1043, opened 2026-08-19
blockchain/vm: copy the jump table before editing opcodes in place

hyunsooda:
## Proposed changes

Problem

- `JumpTable` is `[256]*operation`, so assigning a package-level fork table shares its opcode objects instead of copying them.
- `EnableEIP` and `ChangeGasCostForTest` write through those pointers, so one interpreter's edit rewrites the template that every later interpreter on the same fork reads. `EnableEIP` already documents the contract this breaks: "callers need to ensure that the globally defined jump tables are not polluted".
- Only tests reach this today: `Config.ExtraEips` has no CLI or config field, and `ChangeGasCostForTest` is called only from `tests`.

Fix

- Add `copyJumpTable` and deep-copy before both mutation sites.
- Keep the copy conditional, so an interpreter with no extra EIPs shares the table and allocates nothing.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

`copyJumpTable` and the guard on the `ExtraEips` loop are [geth's](https://github.com/ethereum/go-ethereum/blob/v1.17.5/core/vm/jump_table.go#L1115). geth stops there; `ChangeGasCostForTest` is called with no extra EIPs, so it copies inside the function instead. Making `JumpTable` a value array would cover both at once, but it diverges from upstream and pays the copy on every interpreter.

### DIFF
--- a/blockchain/vm/jump_table_test.go
+++ b/blockchain/vm/jump_table_test.go
@@ -1,3 +1,19 @@
+// Copyright 2026 The Kaia Authors
+// This file is part of the Kaia library.
+//
+// The Kaia library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The Kaia library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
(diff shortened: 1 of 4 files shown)


da1suk8 (dismissed):

blockchain/vm/jump_table_test.go
```diff
@@ -0,0 +1,82 @@
+package vm
```
ian0371, line 17: Add the license header please

hyunsooda: Added.

da1suk8 (approved):

yoomee1313 (approved):

PR #1053, opened 2026-08-27
consensus/istanbul: bound backlog and ROUND CHANGE retention

da1suk8:
## Proposed changes

Bounds the memory retained for Istanbul consensus messages, so that it stays bounded regardless of traffic and no sender can crowd out another.

Retained memory = council × per-sender limit × per-message limit, with no budget shared between senders. A shared budget can be filled by a few senders ahead of an honest one, and during a round change the message dropped that way is the next round's PREPREPARE.

Bound every PREPARE, COMMIT and ROUND CHANGE to 1 KiB in handleCheckedMsg, before the retention paths (backlog, roundChangeSet, messageSet) diverge. The envelope is otherwise unbounded: CommittedSeal is only validated for COMMIT, and rlp does not cap big.Int view fields on decode.
Retain backlog messages per sender only: 128 PREPARE / COMMIT / ROUND CHANGE within the current sequence + 8, and one PREPREPARE, replaced by the sender's newest.
Retain ROUND CHANGE only for the current round through current round + 128, with at most a quorum of distinct senders per round.
Release queue storage and accounting once a sender's messages are processed or discarded, and log retention rejections at Trace, since they are expected under load.

Resulting bounds at a council of 50: backlog PREPARE / COMMIT / ROUND CHANGE ≈ 6.4 MiB, c.current.Prepares ≈ 50 KiB (from ≈ 600 MiB), ROUND CHANGE ≈ 4.3 MiB, and backlog PREPREPARE ≈ 500 MiB (council × block size) — the price of guaranteeing every sender a slot. Lowering it would need proposer-aware retention, which is only computable for the current sequence.

The quorum and wire protocol are unchanged.

## Types of changes

- [x] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Related issues

Complements #1029, which bounds Istanbul message ingress concurrency rather than
retained message memory.

## Further comments

Distant ROUND CHANGE entries are rejected rather than evicting retained round
buckets, so that the near-round buckets needed for catch-up are kept. This
intentionally trades catch-up beyond the round window for bounded memory.

Nodes that are behind the backlog sequence window recover through block
synchronization.

### DIFF
--- a/consensus/istanbul/core/backlog.go
+++ b/consensus/istanbul/core/backlog.go
@@ -23,6 +23,8 @@
 package core
 
 import (
+	"math/big"
+
 	"github.com/kaiachain/kaia/common"
 	"github.com/kaiachain/kaia/common/prque"
 	"github.com/kaiachain/kaia/consensus/bft"
@@ -35,6 +37,22 @@ var msgPriority = map[uint64]int{
 	bft.MsgCommit:     2,
 	bft.MsgPrepare:    3,
 }
+
+const (
+	// The limits below bound both the number and payload size of retained future
+	// messages. A PREPREPARE carries an entire block, so a count-only limit is
+	// not sufficient to bound backlog memory.
+	maxBacklogMessagesPerSender     = 128
+	maxBacklogPayloadBytesPerSender = 16 * 1024 * 1024
+	// Global limits allow up to eight senders to reach their per-sender limits.
+	maxBacklogMessages     = 1024
+	maxBacklogPayloadBytes = 128 * 1024 * 1024
+	// Keep only a small future-sequence window so far-future messages cannot
+	// occupy the global backlog budget until the node catches up. A node further
+	// behind than this window catches up through block synchronization rather
+	// than through retained consensus messages.
+	maxBacklogSequencesAhead = 8
+)
 
 // checkMessage checks the message state
 // return bft.ErrInvalidMessage if the message is invalid
@@ -99,25 +117,78 @@ func (c *core) storeBacklog(msg *bft.Message, src common.Address) {
 	defer c.backlogsMu.Unlock()
 
 	backlog := c.backlogs[src]
+	messageBytes := backlogMessageBytes(msg)
+	if c.backlogLimitReached(src, backlog, messageBytes) {
+		// A full backlog is expected under load; avoid a warning for every
+		// dropped message.
+		logger.Trace("Discarding future message: backlog limit reached")
+		return
+	}
+
+	view, err := msg.GetView()
+	if err != nil || view == nil || view.Sequence == nil || view.Round == nil {
+		logger.Trace("Discarding future message: cannot decode view", "err", err)
+		return
+	}
+	if c.isBacklogSequenceTooFar(view.Sequence) {
+		logger.Trace("Discarding future message: sequence is too far ahead", "sequence", view.Sequence)
+		return
+	}
 	if backlog == nil {
 		backlog = prque.New()
+		c.backlogs[src] = backlog
 	}
-	switch msg.Code {
-	case bft.MsgPreprepare:
-		var p *bft.Preprepare
-		err := msg.Decode(&p)
-		if err == nil {
-			backlog.Push(msg, toPriority(msg.Code, p.View))
-		}
-		// for bft.MsgRoundChange, bft.MsgPrepare and bft.MsgCommit cases
-	default:
-		var p *bft.Subject
-		err := msg.Decode(&p)
-		if err == nil {
-			backlog.Push(msg, toPriority(msg.Code, p.View))
-		}
+	// toPriority truncates the sequence, so it runs only after
+	// isBacklogSequenceTooFar has rejected sequences that do not fit in uint64.
+	backlog.Push(msg, toPriority(msg.Code, view))
+	c.backlogMessageCount++
+	c.backlogSenderBytes[src] += messageBytes
+	c.backlogPayloadBytes += messageBytes
+}
+
+func (c *core) isBacklogSequenceTooFar(sequence *big.Int) bool {
+	if !sequence.IsUint64() {
+		return true
 	}
-	c.backlogs[src] = backlog
+	maxSequence := new(big.Int).Add(c.currentView().Sequence, big.NewInt(maxBacklogSequencesAhead))
+	return sequence.Cmp(maxSequence) > 0
+}
+
+func backlogMessageBytes(msg *bft.Message) uint64 {
+	return uint64(len(msg.Msg)) + uint64(len(msg.Signature)) + uint64(len(msg.CommittedSeal))
+}
+
+func exceedsBacklogLimit(used, additional, limit uint64) bool {
+	return additional > limit || used > limit-additional
+}
+
+// backlogLimitReached reports whether retaining a message would exceed a
+// per-sender or global backlog limit. backlogsMu must be held by the caller.
+func (c *core) backlogLimitReached(src common.Address, backlog *prque.Prque, messageBytes uint64) bool {
+	if backlog != nil && backlog.Size() >= maxBacklogMessagesPerSender {
+		return true
+	}
+	if c.backlogMessageCount >= maxBacklogMessages {
+		return true
+	}
+	if exceedsBacklogLimit(c.backlogSenderBytes[src], messageBytes, maxBacklogPayloadBytesPerSender) {
+		return true
+	}
+	return exceedsBacklogLimit(c.backlogPayloadBytes, messageBytes, maxBacklogPayloadBytes)
+}
+
+// removeBacklogMessage updates accounting while backlogsMu is held. It also
+// drops the sender's byte entry once its last retained message is removed, so
+// no explicit cleanup is needed when the sender's queue becomes empty.
+func (c *core) removeBacklogMessage(src common.Address, msg *bft.Message) {
+	messageBytes := backlogMessageBytes(msg)
+	c.backlogMessageCount--
+	c.backlogPayloadBytes -= messageBytes
+	if senderBytes := c.backlogSenderBytes[src]; senderBytes > messageBytes {
+		c.backlogSenderBytes[src] = senderBytes - messageBytes
+	} else {
+		delete(c.backlogSenderBytes, src)
+	}
 }
 
 func (c *core) processBacklog() {
@@ -130,12 +201,11 @@ func (c *core) processBacklog() {
 		}
 
 		logger := c.logger.NewWith("from", src, "state", c.state)
-		isFuture := false
 
 		// We stop processing if
 		//   1. backlog is empty
 		//   2. The first message in queue is a future message
-		for !(backlog.Empty() || isFuture) {
+		for !backlog.Empty() {
 			m, prio := backlog.Pop()
 			msg := m.(*bft.Message)
 			var view *bft.View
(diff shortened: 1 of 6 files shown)


consensus/istanbul/core/roundchange.go
```diff
@@ -160,34 +166,57 @@ func (c *core) handleRoundChange(msg *bft.Message, src common.Address) error {
...
-	if rcs.roundChanges[round] == nil {
-		rcs.roundChanges[round] = newMessageSet(rcs.qualified)
+	maxRound := new(big.Int).Add(currentRound, big.NewInt(maxRoundChangeRoundsAhead))
+	if messageRound.Cmp(maxRound) > 0 {
+		return 0, errRoundChangeTooFar
 	}
-	err := rcs.roundChanges[round].Add(msg)
-	if err != nil {
+
+	// A bucket is stored only after a message is accepted, so a rejected message
```
2dvorak, line 219: Could we also ensure that messages accepted into these buckets have a bounded size? This retention path is separate from the backlog byte accounting.

consensus/istanbul/core/backlog.go
```diff
@@ -36,6 +38,22 @@ var msgPriority = map[uint64]int{
...
 }
 
+const (
+	// The limits below bound both the number and payload size of retained future
+	// messages. A PREPREPARE carries an entire block, so a count-only limit is
+	// not sufficient to bound backlog memory.
+	maxBacklogMessagesPerSender     = 128
+	maxBacklogPayloadBytesPerSender = 16 * 1024 * 1024
+	// Global limits allow up to eight senders to reach their per-sender limits.
+	maxBacklogMessages     = 1024
```
yoomee1313: Only 8 senders are needed to exhaust maxBacklogMessages, and our committee size can reach 50. A node a few sequences behind can hit this without any malicious sender (50 × 8 × 3 ≈ 1200). 
Can we manage for PREPREPARE separately (tight count + the byte caps) and let the small messages have a committee-sized budget, e.g. 128 * 50?
So the byte size of 1024 of them is ~205 KiB — 0.16% of the byte cap. So I think it will okay if we can increase to 128 * 50.

consensus/istanbul/core/backlog.go
```diff
@@ -99,25 +117,78 @@ func (c *core) storeBacklog(msg *bft.Message, src common.Address) {
...
+}
+
+// backlogLimitReached reports whether retaining a message would exceed a
+// per-sender or global backlog limit. backlogsMu must be held by the caller.
+func (c *core) backlogLimitReached(src common.Address, backlog *prque.Prque, messageBytes uint64) bool {
+	if backlog != nil && backlog.Size() >= maxBacklogMessagesPerSender {
+		return true
+	}
+	if c.backlogMessageCount >= maxBacklogMessages {
+		return true
```
yoomee1313, line 242: It ignores new message only. backlog is prque so it might can utilize priority. How's about that?

consensus/istanbul/core/roundchange.go
da1suk8: Thank you, done in 47fa042: `Add` now rejects a ROUND CHANGE over 1 KiB, which bounds
the path to about 4.3 MiB (129 round buckets x a quorum of senders x 1 KiB).

consensus/istanbul/core/backlog.go
da1suk8: Thank you, done as suggested in c2b4a85: small messages now get a committee-sized global
budget, `maxBacklogMessagesPerSender * maxBacklogSenders` = 128 * 50 = 6400.
PREPREPARE is counted separately at 16 per sender and 128 globally. The byte
caps are unchanged.

The PREPREPARE numbers are mine, so flag them if they look off. Also worth
noting `istanbul.committeesize` has no upper bound, so 50 is an assumption
rather than a ceiling.

consensus/istanbul/core/backlog.go
da1suk8: Fair point. Three reasons I left it out:

- After c2b4a85 that global check is reached only once roughly 50 senders are
  each at their own cap, so drops now happen at the per-sender limit.
- Priority does not stop this case: `toPriority` orders by sequence, so a sender
  can send at the current sequence and survive any eviction.
- `prque` only pops the highest priority, so evicting the farthest-future entry
  needs a new API or an O(n) rebuild.

Happy to add per-sender eviction in a follow-up.

2dvorak (dismissed):

consensus/istanbul/core/backlog.go
yoomee1313: After permissionless, the number of consensus nodes will be limited to the `DefaultMaxValActivePausedCount`.  How's about using this value?

consensus/istanbul/core/backlog.go
yoomee1313: I missed that toPriority part. Sure, there's no need to add new priority rule.

consensus/istanbul/core/backlog.go
da1suk8: Good pointer — it made me notice I justified 50 with the wrong quantity: handleMsg gates on c.current.qualified, so the bound is the council, not the committee. Reworded the comment to say that and to cite DefaultMaxValActivePausedCount. Value unchanged. - eba3bd6


I cited it in the comment rather than importing it: the effective cap is read from ABv2, so the Go constant — only used at genesis setup — would not track it. Importing kaiax/valset/impl into core would also add 37 packages.

yoomee1313: Sure, looks good to me.

2dvorak (dismissed):

consensus/istanbul/core/roundchange.go
```diff
@@ -160,34 +174,63 @@ func (c *core) handleRoundChange(msg *bft.Message, src common.Address) error {
...
+func (rcs *roundChangeSet) Add(currentRound, messageRound *big.Int, msg *bft.Message) (int, error) {
+	// roundChanges is keyed by uint64, so reject values that would truncate to
+	// an existing bucket even if callers bypass checkMessage.
+	if !messageRound.IsUint64() {
+		return 0, bft.ErrInvalidMessage
+	}
+
+	// Retained ROUND CHANGE messages are not covered by the backlog byte
+	// accounting, so bound each one here.
+	if retainedMessageBytes(msg) > maxRoundChangeMessageBytes {
```
yoomee1313, line 207: This limit exists because the envelope's CommittedSeal is only validated for COMMIT, but it's enforced inside `roundChangeSet.Add`, which is one of three paths that retain a message. The other two are unbounded:
* backlog — a future-sequence ROUND CHANGE returns errFutureMessage from checkMessage and reaches storeBacklog without ever passing Add; PREPARE and COMMIT never touch roundChangeSet at all.
* messageSet — addVerifiedMessage stores the whole message per address, so c.current.Prepares can hold council × ProtocolMaxMsgSize (≈600 MiB at 50 validators). This one isn't covered by the PR at all.

How's about moving the check to `handleCheckedMsg`, before the three paths diverge?

consensus/istanbul/core/backlog.go
yoomee1313: @da1suk8 
I rethink about it and I found that eight nodes can take the whole PREPREPARE budget, leaving none for the honest proposer.

`storeBacklog` doesn't check whether the sender is the proposer of the message's view, and a future PREPREPARE is retained before `backend.Verify` runs, so it doesn't even have to be a valid block. The problem is `maxBacklogPreprepareMessages`: an attacker sends small proposals to exhaust the count cap, 16 per sender, so 8 senders are enough.

Once it's full the rule is to throw away the incoming message, so an honest PREPREPARE is dropped too. And the budget doesn't drain: `processBacklog` pops a retained message, finds it still in the future, and pushes it straight back.

I think this should be fixed by not sharing the PREPREPARE budget between senders — retain at most one PREPREPARE per sender at any time.

This matters most during a round change, when checkMessage returns errFutureMessage even for the current view, so the next round's PREPREPARE is always retained rather than handled. If it is dropped there, the proposer never retransmits and the round just times out.

* Retain at most one PREPREPARE per sender at any time, so 8 senders take 8 slots and the honest proposer's slot is always free. maxBacklogPreprepareMessagesPerSender and maxBacklogPreprepareMessages both go away — the policy is just sender.preprepares > 0.
* Exclude PREPREPARE from the byte accounting as well

By the way, if the size check I suggested in the other comment is applied, the byte accounting goes away entirely and the second and third points here are moot.

da1suk8: Done in da3de67.

Confirmed: the proposer check and `backend.Verify` sit after `checkMessage`, 8 senders are within f, and processBacklog runs only from setState, which nothing calls while waiting for a round change.

Taking one PREPREPARE per sender, with one change: when the slot is taken, keep the newest instead of dropping the incoming one — otherwise a sender that proposes twice while we are stuck loses its current PREPREPARE to its own stale one. prque has no removal, so the slot becomes its own map, checked before the queue.

Byte accounting goes entirely, and so does the small-message global we settled on earlier (`maxBacklogMessages`, `maxBacklogSenders`): any global fills the same way — 8 × 16 MiB for bytes, 50 × 128 once the council exceeds the assumption. Per sender stays at 128. What remains is council × block size for PREPREPARE, ≈500 MiB at 50 — the price of guaranteeing every sender a slot; proposer-aware retention would lower it but only works for the current sequence. I would take the 500 MiB; say if you see it differently.

consensus/istanbul/core/roundchange.go
da1suk8: Done in 084c6d0.

Agreed — `handleCheckedMsg` is the only caller of the four handlers (self-broadcast included), so one check there covers all three paths. Two small corrections: backlog bytes are already capped (16 / 128 MiB), only the per-message bound is missing; and COMMIT is already covered by the seal recovery in handleCommit, so the exposed set is c.current.Prepares alone.

One addition: the check must bound the body, not just the seal — our rlp does not cap *big.Int, so a PREPARE with a multi-MiB View.Round is retained today. So PREPARE / COMMIT / ROUND CHANGE at 1 KiB on the envelope, Add loses its check, and PREPREPARE stays on the block bound.

da1suk8:
Pushed 084c6d0 and da3de67 and rewrote the description: no shared budget remains — retention is per sender × per message, PREPREPARE is one slot per sender, and the size check moved to handleCheckedMsg.

consensus/istanbul/core/backlog.go
```diff
@@ -98,67 +127,136 @@ func (c *core) storeBacklog(msg *bft.Message, src common.Address) {
...
+	if c.isBacklogSequenceTooFar(view.Sequence) {
+		logger.Trace("Discarding future message: sequence is too far ahead", "sequence", view.Sequence)
+		return
+	}
+
+	if msg.Code == bft.MsgPreprepare {
+		if _, replaced := c.backlogPreprepares[src]; replaced {
+			logger.Debug("Replacing retained PREPREPARE", "view", view)
+		}
+		c.backlogPreprepares[src] = msg
```
yoomee1313: Minor concern. It just replaces the existing one without checking the view. Can we check the view if backlogPreprepares[src] exists?

da1suk8: @yoomee1313 
Thank you so much!

Done in f65ead3. The slot now keeps the highest view: an incoming PREPREPARE replaces the retained one only when its view is higher, and an equal or lower view is dropped, so a delayed older proposal cannot evict the one we are waiting for. 
The view is decoded once at retention and kept with the message, so neither the comparison nor processBacklog decodes the block again.

2dvorak (approved):

yoomee1313 (approved):

PR #1054, opened 2026-08-28
kaiax/vrank, consensus: record the parent round with a certificate

hyunsooda:
## Proposed changes

The consensus round is stored in a header byte that both header hashes clear, so two nodes can end up holding the same block with different rounds. `pfReport` is derived from that byte, so those nodes then compute different proposal failures, different node-state transitions and different qualified validator sets, and the node on the minority side rejects the next block with `header extra validators do not match qualified validators` and cannot recover.

This records the parent's round in `header.VRank` instead, with the parent's committed seals as proof that a committee quorum committed it at that round. Every node then reads one value agreed by the quorum that committed the block carrying it, rather than whichever round it happened to store. Epoch starts record no parent round, since their parent belongs to the previous epoch whose failures must not score into this one.

Verification checks only that the claimed round happened. It never compares against the node's own stored round: two honest nodes legitimately hold different rounds for one block, so comparing would make them reject each other's headers. A proposer can therefore claim a lower round that really happened, but cannot invent a higher one.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

Verified on a four-node local network with #1022 and #1031 applied on top of this branch. Both are needed for the scenario to exist at all: without #1022 a hash-locked re-proposal is rejected outright, so a second representation never forms. The control was the same two PRs on `dev` without this branch, so the only difference between the two runs is the commits here.

A temporary patch made three nodes drop that height's round-0 COMMITs and its propagated block, so one node committed the block at round 0 while the other three committed the same block, with the same hash, at round 1. On the control the node holding round 0 derived an empty proposal failure report, kept a four-validator qualified set and rejected the next block with `header extra validators do not match qualified validators`; the block went into the bad block database and that node stopped following the chain. With this branch the same split produced identical proposal failure reports, node states and state roots on all four nodes, and the node stayed with the chain. A separate run confirmed that the block following a round change carries the parent round together with a quorum of the parent's committed seals, and that every node accepts it.

The first commit belongs here rather than in a separate PR. This change makes committed seals decide proposal failures, so the question becomes whether seals can be obtained for a round that never happened. They can: the COMMIT rebroadcast for an already-committed block signs whatever round the caller names, without checking it against the round the responder stored. A caller can therefore gather a genuine quorum certificate for a round at which no quorum ever formed, and claim any number of failures. The first commit closes that by answering only for the round the node stored, which is what makes "cannot invent a higher round" true.

Determinism is what this buys; the recorded round is not guaranteed to be the highest one that happened, since absence cannot be proven. The header round byte is left in place for consensus and is no longer read by any state-affecting path. No chain has the permissionless fork enabled, so the format change affects no existing network.

The `GetProposer` fast path returns a block's author rather than the round's scheduled proposer, which makes `pfReport` node-dependent again once a recorded round reaches 2. #1031 removes it and should land alongside this.

### DIFF
--- a/consensus/engine/sealer.go
+++ b/consensus/engine/sealer.go
@@ -87,8 +87,8 @@ func (s *dynamicSealer) CommittersWithRound(header *types.Header) ([]common.Addr
 	return s.implAt(header.Number.Uint64()).CommittersWithRound(header)
 }
 
-func (s *dynamicSealer) RecoverCommitters(hash common.Hash, round byte, seals [][]byte) ([]common.Address, error) {
-	return s.implAt(0).RecoverCommitters(hash, round, seals) // implAt ignores the number
+func (s *dynamicSealer) RecoverCommitters(blockNum uint64, hash common.Hash, round byte, seals [][]byte) ([]common.Address, error) {
+	return s.implAt(blockNum).RecoverCommitters(blockNum, hash, round, seals)
 }
 
 func (s *dynamicSealer) Vanity(header *types.Header) ([]byte, error) {
--- a/consensus/engine/sealer.go
+++ b/consensus/engine/sealer.go
@@ -87,6 +87,10 @@ func (s *dynamicSealer) CommittersWithRound(header *types.Header) ([]common.Addr
 	return s.implAt(header.Number.Uint64()).CommittersWithRound(header)
 }
 
+func (s *dynamicSealer) RecoverCommitters(hash common.Hash, round byte, seals [][]byte) ([]common.Address, error) {
+	return s.implAt(0).RecoverCommitters(hash, round, seals) // implAt ignores the number
+}
+
 func (s *dynamicSealer) Vanity(header *types.Header) ([]byte, error) {
 	if header == nil || header.Number == nil {
 		return nil, consensus.ErrUnknownBlock
--- a/consensus/istanbul/core/preprepare.go
+++ b/consensus/istanbul/core/preprepare.go
@@ -84,11 +84,16 @@ func (c *core) handlePreprepare(msg *bft.Message, src common.Address) error {
 			// Broadcast COMMIT if it is an existing block
 			// 1. The proposer needs to match the given (Sequence + Round)
 			// 2. The given block must exist
+			// 3. Post-permissionless, the stored round must match: the seal binds the round,
+			//    so answering another round would attest a quorum that never formed.
 			proposer, getProposerErr := c.valsetModule.GetProposer(preprepare.View.Sequence.Uint64(), preprepare.View.Round.Uint64())
 			if getProposerErr != nil {
 				return getProposerErr
 			}
-			if proposer == src && c.backend.HasPropsal(preprepare.Proposal.Hash(), preprepare.Proposal.Number()) {
+			storedRound, hasProposal := c.backend.ProposalRound(preprepare.Proposal.Hash(), preprepare.Proposal.Number())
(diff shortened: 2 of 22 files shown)


yoomee1313 (dismissed):

consensus/istanbul/core/preprepare.go
```diff
@@ -84,11 +84,16 @@ func (c *core) handlePreprepare(msg *bft.Message, src common.Address) error {
 			// Broadcast COMMIT if it is an existing block
 			// 1. The proposer needs to match the given (Sequence + Round)
 			// 2. The given block must exist
+			// 3. Post-permissionless, the stored round must match: the seal binds the round,
+			//    so answering another round would attest a quorum that never formed.
```
ian0371: Couldn't understand the comment. Suggestion:

```suggestion
			// This PRE-PREPARE targets an already-finalized height. Reply with a COMMIT
			// to help the sender finish that block, only if:
			// 1. The sender is the scheduled proposer of the given (sequence, round).
			// 2. This node has the given block in its chain.
			// 3. Post-permissionless: the given round equals the round this node stored.
			//    The committed seal signs (hash, round), so answering another round would
			//    sign a commit at a round this node never committed at; callers could
			//    collect such seals into a quorum certificate for a round that never
			//    reached quorum.
```

consensus/engine/sealer.go
```diff
@@ -87,6 +87,10 @@ func (s *dynamicSealer) CommittersWithRound(header *types.Header) ([]common.Addr
 	return s.implAt(header.Number.Uint64()).CommittersWithRound(header)
 }
 
+func (s *dynamicSealer) RecoverCommitters(hash common.Hash, round byte, seals [][]byte) ([]common.Address, error) {
+	return s.implAt(0).RecoverCommitters(hash, round, seals) // implAt ignores the number
```
ian0371: For future-proofing, how about passing the block number here (or simply `header`) and using `implAt(blockNum)` instead of `implAt(0)`?

kaiax/vrank/impl/scoring.go
```diff
@@ -152,10 +152,15 @@ func (v *VRankModule) epochCandidates(blockNum uint64) ([]common.Address, error)
 		return nil, err
 	}
 	logger.Warn("CandTesting unavailable from state, reading the epoch-start header", "num", blockNum, "err", err)
-	return vrank.DecodeReport(header.VRank)
+	payload, decodeErr := vrank.DecodeVRank(header.VRank)
+	if decodeErr != nil {
+		return nil, err
```
ian0371: Let's return `decodeErr` as well:

```suggestion
		return nil, fmt.Errorf("get CandTesting from state err: %v; decode epoch-start header VRank err: %w", err, decodeErr)
```

ian0371 (commented):
Could we add two small tests?
- `TestRecoverCommitters` unit test in `consensus/istanbul/sealer_test.go`: the `len(seal) != IstanbulExtraSeal` → `ErrInvalidCommittedSeals` branch
- A `DecodeVRank` case rejecting a 2-byte round

hyunsooda:
@ian0371  All done, but one side effect is that `IstanbulSealerRecoverCommitters` has an unused parameter.

ian0371:
@hyunsooda please rebase to latest dev (CI failure)

ian0371 (approved):

2dvorak (approved):

PR #1056, opened 2026-09-01
kaiax/gov: validate the governing node successor

hyunsooda:
## Proposed changes

The consistency check for a governing-node vote compares the *current* governing node against the council and never inspects the proposed value, which only has to be a well-formed address. After Permissionless only the governing node may vote, and voting requires council membership, so a successor that is the zero address or outside the council leaves header governance with no way back. Validate the proposed value where that rule applies.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

The check is gated on Permissionless because `VerifyVote` also runs on pre-fork headers, and tightening it unconditionally could reject a header that was accepted before.

### DIFF
--- a/kaiax/gov/headergov/impl/header.go
+++ b/kaiax/gov/headergov/impl/header.go
@@ -164,10 +164,7 @@ func (h *headerGovModule) checkConsistency(blockNum uint64, vote headergov.VoteD
 			return nil
 		}
 
-		// we'll use blockNum-1 for the blocknumber of GetCouncil since blockNum cannot be available(eg. vote)
-		// it's definite that the valSet vote is not included in this block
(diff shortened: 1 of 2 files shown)


ian0371 (dismissed):

2dvorak:
@hyunsooda Is it OK to validate the successor against the council at N-1 instead of N in `header.go:170`?

hyunsooda:
@2dvorak Fixed. It checks whether both are in the council in the N which is a block to be inserted

2dvorak (approved):

ian0371 (approved):

PR #1058, opened 2026-09-02
node/cn: expect the bid conntype check in the test

hyunsooda:
## Proposed changes

`handleBidMsg` gained a conntype check, but the size-gate subtests set up their peer mock without it, so the call is rejected and the handler returns before reaching the pool. Expect it, the way the sibling cases in this file already do.

## Types of changes

- [ ] 🐛 Bug fix
- [x] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

## Further comments

`dev` currently fails this test: the check and these subtests landed in separate PRs, so neither CI run saw the other.


hyunsooda:
Folded into #1054, which now sits on top of the conntype check.

PR #1068, opened 2026-09-09
storage/statedb: separate node and code sync cache keys

2dvorak:
## Proposed changes

- Use distinct cache keys for trie nodes and contract code.
- Apply the typed keys consistently to cache lookups and writes.
- The behavior change is limited to state migration; persisted state data and other synchronization paths remain unchanged.

## Types of changes

- [x] 🐛 Bug fix
- [ ] 🧪 Test improvements
- [ ] 🧰 CI / build tool

## Checklist

### DIFF
--- a/storage/statedb/sync.go
+++ b/storage/statedb/sync.go
@@ -108,6 +108,11 @@ type syncMemBatch struct {
 	codes map[common.Hash][]byte // In-memory membatch of recently completed codes
 }
 
+type (
+	syncNodeExistKey common.Hash
+	syncCodeExistKey common.Hash
+)
+
 // newSyncMemBatch allocates a new memory-buffer for not-yet persisted trie nodes.
 func newSyncMemBatch() *syncMemBatch {
 	return &syncMemBatch{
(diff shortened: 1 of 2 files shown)


da1suk8 (approved):

storage/statedb/sync.go
```diff
@@ -108,6 +108,11 @@ type syncMemBatch struct {
 	codes map[common.Hash][]byte // In-memory membatch of recently completed codes
 }
 
+type (
+	syncNodeExistKey common.Hash
+	syncCodeExistKey common.Hash
+)
```
yoomee1313: This is not a bug in the current implementation, but since `s.exist.Get(hash)` with a `common.Hash` would compile successfully and silently miss entries stored as syncNodeExistKey or syncCodeExistKey.  

How's about wrapping the raw Get/Add operations behind type-specific methods such as GetNode/GetCode and AddNode/AddCode? 

This would centralize the key conversions and reduce the chance of accidentally using a plain common.Hash.

